PR #15, opened 2025-04-03
feat: RpcAdjudicator

nksazonov:


mod:
There is few critical decision and protocol problem we need to address:
- Do we use RPCStateHash or stateHash ?
- Do we use a magic number of funding State.data ?
- How to define types, where the clients sign a different Data, than the server.

The Client can sign his RPC request, with a stateHash,
but the client client include RPC call result has he doesn't know it yet.

The Server can sign both params, and return value, and the stateHash are for different object states.

stateHash and RPCStateHash are redondunt.

@nksazonov maybe the Client sign the stateHash
and the server signs stateHash+response ?

nksazonov:
In your code 
``` solidity
// Compute the state hash for signature verification
bytes32 stateHash = Utils.getStateHash(chan, candidate);

// Verify both signatures
bool hostSigValid = Utils.verifySignature(stateHash, candidate.sigs[HOST], chan.participants[HOST]);
bool guestSigValid = Utils.verifySignature(stateHash, candidate.sigs[GUEST], chan.participants[GUEST]);

if (!hostSigValid || !guestSigValid) {
    return false;
}
```

both request and response have to be signed by Client and Server. I think, we can omit Server sig on request.

---

> Do we use RPCStateHash or stateHash ?

I also agree that both of them simultaneously are redundant. Let's stick with `stateHash` for now, as it also includes the appData, which is request-unique.

---

> Do we use a magic number of funding State.data ?

I am not against, nor for it. I don't think it adds in terms of security in any way. I have not seen any blockchain protocols use magic numbers.

---

> How to define types, where the clients sign a different Data, than the server.

I don't think we need different types. The following type:

```solidity
struct RPCMessage {
    uint64 requestID;
    uint64 timestamp;
    string method;
    bytes[] data;
}
```

... is request-response-agnostic, as for REQUEST `data` is params, while for `RESPONSE` it is `result`. WDYT?

nksazonov:
Not relevant anymore

PR #16, opened 2025-04-03
feat(utils): perform eip-191 eth sign message

nksazonov:


mod (approved):

mod:
Im actually against it, there is no need to prefix again stateHash we shoukd vhange this in all others messages

nksazonov:
> Im actually against it, there is no need to prefix again stateHash we should change this in all others messages

@mod , sure, we can do it later. It is not easy to do, as all common libs (viem, Metamask, geth) do NOT support signing without EIP-191 anymore.

PR #23, opened 2025-04-05
Adding protocol update

mod:


docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+
+For signature verification, the final hash MUST use Ethereum's personal sign message format:
+```solidity
+signatureHash = keccak256(
+    abi.encodePacked(
+        "\x19Ethereum Signed Message:\n32",
+        stateHash
+    )
+);
+```
```
nksazonov, line 128: Why only the final hash has to comply with EIP-191?

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+
+### 7.3. Checkpointing
+
+1. Any participant MAY:
+   - Call the `checkpoint` function with a valid state and required proofs
+
+2. The system MUST:
+   - Verify the submitted state via the adjudicator
+   - If valid, store the state without starting a challenge period
+   - Emit a `Checkpointed` event with the channelId
```
nksazonov, line 221: 2 conditions needs to be added:
- `checkpoint` can substitute older `checkpoint`
- `checkpoint` can remove older `challenge`, moving Channel to ACTIVE status
Which state is older and newer is determined by `IComparable` interface implemented by Adjudicator

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+
+1. Each non-Creator participant MUST:
+   - Verify the channelId and expected allocations
+   - Sign the same funding stateHash (containing the magic number `CHANOPEN`)
+   - Call the `join` function with the channelId, their participant index, and signature
+
+2. The system MUST:
+   - Verify the participant's signature against the funding stateHash
+   - Confirm the signer matches the expected participant at the given index
+   - Lock the participant's funds according to the allocation
```
nksazonov, line 158: If there are different "funding" states signed singlehandedly by each participant, then ChannelImpl MUST check stateHashes are the same (meaning all allocations are the same) to AVOID situations, where some participant changed their allocation to be 0, but other participants think this party has deposited some funds.

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+        stateHash
+    )
+);
+```
+
+## 5. Funding Protocol
+
+### 5.1. Creation Phase
+
+1. The Creator MUST:
```
MaxMoskalenko, line 134: 1. Why do we need to distinguish between Creator and non-creator. Does Creator has some unique abilities?
2. If participants collect all signatures to close, why they cannot collect all signatures to open? In my opinion it doesn't change anything, but are there any justifications for it?

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+   - Call the `reset` function with:
+     - The current channelId
+     - A valid closing state with all signatures
+     - Any required proofs
+     - A new channel configuration
+     - A new initial state with the `CHANOPEN` magic number
+
+2. The system MUST:
+   - Close the existing channel as per the cooperative close process
+   - Create a new channel following the creation process
```
MaxMoskalenko, line 235: "Create channel", does it mean, that configs could be different. Like different participants/challenge or just different allocations?

docs/PROTOCOL.md
mnaichuk: agree on the first point, the new `checkpoint` should replace the previous one

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+
+This document specifies the Nitrolite state channel protocol, a system that enables off-chain interactions between participants with minimal on-chain operations. This specification defines the data structures, message formats, and protocol rules.
+
+## 2. Terminology
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+- **Channel**: A secure communication pathway between participants
+- **State**: A snapshot of the channel at a point in time
+- **Adjudicator**: A contract that validates state transitions
```
dimast-x, line 13: A contract that validates state validity

docs/PROTOCOL.md
dimast-x: Claude is STUPID.

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+## 5. Funding Protocol
+
+### 5.1. Creation Phase
+
+1. The Creator MUST:
+   - Construct a channel configuration with participants, adjudicator, challenge period, and nonce
+   - Prepare an initial state where `state.data` MUST be set to the magic number `CHANOPEN` (7877)
+   - Define expected token deposits for all participants in the `state.allocations` array
```
dimast-x, line 137: required token deposit can be zero

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+   - Sign the stateHash of this initial state
+   - Include their signature in the `state.sigs` array at position 0
```
dimast-x, line 139: Only channel creator signature is needed at this stage

dimast-x, line 139: No more other signatures is needed. Only creator's

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+   - Verify the Creator's signature on the funding stateHash
+   - Lock the Creator's funds according to the allocation
+   - Set the channel status to `INITIAL`
+   - Emit a `Created` event with the channelId, channel configuration, and expected deposits
+
+### 5.2. Joining Phase
+
+1. Each non-Creator participant MUST:
+   - Verify the channelId and expected allocations
+   - Sign the same funding stateHash (containing the magic number `CHANOPEN`)
```
dimast-x, line 152: must have sufficient balance to fund his required allocation

docs/PROTOCOL.md
mnaichuk: This simplifies the channel creation flow for the end user, also there is potential application use cases where the user creates a channel and waits for another participant @MaxMoskalenko

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+
+## 8. Reset Protocol
+
+1. To reset a channel (for resizing allocations), any participant MAY:
+   - Call the `reset` function with:
+     - The current channelId
+     - A valid closing state with all signatures
+     - Any required proofs
+     - A new channel configuration
+     - A new initial state with the `CHANOPEN` magic number
```
nksazonov, line 231: Only in specs (in interface it is present) we also are missing explicit signatures on a new state mentioning

docs/PROTOCOL.md
nksazonov: I also thought about it, but on some lines above it is clearly stated that "Sign the *same* funding stateHash". If anything is changed, then it would not be the same stateHash.

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+   - Be signed by the necessary participants according to application rules
+   - Comply with the validation rules of the channel's adjudicator
+
+## 7. Channel Closure
+
+### 7.1. Cooperative Close
+
+1. To close cooperatively, any participant MAY:
+   - Prepare a final state where `state.data` MUST be set to the magic number `CHANCLOSE` (7879)
+   - Collect signatures from all participants on this final state
```
mnaichuk, line 185: Maybe as an idea, we can also make a close method similar to an open:
1. Initiator of the close submit close request
2. Another participant should receive the needed notification
3. Approve the close request and conclude the channel, or challenge it

And as a security mechanism, add a cooldown mechanism for expired close requests

docs/PROTOCOL.md
nksazonov: I think a "contract that validates state progression to candidate based on proofs" is a better description.

docs/PROTOCOL.md
nksazonov: Here "their" is a genderless pronoun. However, I agree with you, it should better said "Include creator's signature..."

docs/PROTOCOL.md
nksazonov: I think it is better to have "close" atomic to avoid dealing with half-transitioned state.

MaxMoskalenko: One-sided close with approve or disapprove duration is already implemented (designed) it's called "challenge"

docs/PROTOCOL.md
```diff
@@ -0,0 +1,243 @@
...
+   - Call the `join` function with the channelId, their participant index, and signature
+
+2. The system MUST:
+   - Verify the participant's signature against the funding stateHash
+   - Confirm the signer matches the expected participant at the given index
+   - Lock the participant's funds according to the allocation
+   - Track the actual deposit in the channel metadata
+   - Emit a `Joined` event with the channelId and participant index
+
+3. When all participants have joined, the system MUST:
```
dimast-x, line 162: After each participant joins, the system MUST:

docs/PROTOCOL.md
mod: 1. Anyone who propose a funding state is the creator.
2. Yes, we may change this also to incremental closing, we will have to prototype it, maybe it's also painful to collect closing signatures.

I leave the `close()` using consensus for now, in order to save on close() blockchain calls, to avoid having every participant calling close.

MaxMoskalenko: But still, in nitro the same could have been achieved with participant[0]. And having two methods instead of one with no particular purposes sounds more like complication than simplification. 

+ keep in mind, that this methods shouldn't be "user friendly", since end users will not likely see/use them anyways. That's more for the developers, who could understand (and accept) protocol abstractions

docs/PROTOCOL.md
mnaichuk: ok ok, make sense I guess

docs/PROTOCOL.md
mod: Yes, During funding process we should save initial state and make sure stateHash in the same when new participant are joining. I will add in the spec

docs/NitroRPC.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+## Solidity types
+
+```solidity
+struct RPCMessage {
+    uint64 requestID;
+    string method;
+    bytes[] params;
+    bytes[] result;
```
nksazonov, line 78: It should be `bytes` (array of bytes) as `bytes[]` is an array of byte arrays.

docs/PROTOCOL.md
mod: added

docs/PROTOCOL.md
mod: I thought about this too @mnaichuk we may fallback to it, if we don't find an easy way to happy-close

docs/NitroRPC.md
```diff
@@ -0,0 +1,161 @@
...
+  )
+);
+
+// Alternatively from RPCMessage type
+keccak256(
+  abi.encode(
+    rpcMessage
+  )
+)
+```
```
nksazonov, line 109: This is wrong. 
Requester must sign:
```solidity
keccak256(
   abi.encode(
      reqId,
      method,
      params,
      timestamp,
      channelId,
      allocations
   )
)
```
 whereas Responder must sign the same struct with addition of `result`.

contract/src/interfaces/IChannel.sol
```diff
@@ -8,19 +8,28 @@ import {Channel, State} from "./Types.sol";
  * @notice Main interface for the state channel system
  */
 interface IChannel {
-    event ChannelPartiallyFunded(bytes32 indexed channelId, Channel channel);
-    event ChannelOpened(bytes32 indexed channelId, Channel channel);
-    event ChannelChallenged(bytes32 indexed channelId, uint256 expiration);
-    event ChannelCheckpointed(bytes32 indexed channelId);
-    event ChannelClosed(bytes32 indexed channelId);
+    event Created(bytes32 indexed channelId, Channel channel, Amount[] expected);
```
nksazonov, line 11: Why do we need an `Amount[]` only for creator? I would rather we remove it alltogether, replaced with `Allocations[]` or add to `Joined` too. Although it is a bad practice in Solidity in general to put Structs in events

contract/src/interfaces/IComparable.sol
```diff
@@ -0,0 +1,26 @@
...
+ */
+interface IComparable {
+    /**
+     * @notice Compare two states and determine their relative ordering
+     * @param candidate The state being evaluated
+     * @param previous The reference state to compare against
+     * @return result The comparison result:
+     *         -1: candidate < previous
+     *          0: candidate == previous
+     *          1: candidate > previous
```
nksazonov, line 22: An enum can be created for this

contract/src/interfaces/IComparable.sol
```diff
@@ -0,0 +1,26 @@
...
+     */
+    function compare(State calldata candidate, State calldata previous) external view returns (uint8 result);
```
nksazonov, line 24: Moreover, it should be `int8` to allow it to hold `-1` value :D

contract/src/interfaces/Types.sol
```diff
@@ -11,39 +11,53 @@ struct Signature {
...
+struct Amount {
+  address token; // ERC-20 token contract address
+  uint256 amount; // Token amount allocated
+}
+
 struct Allocation {
     address destination; // Where funds are sent on channel closure
     address token; // ERC-20 token contract address
     uint256 amount; // Token amount allocated
+    //TODO: consider using Amount
```
nksazonov, line 23: Nested structs CAN be hard to handle in TS and Go (if they contain reference types, that make encoding much less straightforward).

docs/NitroRPC.md
MaxMoskalenko: I'm not fully understand this part. But protocol has one visible flow: parties don't exchange with allocation and signatures (only appdata). And signing appdata seems to have little effect on channel `close`, `reset` and `checkpoint`. So either there are should be protocol on top of it for exchanging States (appdata + allocations + sigs) or rpc protocol should exchange States in a first place... Oh, no the same problem we had with Nitro

contract/src/interfaces/Types.sol
```diff
@@ -11,39 +11,53 @@ struct Signature {
...
 
 enum Status {
-    VOID, // Channel was never active (zero-initialized)
-    PARTIAL, // Partial funding waiting for other participants
+    VOID, // VOID Channel was not created
+    INITIAL, // Channel is created and in funding process
     ACTIVE, // Channel fully funded and valid state
+    DISPUTE,
     FINAL, // This is the FINAL state, channel can be closed
-    INVALID // Channel state is invalid
```
nksazonov, line 38: Finally. A ton of likes from me 👍👍👍

contract/src/interfaces/Types.sol
```diff
@@ -11,39 +11,53 @@ struct Signature {
...
 }
 
-// This struct has been moved to Custody.sol with additional fields
-// Kept here for backward compatibility, but should be migrated to use the Custody.sol version
+constant uint256 CREATOR = 0; // participant index for the channel creator
+constant uint256 BROKER = 1; // participant index for the broker
+
+// Funding protocol use State with expected deposits
+constant uint32 CHANOPEN = 7877; // State.data value for funding stateHash
+constant uint32 CHANCLOSE = 7879; // State.data value for closing stateHash
```
nksazonov, line 52: In future we might want to move `open` and `close` constants out of IChannel to allow each Adjudicator to define its own values. For example, some may allow start or final states not only contain a magic number, but rather be meaningful for their respective protocol. This way we save 1 participants interaction in each open and close.

contract/src/interfaces/Types.sol
```diff
@@ -11,39 +11,53 @@ struct Signature {
...
+
+// Reference storage type for IChannel implementation to be cdustomized
 struct Metadata {
     Channel chan; // Opener define channel configuration
-    Status status; // Current channel status
+    Status stage;
```
nksazonov, line 57: We should have either `Status status` or `Stage stage`. Having type and field names mixed up is confusing.

contract/src/interfaces/Types.sol
```diff
@@ -11,39 +11,53 @@ struct Signature {
...
-// Kept here for backward compatibility, but should be migrated to use the Custody.sol version
+constant uint256 CREATOR = 0; // participant index for the channel creator
+constant uint256 BROKER = 1; // participant index for the broker
+
+// Funding protocol use State with expected deposits
+constant uint32 CHANOPEN = 7877; // State.data value for funding stateHash
+constant uint32 CHANCLOSE = 7879; // State.data value for closing stateHash
+
+// Reference storage type for IChannel implementation to be cdustomized
 struct Metadata {
```
nksazonov, line 55: I would also move this out of the interface to allow ChannelImpls to have more flexibility.

contract/src/Custody.sol
```diff
@@ -27,459 +28,276 @@ contract Custody is IChannel, IDeposit {
...
-    uint256 constant GUEST = 1;
-
     // Recommended structure to keep track of states
     struct Metadata {
-        Channel chan; // Opener define channel configuration
-        Status status; // Current channel status
-        uint256 challengeExpire; // If non-zero channel will resolve to lastValidState when challenge Expires
-        State lastValidState; // Last valid state when adjudicator was called
-        mapping(address token => uint256 balance) tokenBalances; // Token balances for the channel
+      Channel chan; // Opener define channel configuration
```
nksazonov, line 35:
```suggestion
      Channel chan; // Opener provides consensus channel configuration
```

contract/src/Custody.sol
```diff
@@ -27,459 +28,276 @@ contract Custody is IChannel, IDeposit {
...
-        } else {
-            bool success = IERC20(token).transferFrom(account, address(this), amount);
-            require(success, TransferFailed(token, address(this), amount));
-        }
-        _ledgers[msg.sender].tokens[token].available += amount;
+      address account = msg.sender;
+      if (token == address(0)) {
+        require(msg.value == amount, InvalidAmount());
+      } else {
+        //TODO: Support native token
```
nksazonov, line 64: It is already supported by the line above and `payable` modifier of this method.

contract/src/Custody.sol
```diff
@@ -27,459 +28,276 @@ contract Custody is IChannel, IDeposit {
...
-        Channel chan; // Opener define channel configuration
-        Status status; // Current channel status
-        uint256 challengeExpire; // If non-zero channel will resolve to lastValidState when challenge Expires
-        State lastValidState; // Last valid state when adjudicator was called
-        mapping(address token => uint256 balance) tokenBalances; // Token balances for the channel
+      Channel chan; // Opener define channel configuration
+      Status stage;
+      address creator;
+      Amount[] expectedDeposits; // Creator defines Token per participant
+      Amount[] actualDeposits; // Tracks deposits made by each participant
```
nksazonov, line 39: This can be simplified by counting "processed/joined participants" instead of funds, thus using `EnumerableSet` instead of these 2 arrays (please, let's not use vanilla Solidity arrays at least).

contract/src/Custody.sol
```diff
@@ -27,459 +28,276 @@ contract Custody is IChannel, IDeposit {
...
 
-        // Channel must be in a valid state to close (ACTIVE, PARTIAL, or already FINAL)
-        if (meta.status == Status.VOID || meta.status == Status.INVALID) {
-            revert InvalidStatus();
-        }
+      // Require all participant signatures on funding protocol stateHash
+      // require State.data == CHANCLOSE magic number
+      //
+      // In the case ot meta.stage == Status.DISPUTE and blocktime is higher than challengeExpire ts
+      // you can distribute and close
```
nksazonov, line 151: Nice 👍

contract/src/Custody.sol
```diff
@@ -27,459 +28,276 @@ contract Custody is IChannel, IDeposit {
...
-
-            if (isValid && hasAllSignatures && allSignaturesValid) {
-                // All requirements met, set status to FINAL
-                meta.status = Status.FINAL;
-            } else {
-                revert ChannelNotFinal();
-            }
-        }
-        // At this point, the channel is in FINAL state, so we can close it
-        _distributeAllocation(channelId, meta);
```
nksazonov, line 221: Actually, I would rather comment this logic and ask Claude to update it, and not generate out of nothing.

contract/src/Custody.sol
```diff
@@ -27,459 +28,276 @@ contract Custody is IChannel, IDeposit {
...
+      uint256 participantsLength = meta.chan.participants.length;
+      for (uint256 i = 0; i < participantsLength; i++) {
+        address participant = meta.chan.participants[i];
+        _ledgers[participant].channels.remove(channelId);
+      }
+
+      // Mark channel as closed by removing it
+      delete _channels[channelId];
+
+      emit ChannelClosed(channelId);
```
nksazonov, line 228: This should be moved 1 level up. It is a bad side-effect to this method.

contract/src/interfaces/IChannel.sol
MaxMoskalenko: Here we need to return the whole state, so other participants would know, what they need to sign

mod: Return the whole state, its a mistake of old code

contract/src/interfaces/Types.sol
mod: Dont speak to me anymore about custom funding protocol, I dont want to hear it ever again

contract/src/interfaces/Types.sol
mod: I dont agree, I dont like when type and variable are the same. This is not worth a comment at this stage, lets focus on logic

contract/src/interfaces/Types.sol
mod: After merge we can remove its a reference for AI, we could keep it in README.md

contract/src/Custody.sol
mod: I didnt read yet source code of Enumerable, Im not confortable using it, we have strict index position, we can prealloc the whole array when we create channel with zeros and all access by index its performant.

Im not yet in favor of using too much libs, I prefer vanilla, we can deal with this later

contract/src/Custody.sol
mod: Ok made the edut

contract/src/Custody.sol
nksazonov: Deal later it is. Not resolving this comment to revert back to it.

contract/src/interfaces/IChannel.sol
nksazonov: 1fcd517fd0752ce0986130383b1cda29faedf515

contract/src/interfaces/Types.sol
nksazonov: 1fcd517fd0752ce0986130383b1cda29faedf515

contract/src/Custody.sol
nksazonov: 1fcd517fd0752ce0986130383b1cda29faedf515

docs/NitroRPC.md
mod: Some NitroSpec are outdated, we need to cherry pick our NitroRPC.sol in the other branch

PR #27, opened 2025-04-22
Limit the Custody to only 2 participants

mod:


contract/src/Custody.sol
```diff
@@ -5,14 +5,26 @@ import {IChannel} from "./interfaces/IChannel.sol";
...
+    Channel,
+    State,
+    Allocation,
+    Status,
+    Signature,
+    Amount,
+    CHANOPEN,
+    CHANCLOSE,
+    CREATOR,
+    BROKER
```
nksazonov, line 18: It seems that you have not pushed changes to `Types.sol` file. Nevertheless, I think CREATOR and BROKER should be defined in Custody

nksazonov: Also, I think we better name them `USER` and `BROKER` or `CREATOR` and `RESPONDER` (or else). I don't want to mix up the action ("create") and a role ("broker").

contract/src/Custody.sol
```diff
@@ -212,14 +225,9 @@ contract Custody is IChannel, IDeposit {
...
-        bool allJoined = true;
-        for (uint256 i = 0; i < meta.actualDeposits.length; i++) {
-            if (meta.actualDeposits[i].amount != meta.expectedDeposits[i].amount) {
-                allJoined = false;
-                break;
-            }
-        }
+        // For 2-participant channels, just check if the second participant has joined
+        // since we know the first participant (creator) has already joined
+        bool allJoined = meta.actualDeposits[BROKER].amount == meta.expectedDeposits[BROKER].amount;
```
dimast-x, line 230: It seems that it missed validation for CREATOR. For some reason now it validates only BROKER

dimast-x: we can keep previous implementation with a loop -  that will work

nksazonov (approved):

nksazonov (approved):

PR #29, opened 2025-04-22
Feat/resize

mod:
This PR mainly replaces `reset(...)` with `resize(...)` in Custody.sol, and updates a corresponding test.

`resize(...)` takes a `channelId` and a `candidate`, which contains:
- UPDATED allocations
- `CHANRESIZE` magic num AND `AMOUNTS[]` (should have a length of 2) of allocations deltas.

`resize(...)` logic:
- perform checks on a channel
- checks that `candidate` is unanimously signed
- check that `initialDeposit` + delta = `candidate.allocations`
- for each amount in delta:
    - if it is < 0, then unlocks funds from the channel
    - if it is > 0, locks more funds (if available) to the channel

### DIFF
--- a/contract/src/Custody.sol
+++ b/contract/src/Custody.sol
@@ -5,7 +5,17 @@ import {IChannel} from "./interfaces/IChannel.sol";
 import {IDeposit} from "./interfaces/IDeposit.sol";
 import {IAdjudicator} from "./interfaces/IAdjudicator.sol";
 import {IComparable} from "./interfaces/IComparable.sol";
-import {Channel, State, Allocation, Status, Signature, Amount, CHANOPEN, CHANCLOSE, CHANRESIZE} from "./interfaces/Types.sol";
+import {
+    Channel,
+    State,
+    Allocation,
+    Status,
+    Signature,
+    Amount,
+    CHANOPEN,
+    CHANCLOSE,
+    CHANRESIZE
+} from "./interfaces/Types.sol";
 import {Utils} from "./Utils.sol";
 import {IERC20} from "lib/openzeppelin-contracts/contracts/interfaces/IERC20.sol";
 import {EnumerableSet} from "lib/openzeppelin-contracts/contracts/utils/structs/EnumerableSet.sol";
@@ -387,10 +397,7 @@ contract Custody is IChannel, IDeposit {
      * @param channelId Unique identifier for the channel to resize
      * @param candidate The state with CHANRESIZE magic number and resize amounts
      */
-    function resize(
-        bytes32 channelId,
-        State calldata candidate
-    ) external {
+    function resize(bytes32 channelId, State calldata candidate) external {
         Metadata storage meta = _channels[channelId];
 
         // Verify channel exists and is ACTIVE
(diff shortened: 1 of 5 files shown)


mod:
@nksazonov but we don't need resize and close
resize alone release funds which can be withdrawn without closing the channel anymore.

contract/src/Custody.sol
```diff
@@ -361,25 +381,70 @@ contract Custody is IChannel, IDeposit {
...
+        if (meta.stage != Status.ACTIVE) revert InvalidStatus();
+
+        // Verify all participants have signed the resize state
+        // For our 2-participant channels, we need exactly 2 signatures
+        if (candidate.sigs.length != 2) revert InvalidStateSignatures();
+        if (!_verifyAllSignatures(meta.chan, candidate)) revert InvalidStateSignatures();
+
+        // Decode the magic number and resize amounts
+        (uint32 magicNumber, int256[] memory resizeAmounts) = abi.decode(candidate.data, (uint32, int256[]));
+        if (magicNumber != CHANRESIZE || resizeAmounts.length != 2) revert InvalidState();
```
mod, line 405: I suggest we make a helper which will be inlined by optimizer

contract/src/Custody.sol
```diff
@@ -361,25 +381,70 @@ contract Custody is IChannel, IDeposit {
...
+
+        uint256 initialBalanceSum = meta.actualDeposits[0].amount + meta.actualDeposits[1].amount;
+        int256 finalBalanceSum = int256(initialBalanceSum) + resizeAmounts[0] + resizeAmounts[1];
```
mod, line 408: if 0 and 1 means CREATOR and BROKER better use the const for readability.
I suggest CLIENT / ESCROW naming, the broker could be called the escrow service.

contract/src/Custody.sol
```diff
@@ -361,25 +381,70 @@ contract Custody is IChannel, IDeposit {
...
+        if (finalBalanceSum < 0) revert InvalidState();
+        uint256 candidateBalanceSum = candidate.allocations[0].amount + candidate.allocations[1].amount;
+        if (uint256(finalBalanceSum) != candidateBalanceSum) revert InvalidState();
+
+        // Process resize amounts for each participant
+        for (uint256 i = 0; i < 2; i++) {
```
mod, line 414: Maybe it doesn't require a for loop, but a small helper inlined too, the code is a bit lengthy and we need more readable

contract/test/Custody.t.sol
```diff
@@ -676,49 +676,42 @@ contract CustodyTest is Test {
...
+        Signature[] memory resizeSigs = new Signature[](2);
+        resizeSigs[0] = hostResizeSig;
+        resizeSigs[1] = guestResizeSig;
+        resizeState.sigs = resizeSigs;
 
-        // 5. Reset the channel
+        // 3. Resize the channel
         vm.prank(host);
-        custody.reset(channelId, finalState, new State[](0), newChan, newInitialState);
+        custody.resize(channelId, resizeState);
```
mod, line 703: after resize, and without closing can we test if Charlie can withdraw the full unlocked amount?

contract/src/Custody.sol
nksazonov: as `CHANRESIZE` is a constant, it will be optimized anyway. Apart from that, extracting logic to a separate function does not optimize it more, unfortunately.

contract/src/Custody.sol
nksazonov: 6ce4d7f44750d3fe041bfefb40dfc7e8fb52a0fb

contract/src/Custody.sol
nksazonov: 6ce4d7f44750d3fe041bfefb40dfc7e8fb52a0fb

PR #30, opened 2025-04-24
feat(contracts): add remittance adjudicator

nksazonov:


contract/src/adjudicators/Remittance.sol
```diff
@@ -0,0 +1,134 @@
...
+        // Identify the payer (participant whose allocation is decreasing)
+        uint256 payerIndex = identifyPayer(candidate.allocations, previousState.allocations);
+
+        // Payer must have signed the candidate state
+        if (
+            candidate.sigs.length <= payerIndex
+                || !Utils.verifySignature(stateHash, candidate.sigs[payerIndex], chan.participants[payerIndex])
+        ) {
+            return false;
+        }
```
nksazonov, line 129: I think we can simplify and require only 1 signature, not array with signature of a correct party at some index

```suggestion
        if (
            candidate.sigs.length != 1
                || !Utils.verifySignature(stateHash, candidate.sigs[0], chan.participants[payerIndex])
        ) {
            return false;
        }
```

mod (approved):

PR #31, opened 2025-04-24
Adding the field State.version

mod:


contract/src/Custody.sol
```diff
@@ -404,6 +429,9 @@ contract Custody is IChannel, IDeposit {
         if (meta.stage == Status.VOID) revert ChannelNotFound(channelId);
```
nksazonov, line 429: Add `proof` (either array or a single state) parameter to `resize(...)` to give more security guarantees.

nksazonov (approved):

PR #33, opened 2025-04-25
feat(contract): update adjudicators

nksazonov:


mod (changes requested):
- [ ] Don't export constant, it's better to redefine to avoid unecessary includes
- [ ] If we agree on good name we can redefine them in Types
- [ ] I do not agree with creating Adjudicator Utils, use Utils.sol instead

PR #36, opened 2025-04-28
refactor: nitrolite rpc, typescript sdk

akushniruk:


sdk/src/rpc/types.ts
```diff
@@ -1,85 +1,182 @@
...
+    /** An array of participant addresses (Ethereum addresses) involved in the application. */
+    participants: Hex[];
+    /** An array representing the relative weights or stakes of participants, often used for dispute resolution or allocation calculations. Order corresponds to the participants array. */
+    weights: number[];
+    /** The number of participants required to reach consensus or approve state updates. */
+    quorum: number;
+    /** A parameter related to the challenge period or mechanism within the application's protocol. */
+    challenge: number;
+    /** A unique number used once, often for preventing replay attacks or ensuring uniqueness of the application instance. */
+    nonce: number;
```
dimast-x, line 105: nonce can be optional. If its not specified, broker will take the timestamp of create_app as a nonce

sdk/src/rpc/types.ts
```diff
@@ -1,85 +1,195 @@
...
+    */
+    definition: AppDefinition;
+    /** The address of the token contract used for allocations within the application. */
+    token: Hex;
+    /** The initial allocation distribution among participants. Order corresponds to the participants array in the definition. */
+    allocation: Hex[]; // Assuming Hex representation of amounts
+    /** An array of channel IDs that this application will be associated with or funded by. */
+    channels: Hex[];
+    /** An array of addresses authorized to sign for state updates or actions within this application. */
+    signers: Hex[];
```
dimast-x, line 134: Add a note that channels and signers can be dropped from here. We don't use these on the broker side

sdk/src/client/state.ts
```diff
@@ -19,14 +18,19 @@ export async function _prepareAndSignInitialState(
     params: CreateChannelParams
 ): Promise<{ channel: Channel; initialState: State; channelId: ChannelId }> {
     const { initialAllocationAmounts, stateData } = params;
+
+    if (!stateData) {
+        throw new Errors.MissingParameterError("State data is required for creating the channel.");
+    }
```
nksazonov, line 24: As we have discussed, `stateData` can be empty (`0x`) for some Adjudicators, therefore it might make sense to have an `Adjudicator` class with methods to validate state defined.
Nevertheless, this validation is definitely done on the BE, so I am not sure what's the best approach here.

sdk/src/client/types.ts
```diff
@@ -39,20 +39,31 @@ export interface Channel {
...
+    data: Hex; // Application data encoded, decoded by the adjudicator for business logic
     allocations: [Allocation, Allocation]; // Combined asset allocation and destination for each participant
     sigs: Signature[]; // stateHash signatures
 }
 
 /**
  * Adjudicator status enum
  */
-export enum AdjudicatorStatus {
+export enum adjudicatorStatus {
```
nksazonov, line 66: If this references `ChannelStatus` Solidity enum, then it can be safely removed from SDK.

sdk/src/client/state.ts
akushniruk: We can see if it will be needed, and we can add it at any moment

sdk/src/rpc/api.ts
```diff
@@ -0,0 +1,282 @@
...
+}
+
+/**
+ * Creates the signed, stringified message body for sending a generic 'message' within an application.
+ *
+ * @param signer - The function to sign the request payload.
+ * @param appId - The Application ID the message is scoped to.
+ * @param messageParams - The actual message content/parameters being sent.
+ * @param requestId - Optional request ID.
+ * @param timestamp - Optional timestamp.
```
dimast-x, line 245: I see timestamp is optional here. On the broker side, when we create a virtual app, we use nonce. Nonce can be provided with a request, however it's not provided, we use timestamp as nonce. So make sure we have one of these params (nonce or timestamp)

dimast-x: I suggest making timestamp required, because its a core element of our rpc

akushniruk: its fine because if dev did not provide timestamp, we will generate one for him 
```
export async function createAppSessionMessage(
    signer: MessageSigner,
    params: CreateAppSessionRequest[],
    intent: Intent,
    requestId: RequestID = generateRequestId(),
    timestamp: Timestamp = getCurrentTimestamp()
): Promise<string> {
    const request = NitroliteRPC.createRequest(requestId, "create_app_session", params, timestamp, intent);

    const signedRequest = await NitroliteRPC.signRequestMessage(request, signer);

    return JSON.stringify(signedRequest);
}
```

dimast-x (approved):

PR #55, opened 2025-05-08
feat(Custody): update `checkpoint` and `challenge`

nksazonov:


docs/SECURITY.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Custody `IChannel` implementation
+
+### Challenge
+
+HORIZONTAL: what is the last known on-chain state of a channel being challenged
+VERTICAL: what is the status of a channel being challenged
+CHANOPEN: a magic number showing a state is the first deposit state
+CHANCLOSE: a magic number showing a state is final
+
```
MaxMoskalenko, line 11: FYI, it doesn't look like a list in md preview

docs/SECURITY.md
```diff
@@ -0,0 +1,185 @@
...
+```md
+- X is CHANOPEN not all joined:
+  - if (Y is CHANOPEN):
+    - verify Y is a valid CHANOPEN state (has no proof, has CHANOPEN magic number)
+      verify Y has not less signatures than X
+      verify all Y signatures are valid
+      verify all participants that supplied a signature in Y have deposited
+  - else
+    - verify all participants have deposited
+      verify adjudicate(Y, proof)
```
MaxMoskalenko, line 47: Seems like this "branch" contradicts itself by requiring: a) not all joined and b) all participants have deposited. Or deposit and join are different concepts?

docs/SECURITY.md
```diff
@@ -0,0 +1,185 @@
...
+- X is CHANOPEN all joined:
+  - if (Y is CHANOPEN):
+    - verify Y is a valid CHANOPEN state
+      verify Y contains all signatures
+      verify all Y signatures are valid
+  - else
+    - verify adjudicate(Y, proof)
+- X is operatable state:
+  - verify Y is NOT CHANOPEN
+    verify NOT isMoreRecent(X, Y) (Y it not older than X)
```
MaxMoskalenko, line 64: Just to clarify things for myself: have you added turn number to the state? The last version I remember it was supposed to be inside "appdata"

docs/SECURITY.md
```diff
@@ -0,0 +1,185 @@
...
+    verify adjudicate(Y, proof)
+```
+
+#### DISPUTE status
```
MaxMoskalenko, line 68: What's the DISPUTE status? Seems like concat of INITIAL and ACTIVE

MaxMoskalenko: If it's "challenge resolve" maybe add some conditions to validate Y and previously challenged state (Z?)

docs/SECURITY.md
nksazonov: The `case` statement for this branch states: `X is CHANOPEN not all joined`. The "not all joined" part is a characteristic of State X (can be determined by comparing number of signatures in it to number of participants in a channel). 
On the other hand, the code in question `verify all participants have deposited` references the current "on-chain situation", which may have changed (actually, should have changed as we are in this logical branch), i.e. by some party joining a channel and depositing funds.

I will clarify the fact that "not all joined" references the state (in fact, it references the channel at the moment of that state being added).

docs/SECURITY.md
nksazonov: In this revision of the codebase, the turn number is not yet added. `isMoreRecent` calls an Adjudicator (method with the same name), which should compare these states.

docs/SECURITY.md
nksazonov: `DISPUTE` means "channel is in challenge".

nksazonov: > If it's "challenge resolve" maybe add some conditions to validate Y and previously challenged state (Z?)

Could you clarify what you mean, please?

docs/SECURITY.md
nksazonov: 274ff8ee66e2c2032a8095d37d77f7dc87004c16

docs/SECURITY.md
nksazonov: 274ff8ee66e2c2032a8095d37d77f7dc87004c16

docs/SECURITY.md
MaxMoskalenko: Ok, there are several questions that arise:

1) How more than one challenge could exist at the same time?
2) There is no information regarding how to compare challenged state and checkpoint state?

nksazonov: 1. They would not co-exist, but the newer (if pass all validations) will replace the older. That's why challenging a channel in DISPUTE requires Y being newer than X.
2. Term "challenged state" means either lastValidState of a DISPUTE channel or a `candidate` in `challenge(...)`. "checkpoint state" means `candidate` in `checkpoint(...)`. If I understand correctly, you are referring to `checkpoint(...)` function where "challenged state" is X and "checkpointed" is Y.
For checkpoint to succeed in this case, Y has to be valid and newer than X.

MaxMoskalenko (approved):

contract/src/Custody.sol
```diff
@@ -11,6 +11,7 @@ import {IERC20} from "lib/openzeppelin-contracts/contracts/interfaces/IERC20.sol
 import {SafeERC20} from "lib/openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
 import {EnumerableSet} from "lib/openzeppelin-contracts/contracts/utils/structs/EnumerableSet.sol";
 
+import {console} from "forge-std/console.sol";
```
MaxMoskalenko, line 14: Remove?

nksazonov: Right, thanks!

PR #73, opened 2025-05-12
feat(channelId): add chainId to calculation

nksazonov:


dpatsora (approved):

mnaichuk (approved):

mod:
Are we sure, what's the rational? we removed it in nitro, I also can't remember the logic here

nksazonov:
> Are we sure, what's the rational? we removed it in nitro, I also can't remember the logic here

@mod , yes, this is required for the same Broker to have different channelIds cross-chain.

PR #76, opened 2025-05-19
feat: RPC interface update

akushniruk:
## v0.1.0

> **Heads-up:** This release **breaks** compatibility with any code that still relies on
> `acc` or `int`.  Make sure to follow the migration notes before upgrading.

### ✨ Added
- **`createGetChannelsMessage`** message creator for fetching channel information.  

### ⚙️ Changed
- **RPC interface**
  - Replaced top-level `acc` with `sid` (session ID) everywhere.
  - Removed the deprecated `int` (intent) field.
  - Updated `CreateAppSessionRequest`, `CloseAppSessionRequest`, and `ResizeChannel`
    to use the new field names and allocation structure.
- **Type definitions** (`sdk/src/rpc/types.ts`)
  - Deleted `Intent`.
  - Added `AppSessionAllocation`.
  - Renamed:
    - `acc` → `sid`
    - `app_id` → `app_session_id`
- **API layer** (`sdk/src/rpc/api.ts`)
  - Function signatures now match the refactored RPC schema.
  - Exports new `createGetChannelsMessage` helper.
- **RPC helpers** (`sdk/src/rpc/nitrolite.ts`)
  - Dropped all intent-related logic.
  - `create*` / `parse*` helpers now rely solely on `sid`.

### 🗑️ Removed
- All intent-specific code paths and corresponding tests.

---

### 🚀 Migration Guide

1. **Update your payloads**  
   Change every occurrence of `acc` to `sid` and delete any `int` fields.

2. **Adjust**  
   - Replace `Intent` usages with the new allocation model (`AppSessionAllocation`).  
   - Switch `app_id` → `app_session_id` in your code.

4. **Run your test suite**  
   Expect renamed arguments (`sid`) and remove intent-related assertions.

---

### DIFF
--- a/sdk/src/rpc/types.ts
+++ b/sdk/src/rpc/types.ts
@@ -8,14 +8,22 @@ export type Timestamp = number;
 
 /** Type alias for Account ID (channelId or appId) */
 export type AccountID = Hex;
-/** Represents the allocation intent change as an array of numbers. */
-export type Intent = bigint[];
 
 /** Represents the data payload within a request message: [requestId, method, params, timestamp?]. */
 export type RequestData = [RequestID, string, any[], Timestamp?];
 
 /** Represents the data payload within a successful response message: [requestId, method, result, timestamp?]. */
 export type ResponseData = [RequestID, string, any[], Timestamp?];
+
+/** Represents the allocation of assets within an application session.
+ * This structure is used to define the initial allocation of assets among participants.
+ * It includes the participant's address, the asset (usdc, usdt, etc) being allocated, and the amount.
+ */
+export type AppSessionAllocation = {
+    participant: Address;
+    asset: string;
+    amount: string;
+};
 
 /**
  * Represents the structure of an error object within an error response payload.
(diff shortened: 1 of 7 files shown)


sdk/src/rpc/types.ts
```diff
@@ -52,10 +58,10 @@ export interface NitroliteRPCMessage {
  */
 export interface ApplicationRPCMessage extends NitroliteRPCMessage {
     /**
-     * Ledger account identifier (channelId or appId). Mandatory.
+     * Application Session ID. Mandatory.
      * This field also serves as the destination pubsub topic for the message.
      */
-    acc: AccountID;
+    sid: AccountID;
```
dimast-x, line 64: AppSessionID?

sdk/src/rpc/types.ts
```diff
@@ -8,15 +8,23 @@ export type Timestamp = number;
 
 /** Type alias for Account ID (channelId or appId) */
 export type AccountID = Hex;
```
dimast-x, line 10: AppSessionID, and rename comment

sdk/src/rpc/types.ts
```diff
@@ -8,15 +8,23 @@ export type Timestamp = number;
 
 /** Type alias for Account ID (channelId or appId) */
```
dimast-x, line 9: drop "(channelId or appId)"
channelId is no longer an account. Now it's either participant address, or app session id

akushniruk: I would keep it as we don't know yet how we will use it on the dev console side

PR #79, opened 2025-05-22
Make challengeDuration optional and default to 0, aligning with type definition

v-0-0:
## Refactor: Make `challengeDuration` optional and default to 0, aligning with type definition

This PR updates the `NitroliteClient` to make the `challengeDuration` configuration parameter truly optional, defaulting to `BigInt(0)` as originally intended and documented in the type definitions.

### Problem

Previously, although the type definition for `NitroliteClientConfig` indicated `challengeDuration` as optional with a default of 0 seconds, the client constructor would throw a `MissingParameterError` if it wasn't explicitly provided. This created a discrepancy between the documented type and the actual behavior.

### Solution

- The explicit check for `challengeDuration` in the constructor has been removed.
- The `challengeDuration` property is now assigned `config.challengeDuration || BigInt(0)`, ensuring it defaults to `0` if not supplied.

This change resolves the inconsistency and aligns the client's behavior with the documented type, providing a more predictable and developer-friendly experience.

### DIFF
--- a/sdk/src/client/index.ts
+++ b/sdk/src/client/index.ts
@@ -38,7 +38,6 @@ export class NitroliteClient {
         if (!config.publicClient) throw new Errors.MissingParameterError("publicClient");
         if (!config.walletClient) throw new Errors.MissingParameterError("walletClient");
         if (!config.walletClient.account) throw new Errors.MissingParameterError("walletClient.account");
-        if (!config.challengeDuration) throw new Errors.MissingParameterError("challengeDuration");
         if (!config.addresses?.custody) throw new Errors.MissingParameterError("addresses.custody");
         if (!config.addresses?.adjudicator) throw new Errors.MissingParameterError("addresses.adjudicator");
         if (!config.addresses?.guestAddress) throw new Errors.MissingParameterError("addresses.guestAddress");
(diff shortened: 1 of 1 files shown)


nksazonov (changes requested):
@v-0-0 , thanks for you input!
Recently, we have changed the Custody contract to [accept duration of 1h (3600 seconds) at minimum](https://github.com/erc7824/nitrolite/blob/d7132a16ead77801edf580765be145d8823bc692/contract/src/Custody.sol#L139).

It would be great if you could add this constant in this file and enforce it in the creation method. Currently, SDK only supports `Custody` contract, therefore it is safe to hardcode such logic.

v-0-0:
@nksazonov, I've added validation directly to the constructor.

nksazonov:
@v-0-0 , thanks!
This PR is almost ready, only a slight change remains. Please extract the `3600n` to an explicit `CUSTODY_MIN_CHALLENGE_DURATION` to avoid using magic numbers in the code.

v-0-0:
@nksazonov I’ve added the CUSTODY_MIN_CHALLENGE_DURATION constant at the top of the file as discussed.

nksazonov:
Great thanks, @v-0-0 . I think PR is ready to be merged.

akushniruk (approved):

PR #89, opened 2025-05-26
fix: add createDeterministicPrivateKey helper

MaxMoskalenko:


sdk/src/helpers/createDeterministicPrivateKey.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export const createDeterministicPrivateKey = async (
+    walletClient: WalletClient<Transport, Chain, ParseAccount<Account>>,
+    derivationIndex: number = 0,
+    salt: string = ''
+): Promise<Hex> => {
+    const nonceMessage = `Nitrolite derivation index: ${derivationIndex}`;
+
+    const seed = await walletClient.signMessage({
+        message: nonceMessage,
```
mod, line 25: '$wallet_addr/$adj_addr/$app_addr/0/$nonce'

sdk/src/helpers/createDeterministicPrivateKey.ts
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    const input = DERIVATION_PATH_PREFIX + seed + salt;
+
+    const privateKey = keccak256(toHex(input));
```
mod, line 30: I would not run toHex, just keccak256(secret_signature) and no need to add salt in the hashing.

MaxMoskalenko: I cannot remove toHex, that's keccak256 requirement

sdk/src/helpers/createDeterministicPrivateKey.ts
```diff
@@ -0,0 +1,35 @@
...
+    type ParseAccount,
+    toHex,
+    keccak256,
+    Hex,
+} from 'viem';
+
+const DERIVATION_PATH_PREFIX = 'nitrolite_state_wallet_v1_';
+
+/**
+ * Generates a deterministic private key for wallet clients based on the provided adjudicator and app addresses, and a nonce.
```
nksazonov, line 15:
```suggestion
 * Generates a deterministic private key for wallet clients based on the provided adjudicator, app addresses, and a channel's nonce the account is used in.
```

sdk/src/helpers/createDeterministicPrivateKey.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export const createDeterministicPrivateKey = async (
+    walletClient: WalletClient<Transport, Chain, ParseAccount<Account>>,
+    adjudicatorAddress: string,
+    appAddress: string,
+    nonce: number = 0,
+): Promise<Hex> => {
+    // NOTE: keep in mind that this key should be used only for the respective adjudicator and app address
```
nksazonov, line 23:
```suggestion
    // NOTE: keep in mind that this key should be used only for the respective adjudicator, app address and a channel with the provided nonce.
```

sdk/src/helpers/createDeterministicPrivateKey.ts
```diff
@@ -0,0 +1,35 @@
...
+    const nonceMessage = `nitrolite/${adjudicatorAddress}/${appAddress}/${walletClient.account.address}/0/${nonce}`
```
nksazonov, line 24: Do we not include `chainId` by design? Right now, if this signature is compromised, then all channels cross-chain with this nonce will be compromise.

akushniruk:
It would be nice to have it, it solves an issue with the session key, which is currently stored in the localStorage
@MaxMoskalenko

sdk/src/helpers/createDeterministicPrivateKey.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Generates a deterministic private key for a session based on the wallet client,
+ * adjudicator address, app address, and a unique nonce. This key can be used
+ * for actions within a specific application context (e.g., a state channel) without
+ * requiring the user to sign every action with their main wallet.
+ *
+ * @warning This function generates a key that is valid across all chains where the
+ * adjudicator and app contracts are deployed at the same addresses. This provides a
+ * consistent session address for multi-chain dApps but removes chain-specific replay
+ * protection from the key generation itself.
```
nksazonov, line 18: Are we able to use different session private keys per chain in future?
I think, it is a good security practice.

nksazonov (approved):

sdk/src/helpers/createDeterministicPrivateKey.ts
akushniruk: Probably you're right, but it's not aligned with the idea of having one address

PR #104, opened 2025-05-28
feat(clearnode): add jwt methods and policies

MaxMoskalenko:


clearnode/auth.go
```diff
@@ -175,6 +187,70 @@ func (am *AuthManager) UpdateSession(address string) bool {
 	return true
 }
 
+func (am *AuthManager) generateJWT(address string) (string, error) {
+	claims := JWTClaims{
+		Address: address,
```
dpatsora, line 192: We need to include session key here as well

PR #107, opened 2025-05-28
Update API.md

mod:


clearnode/README.md
```diff
@@ -61,6 +61,34 @@ sequenceDiagram
...
+Application must also connect to clearnode with limited scope permissions (`app.create`) to only `create_app_session`
+for a specific application address, a maximum allocation limit and register and assign the participant session key
+with an expiration date.
+
+`create_app_session` must verify:
+
+- Policy scope allow to create app
+- Wallet in allocation match with policy wallet address
+- The application address match with the policy value
+- Allocation is within the permitted policy Allowance
```
dimast-x, line 77: “Allocation is within the permitted policy Allowance” - do we allow to create many app sessions, each within this policy allowance, or **we need to make sure that sum of allocations in all sessions of this app are less than policy allowance**? I think the second approach would be more correct

PR #109, opened 2025-05-29
fix: update metrics calculation

dpatsora:


clearnode/metrics.go
```diff
@@ -124,12 +124,16 @@ func (m *Metrics) RecordMetricsPeriodically(db *gorm.DB, custodyClients map[stri
...
-			// Refresh the list of tokens to monitor
-			monitoredTokens := GetUniqueTokenAddresses(db)
-
 			// Update metrics for each custody client
 			for _, client := range custodyClients {
-				client.UpdateBalanceMetrics(context.Background(), monitoredTokens, m)
+
+				assets, err := GetAllAssets(db, &client.chainID)
+				if err != nil {
+					log.Error(err.Error())
```
philanton, line 132: Use correct logger

PR #113, opened 2025-05-29
feat(custody): resize to wallets

nksazonov:


mnaichuk (approved):

mnaichuk:
will you add an example with a unit test?

mnaichuk (approved):

PR #114, opened 2025-05-29
ci: run contract and SDK tests

maxpushka:


.github/workflows/main-pr.yml
```diff
@@ -4,26 +4,82 @@ on:
   pull_request:
     branches: [ main ]
 
+env:
```
philanton, line 7: Let's move it to the job's env

.github/workflows/main-push.yml
philanton, line 1: We don't need tests here, it's just a redundant slow-down as we shouldn't merge PR if its tests don't pass

.github/workflows/main-pr.yml
```diff
@@ -4,26 +4,82 @@ on:
...
+        working-directory: sdk
+
+      - name: Run tests
+        run: npm test
+        working-directory: sdk
+
   build-and-publish-clearnode:
     name: Build and Publish (Clearnode)
-    needs: test-clearnode
+    needs: [test-clearnode, test-forge, test-sdk]
```
philanton, line 82: Let's publish Clearnode image even if other components are failing

PR #130, opened 2025-05-30
feat(clearnode): allow session key as participants in create app session

dimast-x:


mod:
session key should be the only participant, nothing else,
we should not accept wallet address as participant.

PR #149, opened 2025-06-03
feat: add resize-channel ADR

dpatsora:


docs/ADR-Resize-Channel-Security.md
```diff
@@ -0,0 +1,130 @@
...
+* This lock creates a verifiable on-chain commitment of total funds allocation
+* Other chains verify against this commitment before processing resizes
+* Each resize reduces the available commitment amount
+* Challenge: Requires smart contract modifications, adds cross-chain dependencies, and significantly complicates the user experience with additional transactions
+
+### Option 4: Server-Side Blockchain Submission
+* Instead of returning signed states to users, server directly submits resize transactions
+* Server tracks on-chain confirmations before allowing additional resizes
+* Server manages transaction queuing, gas prices, and retries
+
```
dimast-x, line 57: Option 5 (Louis suggestion):
`We don't need nitrolite or channels to withdraw funds
The broker can simply send out an ERC20 transfer `

dimast-x: I don't think channel states would require changes. As validations for close or resize channel are based on unified balance. However, I am ok with implementing server-side resize first, and then considering options like ERC20 transfer to simplify the flow in the future

docs/ADR-Resize-Channel-Security.md
```diff
@@ -0,0 +1,130 @@
...
+* System Integrity: Ensure balance accounting remains accurate
+* Technical Feasibility: Implement with minimal changes to existing contracts
+* Operational Overhead: Consider the impact on server operations
+
+## Considered Options
+
+### Option 1: Timeout-Based Fund Locking
+* Lock funds in the database when a resize is requested
+* Set a timeout to release locks if not confirmed on-chain
+* Challenge: The smart contract has no notion of state expiration, so users could still submit expired states
```
nksazonov, line 38: We can add expiration logic to the Adjudicator contract - this would not change Custody and is rather an expected development. However, such feature requires adding timestamp to appdata.

docs/ADR-Resize-Channel-Security.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Option 3: Cross-Chain Lock Commitments
+* Modify smart contracts to require a "lock" transaction on a primary chain
+* This lock creates a verifiable on-chain commitment of total funds allocation
+* Other chains verify against this commitment before processing resizes
+* Each resize reduces the available commitment amount
+* Challenge: Requires smart contract modifications, adds cross-chain dependencies, and significantly complicates the user experience with additional transactions
+
+### Option 4: Server-Side Blockchain Submission
+* Instead of returning signed states to users, server directly submits resize transactions
```
nksazonov, line 54: `resize` state still needs to be unanimously signed, therefore the flow is the following:
- server composes the state
- server sends it to the user to be signed
- user sends the signature back 
- server countersigns and sends tx to chain

docs/ADR-Resize-Channel-Security.md
```diff
@@ -0,0 +1,130 @@
...
+**We want to implement Option 4: Server-Side Blockchain Submission for channel resize operations.**
+
+This approach offers the strongest security guarantees while maintaining a good user experience. The server will directly submit resize transactions to the blockchain after validating user requests, eliminating the window for double-spending.
+
+After thorough analysis, we determined that:
+- Option 1 (Timeout-Based Fund Locking) fails because signed states remain valid indefinitely
+- Option 2 (State Versioning) cannot prevent double-spending without smart contract modifications
+- Option 3 (Cross-Chain Lock Commitments) would work but requires significant contract changes and degrades UX
+
+Server-side submission (Option 4) addresses the vulnerability without contract modifications by ensuring users never receive multiple valid states they could submit independently.
```
nksazonov, line 68: I think this should be reevaluated according to my [this](https://github.com/erc7824/nitrolite/pull/149/files#r2125941563) and [this](https://github.com/erc7824/nitrolite/pull/149/files#r2125963442) comment.

docs/ADR-Resize-Channel-Security.md
nksazonov: I also think that this approach is worse then (1), because we hide valid states from users (no data availability). If server goes down, user will not be able to withdraw funds (because they do not possess the `resize` state). Moreover, if we commit to data availability now or in the future, we will still need to implement the `getLatestSupportedState` on the broker side. If we go with (4), this function would need to have an exception and do NOT send `resize` state if it is the latest, which introduces inconsistency.

docs/ADR-Resize-Channel-Security.md
nksazonov: I still think that in such a case we need to consider managing channel states.
Transferring funds outside the channel can be fine, BUT:
(i) we will need to track such move in a channel somehow, so that user is not able to challenge the channel with the state BEFORE an outside-channel transfer.
(ii) not in case of a challenge. Again, if the broker goes down, user has no way to receive their funds.

These are main reasons why funds transfer should happen inside the channel.

PR #166, opened 2025-06-05
feat: add RPCMessage signers

dpatsora:


clearnode/rpc.go
```diff
@@ -15,22 +15,37 @@ type RPCMessage struct {
...
+	recoveredAddresses := make([]string, len(r.Sig))
+	for i, sigHex := range r.Sig {
+		recovered, err := RecoverAddress(r.Req.rawBytes, sigHex)
+		if err != nil {
+			return nil, err
+		}
+		recoveredAddresses[i] = recovered
+	}
+
+	return recoveredAddresses, nil
```
dimast-x, line 28: better storing as a map so that we can avoid duplications. Current upproach may create a vulnerabilily in quorum where user may send multiple duplicated signatures to achieve quorum weight

PR #170, opened 2025-06-06
refactor: logger and context propagation

philanton:


clearnode/signer.go
```diff
@@ -29,10 +29,6 @@ func NewSigner(privateKeyHex string) (*Signer, error) {
 		return nil, err
 	}
 
-	publicKey := privateKey.Public().(*ecdsa.PublicKey)
-
-	log.Printf("Broker signer initialized with address: %s", crypto.PubkeyToAddress(*publicKey).Hex())
```
dpatsora, line 34: This is still useful, please put it back

PR #171, opened 2025-06-06
feat: improve clearnet_rpc_requests_total metric

philanton:


clearnode/ws.go
```diff
@@ -478,26 +485,32 @@ func forwardMessage(rpc *RPCMessage, msg []byte, fromAddress string, h *UnifiedW
...
-			}
+		messageBodyBytes, err := json.Marshal(rpc)
+		if err != nil {
+			messageBody = "could not marshal rpc message"
+		} else {
+			messageBody = string(messageBodyBytes)
 		}
+
 		log.Printf("error: %s, wallet: %s, method: %s, request: %s", errMsg, sender, rpc.Req.Method, messageBody)
+		reqMethod = rpc.Req.Method
```
dpatsora, line 507: There is a lot of cases, where rpc is nil, so this line will panic

PR #175, opened 2025-06-09
feat(custody): challenge, checkpoint

nksazonov:


contract/src/Custody.sol
```diff
@@ -310,54 +293,64 @@ contract Custody is IChannel, IDeposit {
...
+                    }
+                }
+            } else if (lastValidStateIntent == StateIntent.OPERATE) {
+                if (candidate.intent != StateIntent.OPERATE) revert InvalidState();
+                if (!Utils.statesAreEqual(candidate, meta.lastValidState)) {
+                    if (!_isMoreRecent(meta.chan.adjudicator, candidate, meta.lastValidState)) revert InvalidState();
+                    if (!IAdjudicator(meta.chan.adjudicator).adjudicate(meta.chan, candidate, proofs)) {
+                        revert InvalidState();
+                    }
+                }
```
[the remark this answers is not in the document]
nksazonov: Timer can not be reset as it is only active when the channel is in `Dispute` stage, which is filtered out at the start of the `challenge` function.

contract/src/Custody.sol
```diff
@@ -310,54 +293,64 @@ contract Custody is IChannel, IDeposit {
...
-        ) {
-            // if no state data or intent is not INITIALIZE or RESIZE, assume this is a normal state
+            _closeEffectsAndInteractions(channelId, candidate.allocations);
 
-            // Verify the state is valid according to the adjudicator
-            if (!IAdjudicator(meta.chan.adjudicator).adjudicate(meta.chan, candidate, proofs)) revert InvalidState();
+            emit Challenged(channelId, block.timestamp);
+            emit Closed(channelId, candidate);
+            return;
+        }
```
MaxMoskalenko, line 314: Nothing to change here, but this part feels a little bit odd. The only way to "close" the channel, that wasn't joint by anyone is to issue challenge. And then broadcast Challenged and Closed events all together

That's just.. odd, and feels like something conceptual is missing here

contract/src/Custody.sol
```diff
@@ -310,54 +293,64 @@ contract Custody is IChannel, IDeposit {
...
+        // main goal: verify Candidate is valid and >= LastValidState (in RESIZE case states should be equal)
+        if (lastValidStateIntent == StateIntent.INITIALIZE) {
+            if (candidate.intent == StateIntent.INITIALIZE) {
+                if (!Utils.statesAreEqual(candidate, meta.lastValidState)) revert InvalidState();
+            } else {
+                if (!_isMoreRecent(meta.chan.adjudicator, candidate, meta.lastValidState)) revert InvalidState();
+                if (!IAdjudicator(meta.chan.adjudicator).adjudicate(meta.chan, candidate, proofs)) {
+                    revert InvalidState();
+                }
             }
```
MaxMoskalenko, line 326: Is it possible to send INITILIZE state in non-INITIAL channel, that would be more recent than the previous (presumably the first one)?

MaxMoskalenko: Also, what's the point in not reverting states that are equal - it's just brning gas without changing SC memory state?

contract/src/Custody.sol
```diff
@@ -310,54 +293,64 @@ contract Custody is IChannel, IDeposit {
...
+        } else if (lastValidStateIntent == StateIntent.OPERATE) {
+            if (candidate.intent != StateIntent.OPERATE) revert InvalidState();
+            if (!Utils.statesAreEqual(candidate, meta.lastValidState)) {
+                if (!_isMoreRecent(meta.chan.adjudicator, candidate, meta.lastValidState)) revert InvalidState();
+                if (!IAdjudicator(meta.chan.adjudicator).adjudicate(meta.chan, candidate, proofs)) {
+                    revert InvalidState();
+                }
+            }
+        } else {
+            // lastValidStateIntent == StateIntent.RESIZE
```
MaxMoskalenko, line 336: Same here, maybe it would be better to specify state intent here

contract/src/Custody.sol
```diff
@@ -369,41 +362,42 @@ contract Custody is IChannel, IDeposit {
...
-        if (meta.stage == ChannelStatus.DISPUTE) {
-            meta.stage = ChannelStatus.ACTIVE;
             meta.challengeExpire = 0;
         }
 
+        // effects
+        meta.stage = ChannelStatus.ACTIVE;
+        meta.lastValidState = candidate;
+
         emit Checkpointed(channelId);
```
MaxMoskalenko, line 401: Maybe add turn number (and perhaps state hash) to the Checkpointed and Challenged events? It should be useful

contract/src/Utils.sol
```diff
@@ -63,6 +63,38 @@ library Utils {
         return recoveredSigner == signer;
     }
 
+    /**
+     * @notice Compares two states for equality
+     * @param a The first state to compare
+     * @param b The second state to compare
+     * @return True if the states are equal, false otherwise
+     */
+    function statesAreEqual(State memory a, State memory b) internal pure returns (bool) {
```
MaxMoskalenko, line 72: What's more gas efficient: compare keccaks of the states or compare field by field?

contract/src/Custody.sol
nksazonov: `challenge(...)` usage is forced here as it has (will have in future update) a `challengerSig` param that requires "the approval" of the challenge from one of the participants.
If we are to implement such feature with `close(...)`, then anyone would be able to close any channel before clearnode joins.

Moreover, `challenge(...)` is conceptual for single-party-initiated close, which is the case in such situation.

However, I agree with you on the fact that consecutive `Challenged(...)` and `Closed(...)` seem a little odd, nevertheless we are able to tweak this behaviour in the future.

contract/src/Custody.sol
nksazonov: > Is it possible to send INITILIZE state in non-INITIAL channel, that would be more recent than the previous (presumably the first one)?

If Adjudicator is implemented correctly, than it should not be possible to do so. However, we would like to protect from such case on the protocol level, and thus such checks.
Moreover, `StateIntent` and `ChannelStatus` enums logic partially overlap one another, which causes such odd checks. This should be fixed in future with combining these enums and / or making some states implicit, i.e. defined by some channel storage variables.

nksazonov: > Also, what's the point in not reverting states that are equal - it's just brning gas without changing SC memory state?

Channel still records a challenge if states are equal. The logic is different because if challenged and last valid state are equal, we don't need to call `adjudicate(candidate)`, and thus we save gas.

contract/src/Custody.sol
nksazonov: 8655a27012187f6d247c706367cd5f1a8b804eae

contract/src/Custody.sol
nksazonov: 4f0369c385e79a678e5ec41b16271652a7d944b9

contract/src/Custody.sol
nksazonov: I agree that your suggested approach is more flexible for channels with more than 2 participants.

docs/SECURITY.md
```diff
@@ -0,0 +1,208 @@
+# This document outlines protocol-level security mechanisms
+
+## Custody `IChannel` implementation
+
+### Push-down automaton
+
+Security of the custody implementation is based on a push-down automaton (PDA) that defines the channel lifecycle.
+The PDA is defined on the following image:
+
+![Custody PDA](./img/CustodyPDA.png)
```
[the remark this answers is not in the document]
MaxMoskalenko: Ah, ok. I haven't noticed the special condition on the arrao from challenge to close. Then it makes sense

PR #183, opened 2025-06-09
feat: store contract events

dpatsora:


clearnode/eth_listener.go
```diff
@@ -111,6 +153,103 @@ func listenEvents(
...
+	contractAddress common.Address,
+	chainID uint32,
+	currentBlock uint64,
+	lastBlock uint64,
+	lastIndex uint32,
+	backOffCount *atomic.Uint64,
+	historicalCh chan types.Log,
+	logger Logger,
+) {
+	const blockStep = 10000
```
mnaichuk, line 167: maybe make sense to make it configaruble, for example for world chain, right now we select up 500-1000 blocks

mnaichuk (changes requested):

mnaichuk:
@dimast-x please address comments

clearnode/eth_listener.go
dimast-x: done

clearnode/contract_event.go
```diff
@@ -0,0 +1,106 @@
...
+	Data            datatypes.JSON `gorm:"column:data"`
+	CreatedAt       time.Time      `gorm:"column:created_at"`
+}
+
+func (ContractEvent) TableName() string {
+	return "contract_events"
+}
+
+func StoreContractEvent(tx *gorm.DB, event *ContractEvent) error {
+	// Skip if the event has already been processed
```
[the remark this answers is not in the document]
dimast-x: done

clearnode/custody.go
```diff
@@ -409,18 +409,28 @@ func (c *Custody) handleResized(logger Logger, ev *nitrolite.CustodyResized) {
...
+		newAmount := new(big.Int).SetUint64(channel.Amount)
 		for _, change := range ev.DeltaAllocations {
-			newAmount += change.Int64()
+			newAmount.Add(newAmount, change)
 		}
 
-		channel.Amount = uint64(newAmount)
+		if newAmount.Sign() < 0 {
+			// TODO: what do we do in this case?
+			logger.Error("invalid resize, channel balance cannot be negative", "channelId", channelID)
```
dimast-x, line 419: I think we need to ensure this on the contract level so that there is no such scenario

mnaichuk: we can't isolate logic only on a contract, because here we unified balance across all chains, so we just need to make sure to do not have negative values and notify about them

mnaichuk (approved):

PR #187, opened 2025-06-10
feat(contract): IChainReader interface

nksazonov:


contract/src/interfaces/IChannelReader.sol
```diff
@@ -0,0 +1,41 @@
...
+
+import {Channel, State, ChannelStatus, Amount} from "./Types.sol";
+
+interface IChannelReader {
+    /**
+     * @notice Get the list of open channels for a list of accounts
+     * @param accounts Array of account addresses to check for open channels
+     * @return Array of arrays, where each inner array contains channel IDs for the corresponding account
+     */
+    function getOpenedChannels(address[] memory accounts) external view returns (bytes32[][] memory);
```
dimast-x, line 12: Lets call it getOpenChannels()? as we use this conventions in clearnode

dimast-x (approved):

contract/src/interfaces/IChannelReader.sol
nksazonov: 50425846fed3f1af0d175ffc9cb0fc5bfab4904a

PR #199, opened 2025-06-12
feat: fetch batch contract metrics

dimast-x:


dimast-x:
@gemini-code-assist review

clearnode/custody.go
```diff
@@ -449,86 +452,105 @@ func (c *Custody) UpdateBalanceMetrics(ctx context.Context, assets []Asset, metr
...
+			batch[i] = rpc.BatchElem{
+				Method: "eth_getBalance",
+				Args:   []interface{}{brokerAddr.Hex(), "latest"},
+				Result: nativeRes[i],
 			}
-			walletBalance = decimal.NewFromBigInt(walletBalanceRaw, -int32(asset.Decimals))
-
 		} else {
-			caller, err := nitrolite.NewErc20(tokenAddr, c.client)
+			data, err := erc20Abi.Pack("balanceOf", brokerAddr)
```
mnaichuk, line 488: why not reuse balance_checker for all erc20s in one rpc call and get balance of native token in another rpc ?
imagine we have 1000 erc20s per chain in a assets

and FYI - I've added a TODO with it and seems like you deleted it

mnaichuk (changes requested):

clearnode/custody.go
dimast-x: Maybe you overlooked but on this line it does not call the contract separately and we dont pass client. It appends the call into a batch request that then executed once

dimast-x: "the whole []rpc.BatchElem gets sent in a single JSON-RPC request. Under the hood Geth will marshal your slice of BatchElem into one HTTP (or IPC) call, and then unpack all N results from that one batch response. There’s no per-token round trip"

PR #208, opened 2025-06-18
feat: add ledger transfer method

dimast-x:


mnaichuk (approved):

dpatsora (approved):

dimast-x:
@mod Good point. Yes, in the next iterations we can either create a separate method transfer_from, or simply add an optional field “from” to trigger this use case

PR #211, opened 2025-06-19
feat: prevent using wallets as signers and signers as wallets

dpatsora:


clearnode/wallet.go
```diff
@@ -50,8 +50,31 @@ func AddSigner(db *gorm.DB, wallet, signer string) error {
 	}
 
 	return db.Transaction(func(tx *gorm.DB) error {
-		var existingSigner SignerWallet
```
dimast-x, line 53: These checks can be done in cache without calling db like it is done a few lines upper

clearnode/wallet.go
```diff
@@ -50,8 +50,31 @@ func AddSigner(db *gorm.DB, wallet, signer string) error {
...
+		var count int64
+		if err := tx.Model(&SignerWallet{}).Where("wallet = ?", signer).Count(&count).Error; err != nil {
+			return err
+		}
+		if count > 0 {
+			return fmt.Errorf("cannot use a wallet as a signer")
+		}
+
+		// Address can't be used as a wallet if it is already a signer for another wallet.
+		if err := tx.Model(&SignerWallet{}).Where("signer = ?", wallet).Count(&count).Error; err != nil {
```
dimast-x, line 69: second check is already done line 45 through cache

dpatsora: It's a different check

PR #215, opened 2025-06-20
feat: `get` requests pagination

nksazonov:


clearnode/asset.go
```diff
@@ -27,14 +27,16 @@ func GetAssetByToken(db *gorm.DB, tokenAddress string, chainID uint32) (Asset, e
...
-func GetAllAssets(db *gorm.DB, chainID *uint32) ([]Asset, error) {
+func GetAllAssets(db *gorm.DB, chainID *uint32, params *PaginationParams) ([]Asset, error) {
 	var assets []Asset
 	query := db.Model(&Asset{})
 
 	if chainID != nil {
 		query = query.Where("chain_id = ?", *chainID)
 	}
 
+	query = paginate(params)(query)
```
dimast-x, line 38: add check for nil. If params == nil, then do not paginate

clearnode/handlers.go
```diff
@@ -911,6 +911,7 @@ func HandleGetRPCHistory(policy *Policy, rpc *RPCMessage, store *RPCStore) (*RPC
 // HandleGetAssets returns all supported assets
 func HandleGetAssets(rpc *RPCMessage, db *gorm.DB) (*RPCMessage, error) {
 	var chainID *uint32
+	var page, pageSize *int32
```
dimast-x, line 914: assign a default value (like 1st page and 100 entries). If pagination params are not provided - use default ones

clearnode/handlers.go
```diff
@@ -1002,3 +1015,46 @@ func checkChallengedChannels(tx *gorm.DB, wallet string) error {
...
+	DefaultPageSize = 10
+	MaxPageSize     = 100
+)
+
+type PaginationParams struct {
+	Page     uint32
+	PageSize uint32
+}
+
+func validatePaginationParams(page, pageSize *int32) *PaginationParams {
```
dimast-x, line 1030: method is called validatePaginationParams however it not only validates but also transforms them

clearnode/handlers.go
```diff
@@ -911,6 +911,7 @@ func HandleGetRPCHistory(policy *Policy, rpc *RPCMessage, store *RPCStore) (*RPC
 // HandleGetAssets returns all supported assets
 func HandleGetAssets(rpc *RPCMessage, db *gorm.DB) (*RPCMessage, error) {
 	var chainID *uint32
+	var page, pageSize *int32
```
dimast-x, line 915: also, "order" (asc/desc) param would be useful

clearnode/asset.go
maxpushka: `paginate()` implementation handles this case, so I see not sense in adding those conditions explicitly if the goal is developing a generic pagination experience

nksazonov:
Outdated and migrated to #217

PR #223, opened 2025-06-24
feat: introduce channel service

dimast-x:


clearnode/channel_service.go
```diff
@@ -0,0 +1,263 @@
...
+			TokenAddress: alloc.Token.Hex(),
+			Amount:       alloc.Amount,
+		})
+	}
+
+	return resp, nil
+}
+
+// checkChallengedChannels checks if the participant has any channels in the challenged state.
+func checkChallengedChannels(tx *gorm.DB, wallet string) error {
```
philanton, line 253: If it won't be used anywhere except ChannelService, I'd suggest declaring it as its method

dpatsora (approved):

PR #224, opened 2025-06-25
feat: improve RPC handlers' logs

philanton:


dimast-x:
Suggesting basing your PR on https://github.com/erc7824/nitrolite/pull/223 that includes ChannelService and AppSessionService

PR #229, opened 2025-06-26
fix: add missing fields to JWT

philanton:


clearnode/rpc_router_auth.go
```diff
@@ -264,7 +264,7 @@ func (r *RPCRouter) handleAuthSigVerify(ctx context.Context, sig string, authPar
 		return nil, nil, "failed to create signer in db"
 	}
 
-	claims, jwtToken, err := r.AuthManager.GenerateJWT(challenge.Address, challenge.SessionKey, "", "", challenge.Allowances)
+	claims, jwtToken, err := r.AuthManager.GenerateJWT(challenge.Address, challenge.SessionKey, challenge.Scope, challenge.AppName, challenge.Allowances)
```
dpatsora, line 267: Add TODO: to use expiration specified in the Policy, instead of just setting 1 hour

PR #243, opened 2025-07-01
feat: add user tag

dpatsora:


clearnode/docs/API.md
```diff
@@ -239,6 +240,8 @@ To get balance in a specific virtual app session, specify `app_session_id` as ac
 
 This method allows a user to transfer assets from their unified balance to another account. The user must have sufficient funds for each asset being transferred. The operation will fail if any of the specified assets have insufficient funds.
 
+User may specify the `destination` (wallet address) or `destination_tag` (user tag) to identify the recipient. If both are provided, `destination` takes precedence.
```
alessio, line 243:
```suggestion
User may specify the `destination` (wallet address) or `destination_tag` (user tag) to identify the recipient. `destination` is used if and only if the `destination_tag` field is empty.
```

dimast-x:
The approach is good for me, was thinking about similar logic but calling this feature `user_aliaces`.. having user `tags` is also okay

clearnode/rpc_router_auth.go
```diff
@@ -264,6 +264,13 @@ func (r *RPCRouter) handleAuthSigVerify(ctx context.Context, sig string, authPar
 		return nil, nil, "failed to create signer in db"
 	}
 
+	// Generate the User tag
+	_, err = GenerateOrRetrieveUserTag(r.DB, challenge.Address)
+	if err != nil {
```
alessio, line 269: For brevity's sake, I'd go with:

```suggestion
	if _, err := GenerateOrRetrieveUserTag(r.DB, challenge.Address); err != nil {
```

Just a suggestion, definitely not a blocker 👍

clearnode/rpc_router_private.go
```diff
@@ -199,13 +204,15 @@ func (r *RPCRouter) HandleTransfer(c *RPCContext) {
 	}
 
 	// Allow only ledger accounts as destination at the current stage. In the future we'll unlock application accounts.
-	if params.Destination == "" || params.Destination == c.UserID || !common.IsHexAddress(params.Destination) {
+	switch {
+	case params.Destination == "" && params.DestinationTag == "":
+		c.Fail("either destination or destination_tag must be provided")
```
alessio, line 209: The error message is slightly misleading. We know that both could be provided. It boils down to the word `either`; instead of using the `aut aut` form (i.e. XOR), we should use the `vel` one (i.e. OR):

```suggestion
		c.Fail("destination or destination_tag must be provided")
```

clearnode/rpc_router_private.go
```diff
@@ -199,13 +204,15 @@ func (r *RPCRouter) HandleTransfer(c *RPCContext) {
...
+		return
+	case params.Destination != "" && !common.IsHexAddress(params.Destination):
```
alessio, line 211: The following should suffice I think?

```suggestion
	case !common.IsHexAddress(params.Destination):
```

clearnode/rpc_router_private_test.go
```diff
@@ -153,6 +153,109 @@ func TestRPCRouterHandleGetRPCHistory(t *testing.T) {
 	assert.Equal(t, uint64(1), rpcHistory[2].ReqID, "Third record should be the oldest")
 }
 
+func TestRPCRouterHandleGetUserTag(t *testing.T) {
+	// Create signers
```
alessio, line 157: I'd be curious to see what would happen if we added the following:

```suggestion
	t.Parallel()

	// Create signers
```

clearnode/docs/API.md
```diff
@@ -261,6 +264,25 @@ Currently, `Transfer` supports ledger account of another user as destination (wa
   }], 1619123456789],
   "sig": ["0x9876fedcba..."]
 }
+
+// OR
+
+{
+  "req": [1, "transfer", [{
+    "destination_tag": "UX123D8C", // User tag
```
alessio, line 272: If we decide to go with userTag instead of tag, then this should be changed into something like:

```suggestion
    "destination_usertag": "UX123D8C", // User tag
```

clearnode/rpc_router_private.go
```diff
@@ -17,8 +17,13 @@ import (
 )
 
 type Transfer struct {
-	Destination string               `json:"destination"`
-	Allocations []TransferAllocation `json:"allocations"`
+	Destination    string               `json:"destination"`
+	DestinationTag string               `json:"destination_tag"`
```
alessio, line 21: As per above:

```suggestion
	DestinationUserTag string               `json:"destination_user_tag"`
```

clearnode/tag_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_GetUserTagByWallet(t *testing.T) {
+	db, cleanup := setupTestDB(t)
+	defer cleanup()
```
alessio, line 12: I'd rather use `t.Cleanup()` here, please see https://pkg.go.dev/testing#T.Cleanup

clearnode/tag_test.go
```diff
@@ -0,0 +1,59 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
alessio, line 6: Why using `assert` in the first place? It's generally a good idea for a test case to fail as early as possible

PR #248, opened 2025-07-03
feat: add README.md to integration tests

MaxMoskalenko:


integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+Additionally it will run migrations on database and seed it with initial data.
+
+## Setup SDK
+
+Integration tests require the sdk to be built in respective directory (`sdk` or `../sdk` from this file)
+
+To set up the SDK, navigate to the `sdk` directory and run:
+
+```bash
+npm install
```
nksazonov, line 37: Maybe it is better to use `npm ci`, as in GH workflow?

integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+In order to run the integration tests, you need components to integrate with: DB, Clearnode (BE) and blockchain node. All of these three components can be run using Docker.
+
+Go to the root of the project (`..` from this file) and run:
+
+```bash
+docker-compose up -Vd
+```
+
+In will build clearnode from the current version of the code (you could use docker image instead, but it may not include the latest changes). Compile smart contracts and deploy them on predetermined addresses (based on mnemonic and unique salt):
```
nksazonov, line 19: Please specify that `salt` is a number specified in parenthesis after the contract name.

integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+It's possible to use specific version of the sdk but you will need to change the `package.json` with a specific version of `@erc7824/nitrolite` and change import path in the `tsconfig.json`
+
+## Running Integration Tests
+
+To run the integration tests, ensure you have the necessary environment set up as described above. Then, navigate to the `integration` directory and execute the following command:
+
+```bash
+npm install
```
nksazonov, line 48: Let's use `npm ci` here as well

nksazonov (approved):
Important documentation. It may need some changes.

PR #256, opened 2025-07-07
feat: add wallet tags in returned transactions, update export tool and add tests

dimast-x:
https://www.notion.so/yellow-net/df4e8c6c0c734eac96b3f93bf1ce809c?v=67d1b10ec29f42b983ee011f28317c2b&p=22375b8feb04802f882eedbdf1e5dcca&pm=s


akushniruk (approved):
JS SDK side is ok, need review golang

clearnode/export_transactions_cli.go
```diff
@@ -3,66 +3,140 @@ package main
...
+	if err := e.ExportToCSV(file, options); err != nil {
+		return "", fmt.Errorf("failed to export to CSV: %w", err)
+	}
+
+	return fileName, nil
+}
+
+func runExportTransactionsCli(logger Logger) {
+	logger = logger.NewSystem("export-transactions")
+	if len(os.Args) < 3 {
```
alessio, line 96: Shouldn't the case `len(os.Args) > 5` return a wrong n-arguments kind of error as well?

clearnode/export_transactions_cli_test.go
```diff
@@ -0,0 +1,165 @@
...
+	_, err = RecordLedgerTransaction(db, TransactionTypeTransfer, NewAccountID(account1), NewAccountID(account2), "usdc", decimal.NewFromInt(100))
+	require.NoError(t, err)
+
+	_, err = RecordLedgerTransaction(db, TransactionTypeDeposit, NewAccountID(account2), NewAccountID(account1), "eth", decimal.NewFromInt(50))
+	require.NoError(t, err)
+
+	_, err = RecordLedgerTransaction(db, TransactionTypeAppDeposit, NewAccountID(account1), NewAccountID(account3), "usdc", decimal.NewFromInt(25))
+	require.NoError(t, err)
+
+	t.Run("Export", func(t *testing.T) {
```
alessio, line 40: How about try to run the test cases in parallel?

```suggestion
	t.Parallel()
	
	t.Run("Export", func(t *testing.T) {
```

If something blows up, we may want to take a look at how the fixtures are initialized then (and quite possibly do so in the context of a separate work item/follow-up PR)

dimast-x: updated https://github.com/erc7824/nitrolite/pull/256/commits/a6c4b23c3c3d0bb25c68ce1d073b4758597844ed

PR #257, opened 2025-07-07
feat(custody): add EIP-191, EIP-712 support

nksazonov:


nksazonov:
Also another problem has arose, which relates to EIP-712 requiring some `domainSeparator`, that is expected to be unique per contract (e.g. "contract name", "contract version"). However, during state signature verification, adjudicator has nowhere to get this `domainSeparator` from.

This can be resolved in several ways:
- hardcoding Custody's `domainSeparator` into each Adjudicator... Very blunt way.
- hardcoding Custody's address into each Adjudicator and reading `domainSeparator` dynamically. Better, but still means if Custody is changed, Adjudicators must be redeployed.
- adding a dynamic `custodyAddress` field to each Adjudicator alongside RBAC for Owner, who is able to change the address mentioned. During signature validation, Adjudicator calls `custodyAddress` for the `domainSeparator`.
- adding a `channelImpl` field to `Channel` struct, that must match the address of the Custody. This enables Adjudicators to extract the Custody's address and fetch the `domainSeparator` dynamically. This, however, is a pretty drastic change to the protocol, which means all components must be updated. Nevertheless, IMHO, this is the most "philosophical" solution.

mnaichuk:
> Also another problem has arose, which relates to EIP-712 requiring some `domainSeparator`, that is expected to be unique (e.g. "contract name", "contract version") per contract that a user has signed the message for. However, during state signature verification, adjudicator has nowhere to get this `domainSeparator` from.
> 
> This can be resolved in several ways:
> 
> * hardcoding Custody's `domainSeparator` into each Adjudicator... Very blunt way.
> * hardcoding Custody's address into each Adjudicator and reading `domainSeparator` dynamically. Better, but still means if Custody is changed, Adjudicators must be redeployed.
> * adding a dynamic `custodyAddress` field to each Adjudicator alongside RBAC for Owner, who is able to change the address mentioned. During signature validation, Adjudicator calls `custodyAddress` for the `domainSeparator`.
> * extracting signature verification to an external `SignatureVerificator` contract, which `Custody` and Adjudicators will call to verify state signatures.
> * adding a `channelImpl` field to `Channel` struct, that must match the address of the Custody. This enables Adjudicators to extract the Custody's address and fetch the `domainSeparator` dynamically. This, however, is a pretty drastic change to the protocol, which means all components must be updated. Nevertheless, IMHO, this is the most "philosophical" solution.

I'm good with "adding a dynamic `custodyAddress` field to each Adjudicator alongside RBAC for Owner"

nksazonov:
On a decision-making call (@alessio @mnaichuk @dpatsora @nksazonov) we have came with consensus about continuing with option 3.

> 3. adding a dynamic custodyAddress field to each Adjudicator alongside RBAC for Owner, who is able to change the address mentioned. During signature validation, Adjudicator calls custodyAddress for the domainSeparator.

Option 5 was discarded as we actually want protocol States to be chain-agnostic.

---

Also we discussed that:
- we don't want to add any checks for the number of `allocations` in the State yet
- nor do we want to remove the `IComparable` interface

alessio (approved):
ACK. Please wait for at least another approval before merging.

mnaichuk (approved):

mnaichuk:
good for me @nksazonov

PR #267, opened 2025-07-10
fix(sdk): align transfer response params to the updated backend definition

maxpushka:


sdk/src/rpc/parse/misc.ts
```diff
@@ -45,27 +46,34 @@ const ErrorParamsSchema = z
...
+            z
+                .object({
+                    id: z.number(),
+                    tx_type: txTypeEnum,
+                    from_account: addressSchema,
+                    from_account_tag: z.string().optional(),
+                    to_account: addressSchema,
+                    to_account_tag: z.string().optional(),
+                    asset: z.string(),
+                    amount: z.union([z.string(), z.number()]).transform((a) => a.toString()),
```
akushniruk, line 59: u can use z.string(), as amount always in string

maxpushka: [b65c63a](https://github.com/erc7824/nitrolite/pull/267/commits/b65c63adb3d976959990631298ef5784673800f5)

PR #285, opened 2025-07-17
refactor: API protocol

MaxMoskalenko:
This PR was partially started as a fix of this issue https://github.com/erc7824/nitrolite/issues/277. But there is even a bigger issue in the current API protocol: there are no strict standards for params field. Therefore right now `params` field could contain string, array, object or any other JSON-supported data structure.

In my opinion, it makes  further code maintenance and developing harder. As adding new field to non-object structure (e.g. add `metadata` object to array structure in `get_channels` resp.) will require changes on sdk side and could cause type and compilation errors on product (that uses nitrolite) side. Also support backward compatibility for these changes would create huge amount of legacy code (aka code debt). So it would be nice to solve this issue ASAP

PS. is there are any incentives to wrap params in array, could we simplify it by making: `[id, "message", params{}, ts]` instead of `[id, "message", [params{}], ts]` ?


clearnode/docs/Clearnode.protocol.md
```diff
@@ -1,18 +1,21 @@
 # Clearnode Protocol Specification
```
MaxMoskalenko, line 1: @philanton , actually, yes I would like to merge it after implementing changes to the node and sdk. Also, a question to you as a person that worked on RPCNode, what do you think about using object for params instead of an array (currently I'm fixing `rpc_node_test` and it's quite painful)

clearnode/docs/API.md
```diff
@@ -273,38 +282,46 @@ Supports pagination and sorting.
...
+        "protocol": "NitroSnake",
+        "challenge": 86400,
+        "weights": [70, 30],
+        "quorum": 100,
+        "version": 1,
+        "nonce": 123456790
+      }
+    ],
+    "metadata": {
+      "page": 5,
```
mnaichuk, line 319: before we were using middleware that sets this info into respective headers

mnaichuk:
1. I don't see any reason to wrap params in an array, so this change is good for me
2. I'm not sure we need an extra key for the responses, as inthe  comment above, I've suggested another common way to share pagination info with the response

mnaichuk:
and I think API changes should be more collaborative, and we need @akushniruk opinion here too, not only BE guys

nksazonov:
Please,
- [ ] rebase on #299 now
- [ ] rebase on #299 once "`join(...)` simplification" is merged

PR #303, opened 2025-08-05
test(custody::join): broker auto-join tests

nksazonov:


contract/test/Custody.t.sol
```diff
@@ -509,67 +510,246 @@ contract CustodyTest_challenge is CustodyTest_Base {
...
+        initialState.sigs = sigs;
+
+        vm.prank(hostSK);
+        vm.expectRevert(Custody.InvalidChallengePeriod.selector);
+        custody.create(chan, initialState);
     }
 
-    function test_depositAndCreate() public {
-        // 1. Prepare channel and initial state
+    function test_create_brokerAutoJoin_zeroAllocation_rawECDSA() public {
```
dimast-x, line 631: brokerAutoJoin_zeroAllocation may be an outdated test name, as now we auto join regardless of zero allocation. The requirement is not zero allocation but having initial state signed by all participants

dimast-x: I see, so we currently actually allow opening channel automatically. Then maybe we should harden the validation? Like keeping auto opening only if all participants signed initial state

nksazonov: I guess the test name is not self-explanatory enough and causes a confusion, because the `brokerAutoJoin` part was intended to mean that 2 signatures were supplied.
I will rename this part to `brokerAutoJoinWhen2Sigs`

dimast-x: What's about broker2SigsAutoJoin? There is still a point that in the testcase we name number two while our smart contract is not limiting number of participants (its only clearnode requirement)

dimast-x: test_create_autoJoinWithSignatures?
whats about this one

nksazonov: In fact, Custody does limit the number of participants to 2. However you are correct and I think we should avoid mentioning that not to create another logical dependency.
WDYT about `brokerAutoJoinWithTheirSig`? I want to avoid mentioning "allSignatures"

dimast-x: we can shorten it to brokerAutoJoinWithSigs
or even autoJoinWithSigs, as we don't necessarily have to mention broker here

nksazonov: c8026844ae65d93c7663dcbd055e926605787d2f

contract/test/Custody.t.sol
```diff
@@ -509,67 +510,246 @@ contract CustodyTest_challenge is CustodyTest_Base {
...
-        assertEq(channelCount, 1, "Host should have 1 channel");
+        verifyChannelActive([hostSK, guestSK], channelId, chan);
+        verifyDeposited(hostSK, address(token), DEPOSIT_AMOUNT);
+        verifyDeposited(guestSK, address(token), 0); // No funds have been allocated to guest, so they keep their full deposit
+    }
 
-        // Also check that the channelId is consistent
-        bytes32 expectedChannelId = Utils.getChannelId(chan);
-        assertEq(channelId, expectedChannelId, "Channel ID is incorrect");
+    function test_create_brokerAutoJoin_zeroAllocation_EIP191() public {
```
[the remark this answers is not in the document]
nksazonov: c8026844ae65d93c7663dcbd055e926605787d2f

PR #309, opened 2025-08-19
YNU-41: chain-agnostic signing package and Interface

dimast-x:


clearnode/pkg/sign/ethereum/signer.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	addr, err := recoverAddressFromHash(typedDataHash, sig)
+	if err != nil {
+		return "", err
+	}
+	return addr.Hex(), nil
+}
+
+// RecoverAddress is an Ethereum-specific function to recover an address from a standard message signature.
+func RecoverAddress(message []byte, sig sign.Signature) (string, error) {
```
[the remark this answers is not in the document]
dimast-x: https://github.com/erc7824/nitrolite/pull/309/commits/4b547194e4309b0df0cb4f9f0e7c38c0670c3fdb

clearnode/pkg/sign/ethereum/README.md
```diff
@@ -0,0 +1,72 @@
...
+    fmt.Println("Signer Address:", signer.Address())
+}
+```
+
+
+### 2. Signing Data
+
+The `Sign` method follows the standard Ethereum practice of first hashing the data with Keccak256.
+
+```go
```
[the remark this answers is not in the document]
dimast-x: https://github.com/erc7824/nitrolite/pull/309/commits/b29c6c9151a42a56d56d9a387cf77075ac0ea63f

clearnode/pkg/sign/sign.go
```diff
@@ -0,0 +1,63 @@
...
+
+// String implements the fmt.Stringer interface
+func (s Signature) String() string {
+	return hexutil.Encode(s)
+}
+
+// Address is an interface for a blockchain-specific address.
+type Address interface {
+	fmt.Stringer // All addresses must have a string representation.
+}
```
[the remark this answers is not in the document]
dimast-x: Added Equals(address) https://github.com/erc7824/nitrolite/pull/309/commits/f088b8c22e5fa30dc9cb52cba617c161cf77a0b0
Not sure if we need Compare (we never used it in our codebase). I suggest adding this in case we actually need to use it

clearnode/pkg/sign/example_test.go
philanton, line 1: Maybe move it inside `ethereum` package?

clearnode/pkg/sign/sign.go
```diff
@@ -7,6 +7,29 @@ import (
...
+// Signer is an interface for a blockchain-agnostic signer.
+type Signer interface {
+	PrivateKey() PrivateKey // Private key implements PrivateKey interface.
+}
+
+// PrivateKey is an interface for a blockchain-agnostic private key.
+type PrivateKey interface {
+	PublicKey() PublicKey
+	Sign(data []byte) (Signature, error) // Sign generates a signature for the given data.
+	Bytes() []byte
```
philanton, line 24: I don't really like the idea of having it.. For example, if we have a KMS implementation, it won't give us ability to get the private key. Also, in my opinion, it enables vulnerability to use it in the code somewhere, where ideally we shouldn't print or use the private key for the purposes different from signing.

philanton: That's why I suggested to move `PublicKey()` and `Sign()` function on the level of `Signer`

clearnode/pkg/sign/sign.go
```diff
@@ -12,6 +12,11 @@ type Signer interface {
 	PrivateKey() PrivateKey // Private key implements PrivateKey interface.
 }
 
+// AddressRecoverer is an optional interface that signers can implement
+type AddressRecoverer interface {
+	RecoverAddress(message []byte, signature Signature) (string, error)
```
philanton, line 17: Should return an `Address` instead of string, no?

clearnode/pkg/sign/example_test.go
dimast-x: https://github.com/erc7824/nitrolite/pull/309/commits/fd0d66f9351e073b7f73c59d22344190e478ef2c

clearnode/pkg/sign/sign.go
dimast-x: https://github.com/erc7824/nitrolite/pull/309/commits/fd0d66f9351e073b7f73c59d22344190e478ef2c

clearnode/pkg/sign/sign.go
dimast-x: https://github.com/erc7824/nitrolite/pull/309/commits/fd0d66f9351e073b7f73c59d22344190e478ef2c

clearnode/pkg/sign/ethereum/signer.go
philanton, line 1: Would be good to have constructors for PublicKey and Address as well

dimast-x: https://github.com/erc7824/nitrolite/pull/309/commits/46d5e675b4983feec242858d1759a5cbfa34ef21

clearnode/pkg/sign/ethereum/signer.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	// Adjust V from 0/1 to 27/28 for Ethereum compatibility.
+	if sig[64] < 27 {
+		sig[64] += 27
+	}
+	return sig, nil
+}
+
+// RecoverAddress implements the AddressRecoverer interface.
+func (s *Signer) RecoverAddress(message []byte, signature sign.Signature) (sign.Address, error) {
```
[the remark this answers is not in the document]
dimast-x: Regarding signature interface, here is feedback from Claude:
`Pros of making it an interface:
  1. More flexible - different signature implementations could have different internal representations
  2. Could encapsulate signature-specific logic like public key recovery
  3. Follows the pattern of other types in the package (PublicKey, Address, Signer are all interfaces)

  Cons:
  1. Breaking change - all existing code using Signature as a byte slice would break
  2. The current Signature already has useful methods (JSON marshaling, String representation)
  3. Signatures are often just byte arrays in most cryptographic contexts
  4. Adding RecoverPublicKey() might not make sense for all signature types - some might not support key recovery
  5. The current AddressRecoverer interface already provides address recovery functionality

  I should provide a balanced view considering the current design.

Looking at the current code, I have mixed thoughts about making Signature an interface:

  Pros:
  - Consistent with other types (PublicKey, Address, Signer are interfaces)
  - Could encapsulate signature-specific recovery logic
  - More extensible for different signature schemes

  Cons:
  - Breaking change - current Signature []byte works well and has useful JSON/String methods
  - RecoverPublicKey() might not apply to all signature types
  - The existing AddressRecoverer interface already handles recovery at the appropriate level
  - Signatures are fundamentally byte arrays in most crypto contexts
 
  The current design separates concerns well - signatures are data, recovery is behavior handled by the appropriate
  recoverer. Unless you need signature-type-specific behavior beyond byte representation, the current approach seems cleaner.
`

clearnode/pkg/sign/sign.go
```diff
@@ -35,6 +35,36 @@ type Address interface {
...
+	default:
+		return "Unknown"
+	}
+}
+
+// Alg returns the algorithm used for this signature based on its length and structure.
+func (s Signature) Alg() Algorithm {
+	if len(s) == 65 {
+		// Standard Ethereum signature format (r: 32 bytes, s: 32 bytes, v: 1 byte)
+		return AlgorithmKeccak256
```
philanton, line 63: That is ECDSA, Keccak256 is a hashing algorithm, it shouldn't be listed here :)

philanton: But actually I think we will have one algorithm per platform, so either types here should have the names of platforms (aka package names) or package names should be named as algorithms.

I personally don't like the idea of renaming `ethereum` package to `ecdsa`, maybe just `eth` instead. And then the method name here would be either `Type()` or `Platform()` instead of `Alg()`, because each sign/verify depends not only on algorithm (for example not each ECDSA signature has `v` part incremented with 27)

clearnode/pkg/sign/sign.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil
+}
+
+// String implements the fmt.Stringer interface
+func (s Signature) String() string {
+	return hexutil.Encode(s)
+}
+
+// EthereumRecoverer implements AddressRecoverer for Ethereum signatures.
+type EthereumRecoverer struct{}
```
philanton, line 90: We already have AddressRecoverer for ethereum defined inside `eth` package, so we should use it.. But if you import it here the circular dependency will appear, because `eth` package is already importing types defined here - meaning we will need at least one other package..

philanton (changes requested):

clearnode/pkg/sign/sign.go
dimast-x: ab4964aec0c335d6662f8eaede987aa975111cdd

clearnode/pkg/sign/sign.go
```diff
@@ -85,27 +85,3 @@ func (s *Signature) UnmarshalJSON(data []byte) error {
...
-	switch sigType {
-	case TypeEthereum:
-		return &EthereumRecoverer{}, nil
-	default:
-		return nil, fmt.Errorf("unsupported signature type: %s", sigType.String())
-	}
-}
-
-// NewAddressRecovererFromSignature creates an AddressRecoverer based on signature algorithm detection.
-func NewAddressRecovererFromSignature(signature Signature) (AddressRecoverer, error) {
```
philanton: We still need such a generic function defined in this package somewhere

dimast-x: b49b999295c356458312c39b97cf57e06f1b29db

alessio (approved):
Sounds good. We'll merge it for now and we'll await @nksazonov's review and comments will be incorporated in a follow-up PR, should Nikita see objections/raise suggestions.

nksazonov:
@coderabbitai review

clearnode/pkg/sign/doc.go
```diff
@@ -0,0 +1,41 @@
...
+//
+//	// Create a new Ethereum signer from a hex-encoded private key
+//	signer, err := sign.NewEthereumSigner(privateKeyHex)
+//	if err != nil {
+//	    log.Fatal(err)
+//	}
+//
+//	// Sign a message (provide hash, not raw message)
+//	message := []byte("hello world")
+//	hash := ethcrypto.Keccak256Hash(message)
```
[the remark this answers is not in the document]
nksazonov: Also, if the signing is standardized with the interface, this can be said about hashing. Creating a message hash would be fully dependant on the type of the signature, and thus it will nullify the benefit of the standardized "sign" interface.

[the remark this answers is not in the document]
alessio: > semantically "signing" should be performed over the "message", not the "message hash"

Yes and no. Cryptographically speaking, a signature is indeed generated given a key and "data". A message has is in fact binary data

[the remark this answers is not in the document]
alessio: I like the interface because it is very simple and it works for anything as simple and raw as a string of bytes

clearnode/pkg/sign/README.md
```diff
@@ -0,0 +1,20 @@
...
+* **Blockchain-Agnostic Interfaces**: Defines a standard set of interfaces for cryptographic operations.
+* **EVM Implementation**: Includes a ready-to-use implementation for Ethereum and other EVM-compatible chains.
+* **Easily Extensible**: Simple to add support for new blockchains like Solana, Bitcoin, etc.
+* **Type-Safe**: Provides distinct types for `Address`, `Signature`, `PublicKey`, and `PrivateKey`.
+
+## Usage
+
+See the Go package documentation and examples by running:
+```bash
+go doc -all [github.com/erc7824/nitrolite/clearnode/pkg/sign](https://github.com/erc7824/nitrolite/clearnode/pkg/sign)
```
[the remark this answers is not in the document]
alessio: Did you check this renders correctly?

clearnode/pkg/sign/sign.go
```diff
@@ -0,0 +1,102 @@
...
+func (t Type) String() string {
+	switch t {
+	case TypeEthereum:
+		return "Ethereum"
+	default:
+		return "Unknown"
+	}
+}
+
+// Type returns the signature type for this signature based on its length and structure.
```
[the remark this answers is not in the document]
alessio: If that turns out to be case, we could use the same name. I don't see a problem with that

PR #322, opened 2025-08-26
feat: introduce package for RPC protocol

philanton:


alessio (approved):
Memo for ourselves: @philanton and the team agreed to further refactor this and move it into the upper level `rpc` protocol.

Having said that, this looks good to me: well done 💯

PR #327, opened 2025-08-27
feat: sandbox deployment

philanton:


clearnode/chart/config/sandbox/secrets.yaml
```diff
@@ -0,0 +1,9 @@
+config:
+  database:
+    password: ref+tfstategs://terraform-state-deploy/gke-uat-postgresql-admin/default.tfstate/output.postgresql_user_passwords["clearnet_sandbox_admin"]
+  secretEnvs:
+    BROKER_PRIVATE_KEY: ref+gcpsecrets://ynet-stage/clearnet-sandbox-broker-private-key?version=latest
+    POLYGON_AMOI_INFURA_URL: ref+gcpsecrets://ynet-stage/clearnet-sandbox-polygon-amoi-infura-url?version=latest
```
nksazonov, line 6:
```suggestion
    POLYGON_AMOY_INFURA_URL: ref+gcpsecrets://ynet-stage/clearnet-sandbox-polygon-amoy-infura-url?version=latest
```

And rename the GCP secret and in the Clearnode config as well

clearnode/chart/config/sandbox/clearnode.yaml
```diff
@@ -0,0 +1,68 @@
...
+  requests:
+    cpu: 100m
+    memory: 256Mi
+    ephemeral-storage: 100Mi
+
+autoscaling:
+  enabled: false
+
+networking:
+  externalHostname: sandboxnet.yellow.com
```
nksazonov, line 58: IMHO `sandboxnet` sounds strange. Are we sure we want to stick to the `...net` suffix rule?

nksazonov (approved):
Generally LGTM, with some clarifications and nit picks.

clearnode/chart/config/sandbox/clearnode.yaml
nksazonov: If we want to always use "...net", then we can do `clearnet-sandbox.yellow.com`

PR #332, opened 2025-09-09
YNU-44: export an RPC Client Implementation

philanton:


clearnode/pkg/rpc/client_manual_test.go
```diff
@@ -0,0 +1,305 @@
...
+	for !dialer.IsConnected() {
+		select {
+		case err := <-errCh:
+			if err != nil {
+				t.Fatalf("connection error: %v", err)
+			}
+		default:
+			time.Sleep(100 * time.Millisecond)
+		}
+	}
```
alessio, line 118: This smells really unidiomatic... We should look into it at some point and try redesign the flow

alessio (approved):
Great progress, well done. ACK

PR #334, opened 2025-09-12
feat(clearnode): minor improvements for checkpoint flow

dimast-x:
## Summary by CodeRabbit

- Refactor
  - Standardized channel IDs and transaction hashes to Ethereum 32-byte hashes across public APIs, libraries and tests.
  - Renamed public action/type/status identifiers and updated public signatures to reflect new types.
  - Introduced configurable worker constants (batch size, tick interval) and reduced startup/shutdown log verbosity.

- Bug Fixes
  - Hardened retry handling with a max-retry non-retry failure path and added a non-retry fail option.
  - Improved checkpoint unmarshalling error handling and more reliable completion using native hash types.

### DIFF
--- a/clearnode/blockchain_action.go
+++ b/clearnode/blockchain_action.go
@@ -80,6 +80,13 @@ func (a *BlockchainAction) Fail(tx *gorm.DB, err string) error {
 	return tx.Save(a).Error
 }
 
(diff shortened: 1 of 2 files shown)


alessio:
@coderabbitai review

PR #375, opened 2025-09-30
YNU-274: introduce NitroRPC/0.4 protocol version

philanton:
## Summary by CodeRabbit

* **New Features**
  * Added NitroRPC v0.4 support: submit_app_state now requires an intent (operate/deposit/withdraw) and a version; v0.2 remains supported. SDK and client helpers updated to be protocol-aware/generic.

* **Bug Fixes**
  * v0.4 validation runs earlier; duplicate version increment removed; allocation errors now prefixed with "incorrect app state:"; unsupported protocols return errors.

* **Tests**
  * New unit and integration tests for NitroRPC v0.4 covering success, invalid-version, lifecycle, balance updates, and cleanup.

* **Documentation**
  * API docs updated with v0.2 and v0.4 submit_app_state examples and semantics.


dimast-x (approved):

clearnode/pkg/rpc/api.go
```diff
@@ -569,6 +576,18 @@ type AppAllocation struct {
 	Amount decimal.Decimal `json:"amount"`
 }
 
+// AppSessionIntent indicates the purpose of an application state update.
+type AppSessionIntent string
+
+const (
+	// AppSessionIntentOperate is for normal application operation
+	AppSessionIntentOperate AppSessionIntent = "operate"
```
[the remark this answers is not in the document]
dimast-x: I meant that current submit_app_state embodies checkpoint logic itself. Or do you envision any specific logic for checkpoint?
(but overall I am okay with keeping intent as Operate)

nksazonov (approved):
Generally LGTM with some improvements.

clearnode/rpc_router_private.go
```diff
@@ -39,9 +40,11 @@ type CreateAppSessionParams struct {
 }
 
 type SubmitAppStateParams struct {
-	AppSessionID string          `json:"app_session_id"`
-	Allocations  []AppAllocation `json:"allocations"`
-	SessionData  *string         `json:"session_data"`
+	AppSessionID string               `json:"app_session_id"`
+	Intent       rpc.AppSessionIntent `json:"intent"`
```
dimast-x, line 44: lets make intent and version optional to keep backward support for v0.2 version

philanton: That's just the whole request structure, it is the responsibility of the client or SDK to indicate what us optional, so for go-sdk here I will just explain in the comments

clearnode/pkg/rpc/api.go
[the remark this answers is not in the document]
philanton: Then keeping Operate

clearnode/app_session_service_test.go
```diff
@@ -292,6 +293,100 @@ func TestAppSessionService_SubmitAppState(t *testing.T) {
...
+		// Mock ledger balances for the app session
+		require.NoError(t, GetWalletLedger(db, userAddressB).Record(sessionAccountID, "usdc", decimal.NewFromInt(0)))
+
+		rpcSigners := map[string]struct{}{
+			userAddressA.Hex(): {},
+			userAddressB.Hex(): {},
+		}
+
+		_, err := service.SubmitAppState(params, rpcSigners)
+		require.Equal(t, RPCErrorf("invalid version: expected %d, got %d", 2, params.Version), err)
```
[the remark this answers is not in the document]
philanton: [`e22ee30`](https://github.com/erc7824/nitrolite/pull/375/commits/e22ee3037aaf8bf6d998472b285163f1c6985ec6)

clearnode/app_session_service.go
```diff
@@ -133,6 +133,25 @@ func (s *AppSessionService) SubmitAppState(params *SubmitAppStateParams, rpcSign
...
+		case rpc.VersionNitroRPCv0_4:
+			if newVersion != params.Version {
+				return RPCErrorf("invalid version: expected %d, got %d", newVersion, params.Version)
+			}
+
+			switch params.Intent {
+			case rpc.AppSessionIntentOperate:
+				// no additional actions needed
+			default:
+				return RPCErrorf("unsupported intent: %s", params.Intent)
```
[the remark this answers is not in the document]
philanton: [`e22ee30`](https://github.com/erc7824/nitrolite/pull/375/commits/e22ee3037aaf8bf6d998472b285163f1c6985ec6)

clearnode/rpc_router_private.go
philanton: [`e22ee30`](https://github.com/erc7824/nitrolite/pull/375/commits/e22ee3037aaf8bf6d998472b285163f1c6985ec6)

dimast-x (approved):

sdk/src/rpc/api.ts
```diff
@@ -16,13 +16,14 @@ import {
...
     RPCChannelStatus,
+    ProtocolVersion,
 } from './types';
 import { NitroliteRPC } from './nitrolite';
 import { generateRequestId, getCurrentTimestamp } from './utils';
 import {
     CloseAppSessionRequestParams,
     CreateAppSessionRequestParams,
-    SubmitAppStateRequestParams,
+    SubmitAppStateParamsByProtocol,
```
dimast-x, line 26: I would still call it SubmitAppStateRequestParams. ByProtocol could be used if we had 2 options like normal and by protocol. As there is only one params object for this endpoint. we better keep SubmitAppStateRequestParams

sdk/src/rpc/types/request.ts
```diff
@@ -341,10 +336,46 @@ export type GetUserTagRequestParams = GetUserTagRequest['params'];
...
+    /** The new allocation distribution among participants. Must include all participants and maintain total balance. */
+    allocations: RPCAppSessionAllocation[];
+    /** Optional session data as a JSON string that can store application-specific state or metadata. */
+    session_data?: string;
+};
+
+/**
+ * Represents the request parameters for the 'submit_app_state' RPC method (NitroRPC/0.4).
+ */
+export type SubmitAppStateRequestParamsV04 = {
```
dimast-x, line 354: we could use same set of params for both protocols. In case 0.2 is used, the other fields would be just skipped (or null values would be assigned). Like in GRPC, when protocol evolves and new fields are added, they are added on the bottom but the request is reused for backwards compatibility.

nksazonov: For the BE part I agree, but this is dev-facing code, thus I would explicitly mention which fields should be used in which version

clearnode/rpc_router_private.go
```diff
@@ -39,9 +40,11 @@ type CreateAppSessionParams struct {
 }
 
 type SubmitAppStateParams struct {
```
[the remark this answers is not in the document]
philanton: No, intent is protocol specific, such a log entry should be printed only inside AppSessionService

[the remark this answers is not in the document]
nksazonov: Oh, right. Will it be possible to filter logs for requests with a specific intent still?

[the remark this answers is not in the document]
philanton: I doubt that it will be added for requests specifically, but for calls to AppSessionService - yes, once a granular switch-case depending on protocol and intent is implemented

sdk/src/rpc/types/request.ts
```diff
@@ -8,6 +8,8 @@ import {
     RPCAppSessionAllocation,
     RPCAllowance,
     GetLedgerTransactionsFilters,
+    RPCAppStateIntent,
```
philanton, line 11: On BE side it is called `AppSessionIntent`, I'm not sure whether it is better name than this, but let's have the same for TS and Golang

sdk/src/rpc/types/request.ts
```diff
@@ -8,6 +8,8 @@ import {
...
+    ProtocolVersion,
```
philanton, line 12: Why some types have `RPC` prefix and some don't?

integration/tests/lifecycle_nitrorpc_v02.test.ts
```diff
@@ -0,0 +1,546 @@
...
+    parseCloseChannelResponse,
+    parseCreateAppSessionResponse,
+    parseGetAppSessionsResponse,
+    parseGetLedgerBalancesResponse,
+    parseResizeChannelResponse,
+    parseSubmitAppStateResponse,
+    ProtocolVersion,
+} from '@erc7824/nitrolite';
+import { Hex, parseUnits } from 'viem';
+
```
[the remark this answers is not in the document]
dimast-x: It's still not very clear why do we have two lifecycle test files if we can combine it

sdk/src/rpc/types/request.ts
nksazonov: I will rename `ProtocolVersion` to `RPCProtocolVersion`.

sdk/src/rpc/types/request.ts
philanton: Agh, okay..

sdk/src/rpc/types/request.ts
nksazonov: 91fc916ec2c652bc84aa7a066eba91d110c1e00f

sdk/src/rpc/api.ts
nksazonov: 91fc916ec2c652bc84aa7a066eba91d110c1e00f

PR #414, opened 2025-11-04
YNU-408: deny outcoming offchain ops for users with non-zero channels

philanton:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added pre-checks that deny deposits, transfers, and session creation when a wallet has non-zero allocations in open channels, with clear denial messages.

* **New Features**
  * Aggregation-based guard enforces zero allocations before ledger operations.
  * Channel flows moved from a top-up pattern to explicit resize operations including allocation intent.

* **Tests**
  * Expanded unit and integration tests covering non-zero-allocation denials, resize-before-close flows, transfer rejection, and ledger balance validations.

### DIFF
--- a/clearnode/channel.go
+++ b/clearnode/channel.go
@@ -121,7 +121,7 @@ type ChannelAmountSum struct {
 func GetChannelAmountSumByWallet(tx *gorm.DB, senderWallet string) (ChannelAmountSum, error) {
 	var result ChannelAmountSum
 	err := tx.Model(&Channel{}).
-		Select("COUNT(channel_id) as count, SUM(raw_amount) as sum").
+		Select("COUNT(channel_id) as count, COALESCE(SUM(CAST(raw_amount AS NUMERIC)), 0) as sum").
 		Where("wallet = ? AND status IN (?, ?)", senderWallet, ChannelStatusOpen, ChannelStatusResizing).
 		Scan(&result).Error
 	if err != nil {


clearnode/channel.go
```diff
@@ -114,3 +114,22 @@ func CheckExistingChannels(tx *gorm.DB, wallet, token string, chainID uint32) (*
 
 	return &channel, nil
 }
+
+type ChannelAmountSum struct {
+	Count int             `gorm:"column:count"`
+	Sum   decimal.Decimal `gorm:"column:sum"`
+}
+
+func GetChannelAmountSumByWallet(tx *gorm.DB, senderWallet string) (ChannelAmountSum, error) {
```
nksazonov: Could you explain why have you created a separate structure, and not just returning 2 separate values?

clearnode/rpc_router_private.go
```diff
@@ -833,3 +837,14 @@ func verifySigner(rpc *RPCMessage, channelWallet string) error {
 	}
 	return nil
 }
+
+func ensureWalletHasZeroChannelAllocation(tx *gorm.DB, wallet string) error {
```
nksazonov: Also, the function name implies we want to ensure that the wallet has no allocations  or zero allocation only in one channel.
I think the following name provides a better gist of what this function does.

```suggestion
func ensureWalletHasAllAllocationsEmpty(tx *gorm.DB, wallet string) error {
```

clearnode/rpc_router_private_test.go
```diff
@@ -699,6 +734,50 @@ func TestRPCRouterHandleCreateAppSession(t *testing.T) {
 
 		assertErrorResponse(t, ctx, "has challenged channels")
 	})
+	t.Run("ErrorNonZeroChannelAllocation", func(t *testing.T) {
+		t.Parallel()
+
+		router, db, cleanup := setupTestRPCRouter(t)
+		t.Cleanup(cleanup)
+
+		for i, p := range []string{userAddressA.Hex(), userAddressB.Hex()} {
```
nksazonov: Why have you created a channel for each user and not only for one?

clearnode/app_session_service.go
```diff
@@ -429,6 +433,10 @@ func (s *AppSessionService) CreateAppSession(params *CreateAppSessionParams, rpc
 				return err
 			}
 
+			if err := ensureWalletHasZeroChannelAllocation(tx, walletAddress); err != nil {
+				return err
+			}
```
nksazonov: Currently the check resides inside the "allocations loop", meaning if a participant does NOT have an allocation on channel creation, they may be included in the app session successfully even if they have a non-empty allocation channel.

However, this is mitigated by the same restriction inside the "deposit" app state submission, ensuring a user with non-empty channel can NOT allocate their funds elsewhere.

Still, this may be worth our attention as current logic introduces inconsistency, because a user with a non-empty channel can join an app session.

clearnode/rpc_router_private_test.go
```diff
@@ -699,6 +734,50 @@ func TestRPCRouterHandleCreateAppSession(t *testing.T) {
 
 		assertErrorResponse(t, ctx, "has challenged channels")
 	})
+	t.Run("ErrorNonZeroChannelAllocation", func(t *testing.T) {
```
nksazonov: Why is this test duplicated here and in `app_session_service_test.go`?

integration/tests/close_channel.test.ts
```diff
@@ -67,7 +73,36 @@ describe('Close channel', () => {
...
+        expect(String(resizeResponseParams.state.allocations[1].amount)).toBe('0');
+
+        const {txHash: resizeChannelTxHash} = await client.resizeChannel({
+            ...composeResizeChannelParams(
+                resizeResponseParams.channelId as Hex,
+                resizeResponseParams,
+                state,
+            ),
+        });
+        expect(resizeChannelTxHash).toBeDefined();
```
nksazonov: You can use `TestNitroClient.topUpChannel(...)` for that, but it obviously needs to be renamed and tweaked a bit.

integration/tests/lifecycle_nitrorpc_v04.test.ts
```diff
@@ -353,9 +386,12 @@ describe('nitrorpc_v04 lifecycle', () => {
 
         const resizeReceipt = await blockUtils.waitForTransaction(resizeChannelTxHash);
         expect(resizeReceipt).toBeDefined();
+
+        const postCloseAccountBalance = await bobClient.getAccountBalance(CONFIG.ADDRESSES.USDC_TOKEN_ADDRESS);
+        expect(postCloseAccountBalance).toBe(toRaw(finalBobAmount)); // 1125
     });
 
-    it('should close channel and withdraw with app funds', async () => {
+    it('should close Bob channel', async () => {
```
nksazonov: Do I understand correctly that you have decided to showcase that 2 funds withdrawal cases are possible:
1. resize + close (Alice channel)
2. just close (Bob channel)
?

clearnode/channel.go
```diff
@@ -114,3 +114,22 @@ func CheckExistingChannels(tx *gorm.DB, wallet, token string, chainID uint32) (*
...
+	var result ChannelAmountSum
+	err := tx.Model(&Channel{}).
+		Select("COUNT(channel_id) as count, SUM(raw_amount) as sum").
+		// TODO: maybe we should filter by ChannelStatusResizing as well
```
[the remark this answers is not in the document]
dimast-x: maybe dont filter by status at all, or just in case check the ones with " status NOT ChannelStatusClosed:

clearnode/app_session_service_test.go
```diff
@@ -276,6 +276,44 @@ func TestAppSessionService_CreateApplication(t *testing.T) {
...
+				Nonce:              uint64(time.Now().Unix()),
+			},
+			Allocations: []AppAllocation{
+				{Participant: userAddressA.Hex(), AssetSymbol: "usdc", Amount: decimal.NewFromInt(100)},
+			},
+		}
+
+		_, err := service.CreateAppSession(params, rpcSigners(userAddressA))
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "operation denied: non-zero allocation in 1 channel(s) detected. Please, un-allocate those funds via resize first")
```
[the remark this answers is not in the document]
philanton: Yes, I was going to discuss it with you, as this error is only applicable for handling `transfer` requests

clearnode/rpc_router_private_test.go
philanton: I created it for "just-in-case", but I agree, we can remove one of them

integration/tests/close_channel.test.ts
philanton: Do as you wish, I don't want to dive deeper there))0)

clearnode/channel.go
nksazonov: I am just not sure it provides more benefits then disadvantages.
On one side the structure is an object that hosts all the relevant fields and it makes sense to create it when we need to include a lot of fields or it is passed a lot.
However here it only includes 2 fields and is not passed around a lot.
That is why I am not confident it is the best to include it here.
Maybe you could explain more why is it cleaner on both sides?

clearnode/rpc_router_private_test.go
nksazonov: Let's remove this one then.

clearnode/rpc_router_private.go
philanton: [3f5595b](https://github.com/erc7824/nitrolite/pull/414/commits/3f5595baa0b08bfcb751aec47ec3245d04930119)

clearnode/app_session_service_test.go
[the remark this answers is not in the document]
philanton: [3f5595b](https://github.com/erc7824/nitrolite/pull/414/commits/3f5595baa0b08bfcb751aec47ec3245d04930119)

clearnode/rpc_router_private_test.go
philanton: [3f5595b](https://github.com/erc7824/nitrolite/pull/414/commits/3f5595baa0b08bfcb751aec47ec3245d04930119)

clearnode/channel.go
```diff
@@ -112,3 +112,21 @@ func CheckExistingChannels(tx *gorm.DB, wallet, token string, chainID uint32) (*
...
+	Count int             `gorm:"column:count"`
+	Sum   decimal.Decimal `gorm:"column:sum"`
+}
+
+func GetChannelAmountSumByWallet(tx *gorm.DB, senderWallet string) (ChannelAmountSum, error) {
+	var result ChannelAmountSum
+	err := tx.Model(&Channel{}).
+		Select("COUNT(channel_id) as count, SUM(raw_amount) as sum").
+		Where("wallet = ? AND status IN (?, ?)", senderWallet, ChannelStatusOpen, ChannelStatusResizing).
+		Scan(&result).Error
```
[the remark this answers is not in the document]
alessio: This needs addressing

nksazonov (approved):

PR #607, opened 2026-03-06
feat(cerebro): allow to start without preset wallet account

nksazonov:
## Release Notes

* **Bug Fixes**
  * Improved error handling for disconnected node states with proper validation and early exits to prevent errors when attempting operations without an active connection.

* **New Features**
  * Enabled certain commands to operate without requiring an active wallet connection.
  * Added automatic connection attempt after wallet import with status messaging.

### DIFF
--- a/cerebro/commands.go
+++ b/cerebro/commands.go
@@ -143,6 +143,10 @@ func (o *Operator) showConfig(ctx context.Context) {
 	}
 
 	// Node info
+	if o.client == nil {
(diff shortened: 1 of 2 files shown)


dimast-x:
On the first launch (if nothing is set) you can generate user a new wallet (also useful to add export private key into a file helper). If user wants to import some existing wallet, he would be able to do it anyway, but this way user has all functionality unlocked and visible at the first launch. We planned to do this and other improvements later

dimast-x (changes requested):
described in comment

PR #663, opened 2026-04-08
M-I02: normalize input hex addresses

philanton:
## Description

The clearnode API accepts wallet addresses without the `0x` prefix. `common.IsHexAddress` validates format but accepts both `"0xabc123..."` (42 chars) and `"abc123..."` (40 chars) as valid. Both representations decode to the same 20-byte address via `common.HexToAddress`, producing identical channel IDs, signatures, and on-chain behavior — but they are treated as different strings in the database.

The signature verification in `ChannelSigValidator.Verify` (`pkg/core/channel_signer.go:213`) compares the recovered address (always `"0x..."`-prefixed via `EthereumAddress.String()`) against the wallet using `strings.EqualFold`. A no-prefix wallet will always fail this comparison, causing the entire database transaction to roll back.

However, this is an **accidental** guard — the rejection happens at signature validation (line 150), not at input sanitization (line 27). The code between those lines (channel ID computation, existing channel lookup, state advancement validation) all execute with a wallet string that doesn't match the database's expected format before the transaction is rolled back.

## Summary by CodeRabbit

**Bug Fixes**
- Improved Ethereum address validation with clearer, more detailed error messages across API endpoints for session and channel operations.
- Standardized address formatting for consistent handling across the platform.

**Tests**
- Added comprehensive test coverage for address validation and normalization functionality.

### DIFF
--- a/clearnode/api/app_session_v1/submit_session_key_state.go
+++ b/clearnode/api/app_session_v1/submit_session_key_state.go
@@ -4,10 +4,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 
 	"github.com/layer-3/nitrolite/pkg/app"
+	"github.com/layer-3/nitrolite/pkg/core"
 	"github.com/layer-3/nitrolite/pkg/log"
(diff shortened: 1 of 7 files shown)


nksazonov (approved):

ihsraham (approved):

nksazonov:
@coderabbitai review

PR #667, opened 2026-04-09
M-H09: use channel signer for all channel state node sigs

philanton:
## Description

The app-session router constructs two different signing paths. The channel API is initialized with a channel signer, while the app-session API is initialized with the generic signer directly.

```go
channelV1Handler := channel_v1.NewHandler(useChannelV1StoreInTx, memoryStore, actionGateway, nodeChannelSigner, stateAdvancer, statePacker, nodeAddress, cfg.MinChallenge, runtimeMetrics, cfg.MaxSessionKeyIDs)
appSessionV1Handler := app_session_v1.NewHandler(useAppSessionV1StoreInTx, memoryStore, actionGateway, signer, stateAdvancer, statePacker, nodeAddress, runtimeMetrics, cfg.MaxParticipants, cfg.MaxSessionDataLen, cfg.MaxSessionKeyIDs, cfg.MaxRebalanceSignedUpdates)
```

When an app deposit is processed, the handler verifies that the submitted user channel state uses the channel signature format, then packs that channel state and signs `NodeSig` with the generic app-session signer before storing it. The same pattern is used when app withdrawals and app closes issue release channel states. The release helper packs a channel state, signs it with the generic signer, and stores it as the node signature for that channel state.
This is inconsistent with the channel approach. Channel signatures are not plain Ethereum signatures. The channel signer wraps the underlying signature with a leading signer-type byte.

```go
func (s *ChannelDefaultSigner) Sign(data []byte) (sign.Signature, error) {
    sig, err := s.Signer.Sign(data)
    if err != nil {
        return sign.Signature{}, err
    }

    return append([]byte{byte(ChannelSignerType_Default)}, sig...), nil
}
```

That prefix is required because both the off-chain verifier and the ChannelHub contract treat the first byte of the signature as the validator/signer identifier and verify only the remaining bytes. Because the app-session flows write raw signatures into `core.State.NodeSig`, the first byte of the signature body is misinterpreted as the signer type. The stored value is therefore not a valid channel signature, and it is persisted as if it were.
As a result, app deposits, withdrawals, and closes can store channel states with invalid `NodeSig` values. Later checkpoint, challenge, or settlement flows will reject those states, making the latest app-produced state unusable.

## Summary by CodeRabbit

* **Tests**
  * Updated test fixtures across multiple test suites to use channel-specific mock signer implementation.

* **Refactor**
  * Updated handler signer dependency type to channel-specific implementation.
  * Adjusted RPC router configuration to align signer routing with handler requirements.

### DIFF
--- a/clearnode/api/app_session_v1/create_app_session_test.go
+++ b/clearnode/api/app_session_v1/create_app_session_test.go
@@ -22,7 +22,7 @@ func TestCreateAppSession_Success(t *testing.T) {
 		return fn(mockStore)
 	}
 
-	mockSigner := NewMockSigner()
+	mockSigner := NewMockChannelSigner()
 	mockAssetStore := new(MockAssetStore)
 	mockStatePacker := new(MockStatePacker)
 
@@ -126,7 +126,7 @@ func TestCreateAppSession_QuorumWithMultipleSignatures(t *testing.T) {
 		return fn(mockStore)
 	}
 
-	mockSigner := NewMockSigner()
+	mockSigner := NewMockChannelSigner()
 	mockAssetStore := new(MockAssetStore)
 	mockStatePacker := new(MockStatePacker)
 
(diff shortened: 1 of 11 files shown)


nksazonov (approved):
LGTM
Still, is it possible to create a (non-integration) test that will verify that a signed state produced by an app session operation is valid to be submitted on-chain?

PR #668, opened 2026-04-09
H-L07: fix(contracts/ChannelHub): restrict createChannel to non-existing channels

nksazonov:
## Description

The `createChannel(..)` function unconditionally emits `ChannelCreated(..)` event at line 491 even when operating on existing channels. The function allows `DEPOSIT`, `WITHDRAW`, and `OPERATE` intents on channels with status `OPERATING`, `DISPUTED`, or `MIGRATING_IN` (not just `VOID`), as confirmed in the intent-specific validation functions. When called on an existing channel to deposit additional funds or perform operations, it still emits `ChannelCreated(channelId, def.user, def.node, def, initState)`, misleading off-chain indexers and UIs. The channel definition is only stored if `status == VOID` (line 1064-1066), confirming that subsequent calls are updates, not creations. This behavior is inconsistent with the protocol's established pattern seen in `EscrowDepositEngine.sol:165` and `EscrowWithdrawalEngine.sol:161`, which explicitly check `require(ctx.status == EscrowStatus.VOID, EscrowAlreadyExists())` to prevent operations on existing escrows.

## Impact

Off-chain indexers listening for `ChannelCreated` events will incorrectly count the same channel multiple times, leading to inflated channel creation metrics. UIs that display channels based on `ChannelCreated` events may show duplicate channel listings, confusing users. Analytics dashboards and monitoring systems will report incorrect data about channel creation activity. While on-chain state remains correct (the definition is only stored once when `status == VOID`), the misleading events create data integrity issues for off-chain systems.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced channel creation validation with early status verification to prevent invalid state transitions.

* **Tests**
  * Expanded test coverage for channel creation scenarios, including success cases for multiple operation intents and revert conditions for invalid channel states.

### DIFF
--- a/contracts/src/ChannelHub.sol
+++ b/contracts/src/ChannelHub.sol
@@ -491,6 +491,8 @@ contract ChannelHub is ReentrancyGuard {
         }
 
         bytes32 channelId = Utils.getChannelId(def, VERSION);
+        require(_channels[channelId].status == ChannelStatus.VOID, IncorrectChannelStatus());
+
         address user = def.user;
 
         _requireValidDefinition(def);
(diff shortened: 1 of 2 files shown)


nksazonov:
@coderabbitai review

philanton (approved):

ihsraham (approved):

PR #674, opened 2026-04-10
M-H11: reject issuance of receiver state during escrow ops

philanton:
## Description

When a sender submits a `TransferSend`, the node automatically enters the receiver-side path and calls `issueTransferReceiverState`. That function creates and stores a new `TransferReceive` state for the receiver’s channel, even if the receiver’s latest signed state is already in an escrow phase such as `MutualLock` or `EscrowLock`.  As a result, an inbound transfer still advances the receiver’s channel state while the receiver is in the middle of a suspended escrow flow.
That breaks the normal completion path when the receiver later tries to complete the escrow flow.

## Release Notes

* **Bug Fixes**
  * Added validation to reject release or transfer operations when the receiver has an active escrow lock, with appropriate error messaging.

* **Tests**
  * Added test cases to validate error handling for escrow lock scenarios in both app session and channel operations.

### DIFF
--- a/clearnode/api/app_session_v1/handler.go
+++ b/clearnode/api/app_session_v1/handler.go
@@ -147,18 +147,11 @@ func (h *Handler) issueReleaseReceiverState(ctx context.Context, tx Store, recei
 	}
 
 	// TODO: move to DB query
-	shouldSign := true
-	if lastSignedState != nil {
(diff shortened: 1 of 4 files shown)


nksazonov:
@coderabbitai review

nksazonov (approved):

ihsraham (approved):

PR #676, opened 2026-04-10
YNU-798: monitor node liquidity on-chain and off-chain

philanton:
## Summary by CodeRabbit

* **New Features**
  * Added on-chain enforced balance tracking across the system and metric reporting for node and user balances.

* **API Updates**
  * User balance responses now include an additional `enforced` field (on-chain enforced balance).
  * SDKs updated to surface the new `enforced` value in returned balance entries.

* **Observability**
  * New metrics expose node and aggregated user balance totals, underfunded amounts, and releasable balances.


clearnode/store/memory/memory_store.go
```diff
@@ -131,7 +142,7 @@ func NewMemoryStoreV1(assetsConfig AssetsConfig, blockchainsConfig map[uint64]Bl
 		assets:               assets,
 		channelSigValidators: channelSigValidators,
 		supportedAssets:      supportedAssets,
-		tokenDecimals:        tokenDecimals,
```
nksazonov, line 134: `tokenDecimals` are still used in `GetTokenDecimals(...)`, however are dropped from `MemoryStoreV1` initialization. Please, bring them back.

pkg/blockchain/evm/locking_reactor.go
```diff
@@ -73,44 +93,46 @@ func (r *LockingContractReactor) HandleEvent(ctx context.Context, l types.Log) e
...
+		case lockingContractAbi.Events["UnlockInitiated"].ID:
+			err = r.handleUnlockInitiated(ctx, store, l)
+		case lockingContractAbi.Events["Withdrawn"].ID:
+			err = r.handleWithdrawn(ctx, store, l)
+		default:
+			logger.Warn("unknown event: " + eventID.Hex())
+		}
+		if r.onEventProcessed != nil {
+			r.onEventProcessed(r.blockchainID, err == nil)
+		}
```
nksazonov: Here the `onEventProcessed` callback fires inside the useStoreInTx closure (before commit). However, in `ChannelHubReactor.HandleEvent`, it fires outside:
```go
  err := r.useStoreInTx(...)
  if r.onEventProcessed != nil {
      r.onEventProcessed(r.blockchainID, err == nil)  // fires after commit
  }
```

The locking reactor can report `success=true` while the transaction is still uncommitted (or may roll back on `StoreContractEvent` failure). This is inconsistent with CHR and could skew metrics

clearnode/store/database/lifespan_metric.go
```diff
@@ -416,6 +214,101 @@ func (s *DBStore) GetLifetimeMetricLastTimestamp(name string) (time.Time, error)
...
+		FROM user_balances
+		GROUP BY home_blockchain_id, asset
+	`).Scan(&results).Error
+	if err != nil {
+		return nil, fmt.Errorf("failed to get off-chain liquidity: %w", err)
+	}
+
+	return results, nil
+}
+
```
nksazonov, line 316: Users who never had a channel state saved will have `home_blockchain_id = 0` (the column default). These get aggregated into a phantom "blockchain 0" bucket in the Prometheus metrics. I think we should filter with `WHERE home_blockchain_id > 0`.

clearnode/store/database/lifespan_metric.go
```diff
@@ -416,6 +214,101 @@ func (s *DBStore) GetLifetimeMetricLastTimestamp(name string) (time.Time, error)
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read node balance metrics: %w", err)
+	}
+
+	return results, nil
+}
+
+// UserBalanceSummary holds off-chain liquidity metrics for a given blockchain and asset.
+type UserBalanceSummary struct {
+	BlockchainID uint64          `gorm:"column:home_blockchain_id"`
```
nksazonov: There are inconsistent types in this file. `NodeBalance.BlockchainID` is string, however `UserBalanceSummary.BlockchainID` is `uint64`.
`main.go` uses `strconv.FormatUint` to format the `uint64` for `SetNodeBalance`, but already has a string from `GetNodeBalance`. The `NodeBalance` struct stores it as string because it comes from a JSON-keyed lifespan metric. I think we should either document why this inconsistency exists or unify by parsing to `uint64` in `GetNodeBalance`.

clearnode/store/database/db_store.go
```diff
@@ -49,12 +49,39 @@ func (s *DBStore) GetUserBalances(wallet string) ([]core.BalanceEntry, error) {
...
+			JOIN channel_states s ON s.home_channel_id = c.channel_id AND s.version = c.state_version
+			WHERE c.user_wallet = user_balances.user_wallet
+			  AND c.asset = user_balances.asset
+			  AND c.type = 1
+			  AND c.status <= 1
+			  AND c.state_version > 0
+			LIMIT 1
+		), 0)
+		WHERE user_wallet = ? AND asset = ?
+	`, wallet, asset).Error
```
nksazonov, line 83: Here the the subquery uses `LIMIT 1` without `ORDER BY`.                                                                                                                               

Indeed, the protocol enforces at most one open home channel per user/asset, so in practice this always matches a single row. However, the result is technically non-deterministic if that invariant is ever violated.                                                

The migration query for the same data already uses `DISTINCT ON ... ORDER BY state_version DESC, channel_id DESC` as a defensive measure. 

Could you either:                                                                                                                         
  - Add a comment explaining why ORDER BY is intentionally omitted (invariant is guaranteed upstream),
or
  - Add ORDER BY c.state_version DESC, c.channel_id DESC as a defence-in-depth practice, mirroring the migration

nksazonov (changes requested):
Please, fix the dropped `tokenDecimals` initialization bug. Other comments are warnings, but would be good to have them addressed.

clearnode/store/memory/memory_store.go
philanton: Fixed in [e3e1361](https://github.com/layer-3/nitrolite/pull/676/e3e13618059f35ff6ea17f1f06844ce5339ac74c).

pkg/blockchain/evm/locking_reactor.go
philanton: Good catch. Moved the callback outside the `useStoreInTx` closure so it fires after commit, matching `ChannelHubReactor.HandleEvent` — [e3e1361](https://github.com/layer-3/nitrolite/pull/676/e3e13618059f35ff6ea17f1f06844ce5339ac74c).

clearnode/store/database/lifespan_metric.go
philanton: Kept the `home_blockchain_id = 0` rows intentionally — surfacing funds not yet attributed to any home chain is useful for monitoring. Documented the behaviour in a godoc on `GetUserBalanceSummary` in [e3e1361](https://github.com/layer-3/nitrolite/pull/676/e3e13618059f35ff6ea17f1f06844ce5339ac74c).

clearnode/store/database/lifespan_metric.go
philanton: Aligned `UserBalanceSummary.BlockchainID` to `string` to match `NodeBalance` and the Prometheus exporter label type (so `main.go` no longer needs `strconv.FormatUint`) — [e3e1361](https://github.com/layer-3/nitrolite/pull/676/e3e13618059f35ff6ea17f1f06844ce5339ac74c).

clearnode/store/database/db_store.go
philanton: Added `ORDER BY c.updated_at DESC` as defence-in-depth and a comment noting the protocol-level invariant — [e3e1361](https://github.com/layer-3/nitrolite/pull/676/e3e13618059f35ff6ea17f1f06844ce5339ac74c).

nksazonov (approved):

PR #686, opened 2026-04-15
fix(clearnode): resolve a set of audit findings

philanton:
## Summary by CodeRabbit

* **Bug Fixes**
  * Wallet comparison is now case-insensitive to prevent false "same wallet" rejections.
  * Requests containing duplicate participant+asset allocations are rejected with clear error messaging.

* **New Features**
  * Validation of asset decimal precision for allocation amounts to enforce allowed decimals.

* **Tests**
  * Added tests covering duplicate-allocation rejection, deposit precision validation, rebalance duplicate handling, and same-wallet case-insensitive rejection.

### DIFF
--- a/clearnode/api/app_session_v1/submit_app_state_test.go
+++ b/clearnode/api/app_session_v1/submit_app_state_test.go
@@ -1720,3 +1720,109 @@ func TestSubmitAppState_AppRegistryDisabled(t *testing.T) {
 	mockStore.AssertNotCalled(t, "GetApp", mock.Anything)
 	mockStore.AssertExpectations(t)
 }
+
+func TestSubmitAppState_OperateIntent_DuplicateAllocation_Rejected(t *testing.T) {
+	// Setup
+	mockStore := new(MockStore)
+	storeTxProvider := func(fn StoreTxHandler) error {
+		return fn(mockStore)
+	}
+
+	mockSigner := NewMockChannelSigner()
+	mockAssetStore := new(MockAssetStore)
+	mockStatePacker := new(MockStatePacker)
+
+	handler := NewHandler(
+		storeTxProvider,
+		mockAssetStore,
+		&MockActionGateway{},
+		mockSigner,
+		core.NewStateAdvancerV1(mockAssetStore),
+		mockStatePacker,
+		"0xNode",
+		true,
+		metrics.NewNoopRuntimeMetricExporter(),
+		32, 1024, 256, 16,
+	)
+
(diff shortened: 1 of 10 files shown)


clearnode/api/app_session_v1/submit_app_state_test.go
```diff
@@ -1720,3 +1720,109 @@ func TestSubmitAppState_AppRegistryDisabled(t *testing.T) {
 	mockStore.AssertNotCalled(t, "GetApp", mock.Anything)
 	mockStore.AssertExpectations(t)
 }
+
+func TestSubmitAppState_OperateIntent_DuplicateAllocation_Rejected(t *testing.T) {
```
nksazonov, line 1907: Please, add tests for rejecting duplicate allocations for all other app session endpoints; and for other issues fixed in this PR.

nksazonov (commented):
LGTM, more tests needed

clearnode/api/channel_v1/submit_state_test.go
```diff
@@ -344,6 +344,112 @@ func TestSubmitState_TransferSend_ReceiverWithEscrowLock_Rejected(t *testing.T)
...
+	}
+
+	// Derive senderWallet from a real key — Address().String() returns checksummed (mixed case)
+	userSigner := NewMockSigner()
+	userWalletSigner, _ := core.NewChannelDefaultSigner(userSigner)
+	senderWallet := userSigner.PublicKey().Address().String() // e.g. "0xAbCdEf..."
+
+	// Use the SAME address as receiver — the transition AccountID stores it as-is,
+	// but issueTransferReceiverState calls NormalizeHexAddress which lowercases it.
+	// The fix uses strings.EqualFold to catch this.
```
nksazonov: Wait. Does it mean that `issueTransferReceieverState` function itself introduces the bug of a lower-cased address? Why would it do that? Is it possible to remove such logic from it?

Also, if `issueTransferReceieverState` does lower-case the address, then if used directly, does it mean that the `strings.EqualFold` will always fail unless the address is checksummed-back after `issueTransferReceieverState`?

philanton: Inside our system, we use only lower-cased addresses right now (for storing and logging, so the data we browse has the same format everywhere). When the time comes, it would be good to move to just bytes. Comparison always uses `strings.EqualFold`, which is the alternative to converting to bytes and comparing them.

nksazonov (approved):

PR #699, opened 2026-04-21
YNU-869: Refine ts-compat runtime migration surface

ihsraham:
This is part of the ongoing refinements to the TS SDKs, focused on the compat TypeScript SDK. It is now rebased on current `main` after #701.

It makes the compat runtime surface more honest and more useful during migration:
- direct-mapping legacy RPC helpers emit real v1-compatible payloads
- workflow-only legacy helpers fail fast with clear migration guidance instead of returning fake wire payloads
- selected legacy client conveniences are restored on top of the current SDK behavior
- compat amount handling keeps the domains explicit: `transfer()` and `getBalances()` use canonical asset raw-unit strings, on-chain token helpers use raw token-unit `bigint`, and app-session allocations stay human-decimal strings
- the rebase preserves #701 app-session `appDefinition` mapping and current `nitronode` naming
- runtime-focused compat tests cover wire shape, client mappings, and amount semantics

Validation run locally:
- `npm --prefix sdk/ts-compat run typecheck`
- `npm --prefix sdk/ts-compat test`
- `npm --prefix sdk/ts-compat run lint`

## Summary by CodeRabbit

* **New Features**
  * Added on-chain helpers (approve, allowance, token balance) and an open-channels query; RPC helpers now emit real v1-compatible requests and normalize envelope responses.

* **Behavior Changes**
  * Several legacy helper flows now fail fast as migration shims instead of returning placeholders.
  * App-session submit/close helpers now require an explicit version and preserve human-readable allocation amounts.

* **Documentation**
  * Clarified amount conventions and updated migration guides.

* **Tests**
  * Added unit tests for amount semantics and RPC wire/request-response shapes.


ihsraham:
pushed a small follow up here. gettokenbalance now rounds down to token precision before parseunits, and the compat docs now call out more clearly that createchannel and the workflow style helper shims throw with migration guidance instead of warning or returning fake wire payloads.

ihsraham (commented):
@coderabbitai review

nksazonov:
@coderabbitai review

sdk/ts-compat/src/client.ts
```diff
@@ -298,11 +379,44 @@ export class NitroliteClient {
...
+        if (cached) {
+            return cached;
+        }
+
+        const client = createPublicClient({
+            transport: http(this.getRPCUrl(key)),
+        });
+        this._publicClients.set(key, client);
+        return client;
+    }
```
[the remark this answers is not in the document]
ihsraham: good catch. i tightened the fallback so we only reuse the wallet chain rpc urls when the requested chain id actually matches, and added a regression test for the wrong-chain case.

sdk/ts-compat/src/client.ts
```diff
@@ -334,9 +457,9 @@ export class NitroliteClient {
...
 
     async deposit(tokenAddress: Address, amount: bigint): Promise<any> {
-        const { symbol, chainId, decimals, tokenAddress: resolvedAddr } = await this.resolveToken(tokenAddress);
+        const { symbol, chainId, tokenDecimals, tokenAddress: resolvedAddr } = await this.resolveToken(tokenAddress);
         await this.innerClient.setHomeBlockchain(symbol, chainId).catch(() => {});
-        const humanAmount = this.toHumanAmount(amount, decimals);
+        const humanAmount = this.toHumanAmount(amount, tokenDecimals);
         await this.innerClient.deposit(chainId, symbol, humanAmount);
         return await this.checkpointWithApproval(symbol, chainId, resolvedAddr);
     }
```
nksazonov, line 467: `getBalances()` converts ledger balances using asset decimals (e.g. 6 for YUSD), while `deposit()` and `withdrawal()` convert using token decimals (e.g. 8 for YUSD on Base Sepolia). For a token where these differ, the raw amount returned by `getBalances()` is not in the same unit as the raw amount expected by `deposit()`. A user who tries `deposit(token, BigInt(balance.amount))` will pass a value off by          `10^(tokenDecimals - assetDecimals)`. The per-method docs in README cover the conventions, but a JSDoc `@remarks` on `getBalances()` noting that it uses asset-level decimals (not chain-specific token decimals) would prevent this mistake at the call site.

sdk/ts-compat/src/rpc.ts
```diff
@@ -54,77 +229,294 @@ export const NitroliteRPC = {
...
+                app_session_id: params.app_session_id,
+                intent: RPCAppStateIntent.Close,
+                version: requireSubmitVersion('createCloseAppSessionMessage', params.version),
+                allocations: params.allocations,
+                session_data: params.session_data,
+                quorum_sigs: params.quorum_sigs,
+            } as SubmitAppStateRequestParamsV04,
+            RPCAppStateIntent.Close,
+            params.quorum_sigs,
+        ),
```
nksazonov, line 387: `requireSubmitVersion('createCloseAppSessionMessage', params.version)` is called twice on the close path: once when building the inner object literal (line 378) and again inside `buildSubmitAppStateParams` (line 193). Both calls throw the same error with the same message. The outer call fires first, so the user experience is correct, but the double-validation is noise. Remove the inner call by passing the already-validated version value, or extract the guard to the outer scope and pass it in.

nksazonov (approved):
LGTM with some comments

philanton (approved):

sdk/ts-compat/src/client.ts
ihsraham: [d057a64](https://github.com/layer-3/nitrolite/commit/d057a64cf4862774ec1f84bf85e17b5daa9e1912)

sdk/ts-compat/src/rpc.ts
ihsraham: [d057a64](https://github.com/layer-3/nitrolite/commit/d057a64cf4862774ec1f84bf85e17b5daa9e1912)

philanton (approved):

PR #701, opened 2026-04-21
YNU-869: Guard TS SDK drift and CI validation

ihsraham:
this is part of the ts sdk refinement work.

it does three things:
- removes the unsupported `channels.v1.get_states` surface from shared rpc, ts raw rpc, and the protocol docs
- adds a static drift test in `sdk/ts` to keep the ts raw rpc surface aligned with `pkg/rpc` and the live clearnode router
- re-enables sdk validation on `main` prs and pushes for both `sdk/ts` and `sdk/ts-compat`, using the real package paths and the full `typecheck` / `test` / `lint` / `build` gate

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the channels.get_states RPC method and its client interfaces (API, SDK clients, and related docs/examples).

* **Tests**
  * Added automated test to ensure RPC method lists stay aligned across implementations.
  * Expanded CI validations to include typecheck, lint, and CI builds for SDKs.

* **Chores**
  * CI updated to run separate jobs for TS SDK and TS-compat variants; added CI-only build scripts.

* **SDK**
  * TS-compat client now prefers embedded appDefinition payloads when mapping app session data.

Additional compat fix discovered during stress Clearnode testing:

While wiring `cosign-demo` to `wss://clearnode-stress.yellow.org/v1/ws`, `getAppSessionsList()` exposed a compat drift: current v1 SDK sessions return participant/quorum/nonce data under `appDefinition`, but compat was still reading the old flat `participants` shape. This PR now maps `appDefinition.participants` first and keeps a flat-shape fallback for older callers.

### DIFF
--- a/.github/workflows/main-pr.yml
+++ b/.github/workflows/main-pr.yml
@@ -29,6 +29,7 @@ jobs:
     with:
       project-path: 'sdk/ts-compat'
       project-name: 'TS SDK Compat'
+      bootstrap-project-path: 'sdk/ts'
 
   build-and-publish-clearnode:
     name: Build and Publish (Clearnode)
--- a/.github/workflows/main-pr.yml
+++ b/.github/workflows/main-pr.yml
@@ -16,12 +16,19 @@ jobs:
     name: Test (Foundry)
     uses: ./.github/workflows/test-forge.yml
 
-  # test-sdk:
-  #   name: Test (SDK)
-  #   uses: ./.github/workflows/test-sdk.yml
-  #   with:
-  #     project-path: 'sdk'
-  #     project-name: 'SDK'
+  test-sdk-ts:
+    name: Test (TS SDK)
+    uses: ./.github/workflows/test-sdk.yml
+    with:
+      project-path: 'sdk/ts'
+      project-name: 'TS SDK'
+
+  test-sdk-compat:
+    name: Test (TS SDK Compat)
+    uses: ./.github/workflows/test-sdk.yml
+    with:
+      project-path: 'sdk/ts-compat'
+      project-name: 'TS SDK Compat'
 
   build-and-publish-clearnode:
     name: Build and Publish (Clearnode)
@@ -56,4 +63,3 @@ jobs:
           cache-from: type=gha
           build-args: |
             VERSION=${{ steps.sha.outputs.short_sha }}
-
--- a/.github/workflows/test-sdk.yml
+++ b/.github/workflows/test-sdk.yml
@@ -11,6 +11,11 @@ on:
         description: 'Human-readable name for the project'
         required: true
         type: string
+      bootstrap-project-path:
+        description: 'Optional SDK project that must be built before validating the target project'
+        required: false
+        type: string
+        default: ''
 
 jobs:
   test:
@@ -27,6 +32,16 @@ jobs:
           node-version-file: '${{ inputs.project-path }}/package.json'
           cache: 'npm'
           cache-dependency-path: '${{ inputs.project-path }}/package-lock.json'
+
+      - name: Install bootstrap dependencies
+        if: ${{ inputs.bootstrap-project-path != '' }}
+        run: npm ci
+        working-directory: ${{ inputs.bootstrap-project-path }}
+
+      - name: Build bootstrap project
+        if: ${{ inputs.bootstrap-project-path != '' }}
+        run: npm run build:ci
+        working-directory: ${{ inputs.bootstrap-project-path }}
 
       - name: Install dependencies
         run: npm ci
--- a/.github/workflows/test-sdk.yml
+++ b/.github/workflows/test-sdk.yml
@@ -31,7 +31,9 @@ jobs:
         with:
           node-version-file: '${{ inputs.project-path }}/package.json'
           cache: 'npm'
-          cache-dependency-path: '${{ inputs.project-path }}/package-lock.json'
+          cache-dependency-path: |
+            ${{ inputs.project-path }}/package-lock.json
+            ${{ inputs.bootstrap-project-path != '' && format('{0}/package-lock.json', inputs.bootstrap-project-path) || '' }}
 
       - name: Install bootstrap dependencies
         if: ${{ inputs.bootstrap-project-path != '' }}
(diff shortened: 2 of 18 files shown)


.github/workflows/main-pr.yml
```diff
@@ -16,12 +16,19 @@ jobs:
...
+    with:
+      project-path: 'sdk/ts'
+      project-name: 'TS SDK'
+
+  test-sdk-compat:
+    name: Test (TS SDK Compat)
+    uses: ./.github/workflows/test-sdk.yml
+    with:
+      project-path: 'sdk/ts-compat'
+      project-name: 'TS SDK Compat'
```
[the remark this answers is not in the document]
ihsraham: good catch on the underlying issue. i fixed it in 611a9cbc, but took a different route than adding `needs: test-sdk-ts`.

`test-sdk-compat` now bootstraps `sdk/ts` inside the same reusable workflow via `bootstrap-project-path: "sdk/ts"`, runs `npm ci` + `npm run build:ci` there first, and then validates `sdk/ts-compat`.

that keeps the ts sdk and compat jobs parallel while still making clean checkouts pass. i also reproduced the original failure in a clean temp checkout before pushing the fix.

sdk/ts/test/unit/rpc-drift.test.ts
```diff
@@ -0,0 +1,86 @@
...
+            fs.readFileSync(path.join(repoRoot, 'pkg/rpc/methods.go'), 'utf8')
+        );
+        const tsMethods = extractTsMethodLiterals(
+            fs.readFileSync(path.join(repoRoot, 'sdk/ts/src/rpc/methods.ts'), 'utf8')
+        );
+
+        const { missing, extra } = diff(tsMethods, goMethods);
+
+        expect({ missing, extra }).toEqual({ missing: [], extra: [] });
+    });
```
nksazonov, line 72: This drift guard reads `rpc_router.go` statically and always picks up `app_sessions.v1.rebalance_app_sessions`, but in `clearnode/api/rpc_router.go:119` that handler is conditionally registered:                                                                 

```go                                                                                                   
if cfg.MaxRebalanceSignedUpdates >= 2 {                                                                                                 
     appSessionV1Group.Handle(rpc.AppSessionsV1RebalanceAppSessionsMethod.String(), ...)
}                                                                                                                                       
```
                                                    
The regex `Handle\(rpc\.[A-Za-z0-9]+Method\.String\(\)`, matches the call regardless of the enclosing if, so the second test will pass even on a clearnode config where the method is not actually served. The guard gives a false sense of alignment for conditionally-registered methods. If future callers rely on this method being available, it may not be.

.github/workflows/test-sdk.yml
```diff
@@ -30,10 +33,32 @@ jobs:
           cache: 'npm'
```
nksazonov, line 33: The `setup-node` step uses cache-dependency-path: `${{ inputs.project-path }}/package-lock.json`
For the compat job, `project-path` is `sdk/ts-compat`, so `sdk/ts/package-lock.json` (the bootstrap) is not part of the cache key. If bootstrap dependencies change, the cached `node_modules` for that step will be stale and the old modules will be used until the cache expires. Correctness is preserved (`npm ci` always installs), but unnecessary cache misses or stale installs can occur. Consider:
```yml
 cache-dependency-path: |                                                                                                                
   ${{ inputs.project-path }}/package-lock.json                                                                                        
   ${{ inputs.bootstrap-project-path != '' && format('{0}/package-lock.json', inputs.bootstrap-project-path) || '' }}                    
```

.github/workflows/test-sdk.yml
```diff
@@ -30,10 +33,32 @@ jobs:
...
           cache-dependency-path: '${{ inputs.project-path }}/package-lock.json'
 
+      - name: Install bootstrap dependencies
+        if: ${{ inputs.bootstrap-project-path != '' }}
+        run: npm ci
+        working-directory: ${{ inputs.bootstrap-project-path }}
```
nksazonov, line 41: The CI runs `npm run typecheck (tsc --noEmit)` and then later `npm run build:ci (tsc)`. Both do a full type check. The `build:ci` step already validates types as part of compilation, so the separate typecheck step is redundant and doubles CI time. Consider removing typecheck from the workflow and relying solely on `build:ci` to catch type errors.

sdk/ts/test/unit/rpc-drift.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    it('keeps every TS raw RPC method registered by the live router', () => {
```
nksazonov: The test name reads as "TS ⊆ router". But `diff(tsMethods, routerMethods)` and `expect({ missing, extra }).toEqual({ missing: [], extra: [] })` enforces full bidirectional equality (TS == router). The name should reflect this: `'keeps TS raw RPC method surface aligned with the live router'`.

nksazonov (approved):
Generally good, with some comments

sdk/ts/test/unit/rpc-drift.test.ts
philanton, line 1: Cool initiative! Later, we could upgrade this one or/and add something similar for SDK Client (and anything else worth checking integrity)

philanton (approved):

sdk/ts/test/unit/rpc-drift.test.ts
philanton: This condition couldn't be checked at build time. I test wouldn't assume for this method - that'd be an error

philanton: P.S.: @ihsraham it's fine

sdk/ts/test/unit/rpc-drift.test.ts
ihsraham: [5b0327e](https://github.com/layer-3/nitrolite/commit/5b0327ed310b7188becd67548422e82091511f80)

.github/workflows/test-sdk.yml
ihsraham: [5b0327e](https://github.com/layer-3/nitrolite/commit/5b0327ed310b7188becd67548422e82091511f80)

.github/workflows/test-sdk.yml
ihsraham: [5b0327e](https://github.com/layer-3/nitrolite/commit/5b0327ed310b7188becd67548422e82091511f80)

nksazonov (approved):

PR #704, opened 2026-04-24
M2-H01: fix(contracts/ChannelHub): add finalize escrow home chain intent check

nksazonov:
## **Description**

`finalizeEscrowDeposit` and `finalizeEscrowWithdrawal` route home-chain calls before checking the candidate state intent. When escrow metadata is absent and the channel is home on the current chain, the functions call `_processHomeChainEscrowFinalize`, which validates signatures and forwards the candidate to `ChannelEngine.validateTransition`.

```
        if (_isEscrowDepositHomeChain(channelId, escrowId)) {
            // HOME CHAIN: Get user from channel definition
            ChannelMeta storage channelMeta = _channels[channelId];
            _processHomeChainEscrowFinalize(channelId, candidate, channelMeta.definition.user);
            emit EscrowDepositFinalizedOnHome(escrowId, channelId, candidate);
            return;
        }
```

The intent check is only reached on the non-home path. The home path therefore accepts any valid next signed channel transition supported by ChannelEngine. This breaks the intended escrow deposit flow. The user receives a co-signed `INITIATE_ESCROW_DEPOSIT` state so they can submit it on the non-home chain and lock their funds in escrow. The home-chain submission of that same state is intentionally restricted to the node.

```
        if (_isChannelHomeChain(channelId)) {
            require(msg.sender == NODE, IncorrectMsgSender());
            _processHomeChainEscrowInitiate(channelId, candidate);
            emit EscrowDepositInitiatedOnHome(escrowId, channelId, candidate);
        }
```

However, the user can submit that signed initiation state through `finalizeEscrowDeposit` on the home chain. Since home-chain escrow metadata is empty,  `_isEscrowDepositHomeChain` returns true and the candidate is processed as a generic channel transition. As a result, the home channel advances to the `INITIATE_ESCROW_DEPOSIT` state and the node’s matching liquidity is moved from the node vault into the channel’s locked funds, even if the non-home escrow was never created and the user never locked the corresponding funds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation of state intent across escrow initiation and finalization operations, ensuring correct operation types are supplied.
  * Strengthened signature validation for home-chain escrow finalizations using channel definition validators.

* **Tests**
  * Added comprehensive unit test coverage validating incorrect state intent rejection across all core ChannelHub operations.

### DIFF
--- a/contracts/src/ChannelHub.sol
+++ b/contracts/src/ChannelHub.sol
@@ -696,7 +696,7 @@ contract ChannelHub is ReentrancyGuard {
 
         if (_isChannelHomeChain(channelId)) {
             require(msg.sender == NODE, IncorrectMsgSender());
-            _processHomeChainEscrowInitiate(channelId, candidate);
+            _processHomeChainEscrow(channelId, candidate);
             emit EscrowDepositInitiatedOnHome(escrowId, channelId, candidate);
         } else {
             // NON-HOME CHAIN: Create escrow record - recover addresses from signatures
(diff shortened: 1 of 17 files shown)


nksazonov:
@coderabbitai review

philanton (approved):

ihsraham (approved):

PR #705, opened 2026-04-24
YNU-869: Add TS SDK protocol drift guards

ihsraham:
## Summary

Adds protocol, TS SDK, and compat SDK drift guards on top of current `main`.

#699 and #701 are now merged, so this branch has been rebased directly on `main` and preserves current behavior for:

- `nitronode` naming and paths
- #699 compat runtime semantics and migration surface
- #701 `appDefinition` mapping and SDK `applicationID` exports
- package versions at `1.2.1`

### Drift Guards

- Adds static guards for RPC method coverage, RPC DTO shape, public API snapshots, consumed ABI surface, app/session-key signing vectors, and response transforms.
- Adds compat drift coverage for v1 app-session mapping, legacy flat fallback behavior, empty-session behavior, asset decimal conversion, config pass-through, and session-key public APIs.
- Adds runtime smoke coverage that builds and starts an isolated local `nitronode`, then verifies live SDK and compat calls.

### SDK Validation

- Validates key-state `states` containers before mapping.
- Validates app-session payloads, allocations, app definitions, participants, and quorum at the wire boundary.
- Rejects arrays in session key-state record transforms.

### Nitronode Database

- Propagates SQLite migration errors from `connectToSqlite`.
- Returns Postgres migration errors instead of panicking.
- Keeps SQLite auto-migration aligned with current runtime tables, including state, transaction, app participants, and key-state tables.
- Uses a parameterized schema existence check and quoted schema identifiers for Postgres schema setup.

### Runtime Smoke And CI

- Builds `./nitronode` and uses `NITRONODE_RUNTIME_SMOKE_*` env names in new scripts/workflows/docs.
- Requires `NITRONODE_RUNTIME_SMOKE_PRIVATE_KEY` for external smoke mode.
- Keeps the hardcoded Anvil key only for isolated local smoke.
- Restricts runtime-smoke log directories to repo-local paths.
- Passes an allowlisted child-process env instead of forwarding all parent env.
- Captures compat logs for failure diagnostics.
- Builds Foundry artifacts before TS SDK/static drift ABI checks in CI.

## Review Notes

Addressed current actionable review feedback around SDK response validation, database migration error handling, runtime-smoke env/log safety, public API tautology tests, ABI drift-test stability, and stale Clearnode wording in PR-added files.

Deferred CodeRabbit docstring coverage and shared public API test utility extraction because they are broader cleanup items, not blockers for this PR.

## Verification

Local checks run from a fresh worktree:

- `npm --prefix sdk/ts run typecheck`
- `npm --prefix sdk/ts test`
- `npm --prefix sdk/ts run lint`
- `npm --prefix sdk/ts run build:ci`
- `npm --prefix sdk/ts-compat run typecheck`
- `npm --prefix sdk/ts-compat test`
- `npm --prefix sdk/ts-compat run lint`
- `npm --prefix sdk/ts-compat run build:ci`
- `forge build` for local ABI artifacts
- `./scripts/check-protocol-drift.sh --static`
- `./scripts/check-protocol-drift.sh --runtime`
- `go test ./nitronode/... ./pkg/rpc/...`
- `git diff --check`

GitHub checks have started on the rebased branch.


philanton (approved):

nitronode/store/database/database.go
```diff
@@ -214,7 +214,24 @@ func migratePostgres(cnf DatabaseConfig, embedMigrations embed.FS) error {
 }
 
 func migrateSqlite(db *gorm.DB) error {
```
nksazonov, line 218: **[CRITICAL]** `migrateSqlite` returns an `error` (line 234) but its call site in `connectToSqlite` drops it — line 123 calls `migrateSqlite(db)` with no error check, and line 125 prints "Successfully auto-migrated" regardless. If `AutoMigrate` fails the function returns `(db, nil)`, giving the caller a non-nil DB with a partial or missing schema. The first query against a missing table is where the failure surfaces.

Check the error and propagate it, consistent with every other error in `connectToSqlite`.

nitronode/store/database/database.go
```diff
@@ -214,7 +214,24 @@ func migratePostgres(cnf DatabaseConfig, embedMigrations embed.FS) error {
 }
```
nksazonov, line 216: **[CRITICAL]** `migratePostgres` (whose closing `}` is this line) has an `error` return type and its caller at line 59 checks that return — but line 209 `panic`s unconditionally when Goose fails, so the caller's error handling is never reached. A failed Postgres migration kills the entire process with a raw stack trace instead of going through the structured shutdown the caller would provide.

Replace `panic(err)` at line 209 with `return fmt.Errorf(...)` so the error propagates normally.

nitronode/store/database/database.go
```diff
@@ -214,7 +214,24 @@ func migratePostgres(cnf DatabaseConfig, embedMigrations embed.FS) error {
...
 
 func migrateSqlite(db *gorm.DB) error {
-	if err := db.AutoMigrate(&AppV1{}, &AppLedgerEntryV1{}, &Channel{}, &AppSessionV1{}, &ContractEvent{}, &BlockchainAction{}, &AppSessionKeyStateV1{}, &AppSessionKeyApplicationV1{}, &AppSessionKeyAppSessionIDV1{}, &UserBalance{}, &UserStakedV1{}, &ActionLogEntryV1{}, &LifespanMetric{}); err != nil {
+	if err := db.AutoMigrate(
```
nksazonov, line 219: **[INFORMATIONAL]** This `AutoMigrate` call does not include `State` or `Transaction`, but `testing.go:57` migrates both. `blockchain_worker.go:146` calls `GetStateByID`, which queries `channel_states`. A developer who configures SQLite with even one blockchain entry will hit `no such table: channel_states` the first time the blockchain worker runs. The runtime smoke avoids this by setting `blockchains: []`, but the discrepancy is a silent trap outside the smoke context.

Either add `State` and `Transaction` to match the test helper, or add a startup guard that rejects a SQLite + configured-blockchains combination and documents the constraint.

_Also: `ensurePostgresqlSchema` at lines 168 and 178 interpolates `cnf.Schema` unquoted into raw SQL. Line 178's `CREATE SCHEMA IF NOT EXISTS %s` has no quoting — use `pq.QuoteIdentifier` for the DDL and a parameterised query for the `information_schema` check._

scripts/drift/runtime-smoke.mjs
```diff
@@ -0,0 +1,354 @@
...
+
+const scriptDir = path.dirname(fileURLToPath(import.meta.url));
+const repoRoot = path.resolve(scriptDir, '../..');
+const wsURL = process.env.CLEARNODE_RUNTIME_SMOKE_WS_URL ?? 'ws://127.0.0.1:7824/ws';
+const readyTimeoutMs = Number(process.env.CLEARNODE_RUNTIME_SMOKE_READY_TIMEOUT_MS ?? 15000);
+const adversarialMode = process.env.CLEARNODE_RUNTIME_SMOKE_ADVERSARIAL ?? '';
+const externalLogDirInput = process.env.CLEARNODE_RUNTIME_SMOKE_LOG_DIR ?? '';
+const externalLogDir = externalLogDirInput ? path.resolve(repoRoot, externalLogDirInput) : '';
+const useExternalNode = process.env.CLEARNODE_RUNTIME_SMOKE_EXTERNAL === '1';
+const privateKey =
```
nksazonov: **[INFORMATIONAL]** This key is used in two roles: as the local Clearnode's signer identity (line 165) and as the SDK client wallet for all RPC calls (line 239). For the isolated local smoke that's fine — it's Hardhat account #1, a publicly documented zero-value test address. For external smoke mode (`CLEARNODE_RUNTIME_SMOKE_EXTERNAL=1`) though, this same key signs state transitions against an operator-supplied external Clearnode. If that node has state tied to this address from a previous run, the smoke operates on it with a key that appears verbatim in the Hardhat docs. Secret-scanning tools will also flag this on every CI run.

Read the key from an environment variable and only fall back to the hardcoded value when running against the isolated local node; require the env var and fail early if it's missing in external mode.

scripts/drift/runtime-smoke.mjs
```diff
@@ -0,0 +1,354 @@
...
+import { Client, createSigners, withErrorHandler } from '../../sdk/ts/dist/index.js';
+import { NitroliteClient } from '../../sdk/ts-compat/dist/index.js';
+
+const scriptDir = path.dirname(fileURLToPath(import.meta.url));
+const repoRoot = path.resolve(scriptDir, '../..');
+const wsURL = process.env.CLEARNODE_RUNTIME_SMOKE_WS_URL ?? 'ws://127.0.0.1:7824/ws';
+const readyTimeoutMs = Number(process.env.CLEARNODE_RUNTIME_SMOKE_READY_TIMEOUT_MS ?? 15000);
+const adversarialMode = process.env.CLEARNODE_RUNTIME_SMOKE_ADVERSARIAL ?? '';
+const externalLogDirInput = process.env.CLEARNODE_RUNTIME_SMOKE_LOG_DIR ?? '';
+const externalLogDir = externalLogDirInput ? path.resolve(repoRoot, externalLogDirInput) : '';
```
nksazonov: **[INFORMATIONAL]** `path.resolve(repoRoot, externalLogDirInput)` honours `..` sequences and absolute paths. A value like `../../home/runner/.ssh` escapes the repo tree, and lines 187-190 will then `mkdir` and `writeFile` at that resolved path. In CI the value is always `runtime-smoke-logs` (safe), but nothing prevents an out-of-tree value.

After resolving, check that the result starts with `repoRoot` and throw if it doesn't.

scripts/drift/runtime-smoke.mjs
```diff
@@ -0,0 +1,354 @@
...
+      logStep(`writing isolated config in ${configDir}`);
+      await writeConfig(configDir);
+      logStep('building temporary Clearnode binary');
+      await runCommand('go', ['build', '-o', binaryPath, './clearnode'], { cwd: repoRoot }, 'setup');
+
+      logStep(`starting Clearnode and waiting for ${wsURL}`);
+      child = spawn(binaryPath, {
+        cwd: repoRoot,
+        env: {
+          ...process.env,
```
nksazonov: **[INFORMATIONAL]** Spreading `...process.env` passes the entire parent environment — including `GITHUB_TOKEN`, repository secrets, and cloud credentials present in the runner — into the clearnode subprocess. The binary is trusted code, so the current risk is low, but any future log path in clearnode that prints env vars would expose runner secrets in the CI artifact.

Pass only the variables the clearnode process actually needs (`PATH`, `HOME`, `TMPDIR`, and the `CLEARNODE_*` keys) rather than forwarding the entire environment.

sdk/ts/package.json
```diff
@@ -21,6 +22,7 @@
         "lint": "eslint src test --ext .ts",
```
nksazonov, line 22: **[INFORMATIONAL]** Both `sdk/ts` and `sdk/ts-compat` were at `1.2.1` on `main`; this PR reverts them to `1.2.0` (line 3 in each `package.json`). `npm view @yellow-org/sdk version` returns `1.2.0` — that version is already on the registry, so publishing will fail with `E403`. Additionally, this PR removes three previously-exported symbols (`withApplicationID`, `APPLICATION_ID_QUERY_PARAM`, `appendApplicationIDQueryParam`), which is a breaking change. The version revert looks like a rebase artefact from being stacked on `#701`.

Bump both packages to at least `1.2.1` before marking ready; if the export removals are intentional and permanent, consider a minor or major bump instead.

sdk/ts/test/unit/public-api-drift.test.ts
```diff
@@ -0,0 +1,250 @@
...
+describe('SDK public runtime API drift guard', () => {
+    it('keeps root runtime exports intentional', () => {
+        expect(Object.keys(publicApi).sort()).toMatchSnapshot();
+    });
+
+    it('keeps root TypeScript public API signatures intentional', () => {
+        expect(serializePublicApi()).toMatchSnapshot();
+    });
+
+    it('proves adversarial public export removal is observable', () => {
```
nksazonov: **[INFORMATIONAL]** This test creates a local `Set` copy of `Object.keys(publicApi)`, calls `.delete('Client')` on that copy, then asserts the copy no longer contains `'Client'`. `Set.delete` followed by `Set.has` is always `false` — the test passes identically whether or not `Client` is actually exported. The real guard is already provided by `toMatchSnapshot()` two tests above.

Replace the second assertion with a direct `toContain('Client')` check on the actual `publicApi`, or drop the test entirely.

sdk/ts-compat/test/unit/public-api-drift.test.ts
```diff
@@ -0,0 +1,272 @@
...
+                expect.stringContaining('submitSessionKeyState:'),
+                expect.stringContaining('getLastKeyStates:'),
+            ])
+        );
+
+        const helper = api.find((member) => member.name === 'toSessionKeyQuorumSignature');
+        expect(helper?.signatures?.[0]).toContain('signature: Hex | string');
+    });
+
+    it('proves adversarial public export removal is observable', () => {
```
nksazonov: **[INFORMATIONAL]** Same tautological pattern: deletes `'NitroliteClient'` from a local copy of `Object.keys(publicApi)`, then asserts the local copy no longer has it. Always passes regardless of the real export surface.

Assert `Object.keys(publicApi)` contains `'NitroliteClient'` directly, and drop the local-copy assertion.

sdk/ts/test/unit/rpc-drift.test.ts
```diff
@@ -0,0 +1,171 @@
...
+            `rpc_router.go references unresolved rpc method constants: ${unresolvedMethodNames.join(
+                ', '
+            )}`
+        );
+    }
+
+    return new Set(methodNames.map((name) => namedLiterals.get(name) as string));
+}
+
+function extractTsClientMethods(source: string): Set<string> {
```
nksazonov, line 49: **[INFORMATIONAL]** The regex matches method names by looking for exactly two leading spaces. If `client.ts` is ever reformatted or the class extracted to a separate file, the regex returns an empty set and `missingClientMethods` becomes every entry in `publicClientMethodsByRPCMethod` — the test fails loudly but with a confusing "all methods missing" message rather than "regex broke".

Add a `expect(clientMethods.size).toBeGreaterThan(N)` guard directly before the main assertion so a broken regex produces an immediately actionable failure, and add a comment noting the indentation dependency.

sdk/ts/test/unit/abi-drift.test.ts
```diff
@@ -0,0 +1,269 @@
...
+        signaturesByName.set(entry.name, signatures);
+    }
+
+    return new Map(
+        [...signaturesByName].map(([name, signatures]) => [name, signatures.sort().join('\n')])
+    );
+}
+
+function loadArtifact(relativePath: string): readonly AbiEntry[] {
+    const artifact = JSON.parse(fs.readFileSync(path.join(repoRoot, relativePath), 'utf8'));
```
nksazonov: **[INFORMATIONAL]** `readFileSync` is called without checking whether the file exists. On a fresh clone or any CI run without a prior `forge build`, the test throws a raw `ENOENT` path error that gives no hint about the fix.

Check for the file first and throw a message like `"ABI artifact not found — run: cd contracts && forge build"` if it's missing.

sdk/ts/src/utils.ts
```diff
@@ -374,7 +394,7 @@ export function transformAppSessionInfo(raw: any): AppSessionInfoV1 {
  * The server returns snake_case JSON that needs conversion to SDK types.
  */
 export function transformAppDefinitionFromRPC(raw: any): AppDefinitionV1 {
-  if (!raw.application_id || raw.nonce === undefined || raw.nonce === null) {
+  if (!raw || !raw.application_id || raw.nonce === undefined || raw.nonce === null) {
```
nksazonov: **[INFORMATIONAL]** The guard added here checks `application_id` and `nonce`, but `quorum` (line 406) is taken directly from the raw object — `undefined` if the server omits it, silently producing `NaN` in any downstream numeric operation. Individual participant fields at lines 402-405 are also cast without checks. The sibling `transformAppAllocationFromRPC` added in this same PR validates every field explicitly; the inconsistency means allocation shape errors are caught at the wire boundary while definition shape errors are not.

Add explicit presence checks for `quorum` and `participants` to match the validation style of `transformAppAllocationFromRPC`.

sdk/ts-compat/src/client.ts
```diff
@@ -551,29 +551,38 @@ export class NitroliteClient {
     }
```
nksazonov, line 746: **[INFORMATIONAL]** Separately from this hunk: the `@deprecated` JSDoc tags were removed from `getAssetsList()` (line 641 of this file) and `ClearNodeAsset` (`src/types.ts:339`) without explanation. Without them, IDEs stop showing strikethrough on call sites and no longer suggest the migration path, which may lead developers to treat these as promoted first-class APIs rather than migration holdovers.

If the intent is to stabilise these as supported compat APIs, restore the tags with a note reflecting that decision; if the removal was accidental, restore the original tags as-is.

nitronode/store/database/database.go
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). sqlite migration errors now return from `connectToSqlite` instead of being dropped.

nitronode/store/database/database.go
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). goose migration errors now return normally instead of panicking.

nitronode/store/database/database.go
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). sqlite migration now includes the missing runtime tables, and postgres schema setup uses a parameterized check plus quoted names.

scripts/drift/runtime-smoke.mjs
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). external smoke now requires `NITRONODE_RUNTIME_SMOKE_PRIVATE_KEY`; the anvil key is local-only.

scripts/drift/runtime-smoke.mjs
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). smoke log paths now have to resolve inside the repo.

scripts/drift/runtime-smoke.mjs
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). the child process now gets only the env values it needs.

sdk/ts/package.json
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). both packages stay at `1.2.1`, and the application id exports are preserved.

sdk/ts/test/unit/public-api-drift.test.ts
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). the test now checks the actual sdk export list.

sdk/ts-compat/test/unit/public-api-drift.test.ts
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). the compat test now checks the actual export list.

sdk/ts/test/unit/rpc-drift.test.ts
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). added a direct guard so a broken method extractor fails clearly.

sdk/ts/test/unit/abi-drift.test.ts
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). missing abi artifacts now get a clear message, and ci builds foundry artifacts before the sdk drift checks.

sdk/ts/src/utils.ts
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). app definition validation now checks participants, quorum, and participant fields.

sdk/ts-compat/src/client.ts
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). the legacy compat deprecation notes are preserved.

scripts/drift/runtime-smoke.mjs
```diff
@@ -0,0 +1,399 @@
...
+      if (err) reject(err);
+      else resolve();
+    };
+
+    const timer = setTimeout(() => finish(new Error('WebSocket connect timeout')), timeoutMs);
+    ws.onopen = () => finish();
+    ws.onerror = () => finish(new Error('WebSocket connection error'));
+    ws.onclose = () => finish(new Error('WebSocket closed before open'));
+  });
+}
```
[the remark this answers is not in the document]
ihsraham: fixed in [32ad8fc](https://github.com/layer-3/nitrolite/commit/32ad8fc9a297c30b7d64de2dde8c03e84429145a). runtime smoke now falls back to `ws` when global `WebSocket` is missing on node 20.

nksazonov (approved):
Good job!

PR #716, opened 2026-04-30
YNU-912: Package SDK MCP for npm publishing

ihsraham:
## Summary

Packages `sdk/mcp` for public npm distribution as `@yellow-org/sdk-mcp`, so Claude, Codex, Cursor, VS Code, and other stdio MCP clients can run the Yellow SDK / Nitrolite protocol context server without cloning the Nitrolite monorepo.

## What changed

- Adds npm publish metadata for `sdk/mcp`: compiled `dist` entrypoint, executable `yellow-sdk-mcp` bin, public publish config, minimal package `files`, and `mcpName`.
- Adds a release-time `content/` snapshot builder so published installs can serve SDK/protocol docs from packaged files while local monorepo development still reads live source first.
- Adds `server_info` for package, SDK, compat, Go module, protocol, transport, and content-mode debugging.
- Adds MCP Registry metadata in `server.json` and a tag-triggered `mcp-v*` workflow that builds, packs, validates, smoke-tests, publishes to npm, then publishes registry metadata.
- Updates the README with install snippets for Claude Code, Claude Desktop, Codex, Cursor, VS Code, and local development.

## Release safety

The workflow now publishes the exact tarball it verifies. It asserts critical package paths, blocks source/test file leakage, installs the packed tarball into a temp app, and exercises packaged-snapshot mode through `server_info`, `lookup_method`, and `lookup_rpc_method` before publish.

`mcp-publisher` is pinned to `v1.7.6` with SHA256 verification instead of using a floating `latest` download.

## Validation run locally

- `npm run typecheck` from `sdk/mcp`
- `npm run build` from `sdk/mcp`
- package contents assertion against `npm pack --json`
- temp tarball install + MCP client smoke test for `server_info`, `lookup_method`, and `lookup_rpc_method`
- MCP Registry metadata validation
- `git diff --cached --check`

## Notes for reviewers

- Package version intentionally mirrors `@yellow-org/sdk` (`1.2.1`) so `@yellow-org/sdk-mcp@^1` tracks v1 SDK docs.
- Commit was created with `--no-gpg-sign` because the local 1Password SSH signing agent failed during commit creation.

## Release Notes

* **New Features**
  * Added `server_info` tool providing server metadata and version information.
  * Enabled npm package distribution with automated MCP registry publishing.
  * Runtime now prefers local sources with fallback to packaged content snapshots.

* **Documentation**
  * Expanded README with setup guides for multiple MCP clients (Claude, Cursor, VS Code).
  * Added publishing documentation and version policy.

* **Chores**
  * Version bumped to 1.2.1.
  * Enhanced package validation and verification workflows.

### DIFF
--- a/.github/workflows/publish-sdk-mcp.yml
+++ b/.github/workflows/publish-sdk-mcp.yml
@@ -1,0 +1,177 @@
+name: Publish SDK MCP
+
+on:
+  push:
+    tags:
+      - 'mcp-v*'
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  publish:
+    name: Publish @yellow-org/sdk-mcp
+    runs-on: ubuntu-latest
+    env:
+      MCP_PUBLISHER_VERSION: v1.7.6
+      MCP_PUBLISHER_SHA256_LINUX_AMD64: bcc96ca630cae4cf503b4550bd4a17462d42ad4819273bee56f4385bb059ddee
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          registry-url: 'https://registry.npmjs.org'
+          cache: npm
+          cache-dependency-path: 'sdk/mcp/package-lock.json'
+
+      - name: Install dependencies
+        working-directory: ./sdk/mcp
+        run: npm ci
+
+      - name: Build package
+        working-directory: ./sdk/mcp
+        run: npm run build
+
+      - name: Pack and verify package contents
+        id: pack
+        working-directory: ./sdk/mcp
+        run: |
+          PACK_JSON=$(npm pack --json)
+          printf '%s\n' "$PACK_JSON" > pack.json
+          TARBALL=$(node -e "const pack = require('./pack.json')[0]; console.log(pack.filename)")
+          echo "tarball=$TARBALL" >> "$GITHUB_OUTPUT"
+
+          node - <<'NODE'
+          const pack = require('./pack.json')[0];
+          const paths = new Set(pack.files.map((file) => file.path));
+          const required = [
+            'dist/index.js',
+            'dist/index.d.ts',
+            'content/docs/api.yaml',
+            'content/docs/protocol/overview.md',
+            'content/sdk/ts/package.json',
+            'content/sdk/ts/src/client.ts',
+            'content/sdk/ts-compat/package.json',
+            'content/sdk/ts-compat/src/client.ts',
+            'content/sdk/go/client.go',
+            'README.md',
+            'package.json',
+            'server.json',
+          ];
+
+          for (const path of required) {
+            if (!paths.has(path)) throw new Error(`missing package file: ${path}`);
+          }
+
+          for (const path of paths) {
+            if (path.startsWith('src/') || path.startsWith('scripts/')) {
+              throw new Error(`source-only file leaked into package: ${path}`);
+            }
+            if (path.endsWith('.test.ts') || path.endsWith('.spec.ts') || path.endsWith('_test.go')) {
+              throw new Error(`test file leaked into package: ${path}`);
+            }
+          }
+          NODE
+
+      - name: Smoke test packed MCP server
+        working-directory: ./sdk/mcp
+        run: |
+          set -euo pipefail
+          TMPDIR=$(mktemp -d)
+          mkdir "$TMPDIR/app"
+          EXPECTED_VERSION=$(node -p "require('./package.json').version")
+          npm --prefix "$TMPDIR/app" install "./${{ steps.pack.outputs.tarball }}"
+          BIN="$TMPDIR/app/node_modules/.bin/yellow-sdk-mcp"
+          cd "$TMPDIR/app"
+          BIN="$BIN" EXPECTED_VERSION="$EXPECTED_VERSION" node --input-type=module <<'NODE'
+          import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+          import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';
+
+          const transport = new StdioClientTransport({ command: process.env.BIN, args: [] });
+          const client = new Client({ name: 'sdk-mcp-smoke', version: '0.0.0' });
+          await client.connect(transport);
+
+          const tools = await client.listTools();
+          if (!tools.tools.some((tool) => tool.name === 'server_info')) {
+            throw new Error('server_info tool missing');
+          }
+
+          const infoResult = await client.callTool({ name: 'server_info', arguments: {} });
+          const info = JSON.parse(infoResult.content[0].text);
+          if (info.version !== process.env.EXPECTED_VERSION) {
+            throw new Error(`unexpected server version ${info.version}`);
+          }
+          if (info.sdkVersion !== process.env.EXPECTED_VERSION) {
+            throw new Error(`unexpected sdk version ${info.sdkVersion}`);
+          }
+          if (info.contentMode !== 'packaged-snapshot') {
+            throw new Error(`expected packaged-snapshot mode, got ${info.contentMode}`);
+          }
+
+          const methodResult = await client.callTool({
+            name: 'lookup_method',
+            arguments: { name: 'deposit', language: 'typescript' },
+          });
+          if (!methodResult.content[0].text.includes('deposit')) {
+            throw new Error('lookup_method did not return packaged TypeScript SDK content');
+          }
+
+          const rpcResult = await client.callTool({
+            name: 'lookup_rpc_method',
+            arguments: { method: 'channels.v1.get_home_channel' },
+          });
+          if (!rpcResult.content[0].text.includes('channels.v1.get_home_channel')) {
+            throw new Error('lookup_rpc_method did not return packaged API docs content');
+          }
+
+          await client.close();
+          NODE
+
+      - name: Validate MCP registry metadata
+        working-directory: ./sdk/mcp
+        run: |
+          node - <<'NODE'
+          const pkg = require('./package.json');
+          const server = require('./server.json');
+          const npmPackage = server.packages.find((entry) => entry.registryType === 'npm');
+
+          if (server.name !== pkg.mcpName) {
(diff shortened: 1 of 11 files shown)


.github/workflows/publish-sdk-mcp.yml
```diff
@@ -0,0 +1,177 @@
...
+          tar xzf "$ASSET" mcp-publisher
+          chmod +x mcp-publisher
+
+      - name: Authenticate to MCP Registry
+        working-directory: ./sdk/mcp
+        run: ./mcp-publisher login github-oidc
+
+      - name: Publish MCP registry metadata
+        working-directory: ./sdk/mcp
+        run: ./mcp-publisher publish
```
nksazonov, line 266: If npm publish succeeds but the MCP registry step fails (OIDC expiry, network error, registry outage), there is no way to retry just the registry step. Re-triggering the workflow on the same tag fails at npm publish with a version-conflict error — npm rejects duplicate version uploads. Recovery requires running `mcp-publisher publish` by hand outside CI, with no traceability.

Split npm publish and MCP registry publish into separate jobs so each can be re-run independently from the GitHub Actions UI. This also makes it easier to diagnose which step failed from the Actions summary.

.github/workflows/publish-sdk-mcp.yml
```diff
@@ -0,0 +1,177 @@
...
+          }
+
+          for (const path of paths) {
+            if (path.startsWith('src/') || path.startsWith('scripts/')) {
+              throw new Error(`source-only file leaked into package: ${path}`);
+            }
+            if (path.endsWith('.test.ts') || path.endsWith('.spec.ts') || path.endsWith('_test.go')) {
+              throw new Error(`test file leaked into package: ${path}`);
+            }
+          }
```
nksazonov: The `files` field in `package.json` is an explicit allowlist (`dist/`, `content/`, `README.md`, `server.json`). npm only includes paths that match the allowlist — `src/` and `scripts/` are excluded by definition. These assertions can never trigger; they are dead code that looks like an active safety check but provides no real protection.

Either remove the assertions and rely on the `files` allowlist, or add a comment explaining they are defense-in-depth against someone accidentally extending `files` in the future. Without context, readers assume they guard against something real.

.github/workflows/publish-sdk-mcp.yml
```diff
@@ -0,0 +1,177 @@
...
+          NODE
+
+      - name: Smoke test packed MCP server
+        working-directory: ./sdk/mcp
+        run: |
+          set -euo pipefail
+          TMPDIR=$(mktemp -d)
```
nksazonov: `TMPDIR` is a POSIX-standard environment variable that npm and some Node.js internals use to locate temporary storage. Shadowing it within the step means any tool invoked in the same shell that respects `TMPDIR` could redirect its temp files into the smoke-test app directory. On Ubuntu CI `TMPDIR` is typically unset, so the immediate risk is low — but it is still a footgun that could silently corrupt a future step if tooling changes.

Use a name that does not conflict with system conventions, such as `SMOKE_DIR` or `TMP_INSTALL_DIR`.

.github/workflows/publish-sdk-mcp.yml
```diff
@@ -0,0 +1,177 @@
...
+          }
+
+          const infoResult = await client.callTool({ name: 'server_info', arguments: {} });
+          const info = JSON.parse(infoResult.content[0].text);
+          if (info.version !== process.env.EXPECTED_VERSION) {
+            throw new Error(`unexpected server version ${info.version}`);
+          }
+          if (info.sdkVersion !== process.env.EXPECTED_VERSION) {
+            throw new Error(`unexpected sdk version ${info.sdkVersion}`);
+          }
```
nksazonov: The smoke test asserts that `info.sdkVersion` exactly equals `EXPECTED_VERSION` (the MCP package version). This works under the current version policy where SDK and MCP versions are always identical.

The assertion conflates two concerns: (1) verifying the server correctly reads the content snapshot, and (2) enforcing the version-coupling policy. If the project ever needs an MCP-only patch release — fixing a packaging or workflow bug without a corresponding SDK change — the versions would diverge and this single assertion would block the entire publish, even though the snapshot is being read correctly.

Separate the checks: assert that `sdkVersion` is a non-empty semver string (validating snapshot readability), and document the version-coupling constraint as a release process requirement rather than a CI gate.

sdk/mcp/scripts/prepare-package-content.mjs
```diff
@@ -0,0 +1,67 @@
...
+
+rmSync(contentRoot, { recursive: true, force: true });
+mkdirSync(contentRoot, { recursive: true });
+
+for (const spec of copySpecs) {
+  const sourcePath = resolve(repoRoot, spec.from);
+  const targetPath = resolve(contentRoot, spec.to);
+  if (!existsSync(sourcePath)) {
+    console.warn(`Skipping missing content source: ${relative(repoRoot, sourcePath)}`);
+    continue;
```
nksazonov, line 120: When a `copySpecs` source path is absent, the script logs a warning to stderr and continues. `npm run build` exits 0 even if the content snapshot is incomplete.

A missing required source (e.g., `sdk/ts/src` not present) silently produces a partial snapshot. The failure only surfaces later during `npm pack` verification — after TypeScript compilation has already run — making the error harder to trace. The build output and the actual failure are decoupled.

Distinguish required from optional sources. For required ones, exit with a non-zero code immediately so `npm run build` itself reports the failure at the right moment.

sdk/mcp/package.json
```diff
@@ -1,16 +1,26 @@
...
+    "content",
+    "README.md",
+    "server.json"
+  ],
   "scripts": {
     "start": "tsx src/index.ts",
-    "build": "tsc",
-    "typecheck": "tsc --noEmit"
+    "prepare-content": "node scripts/prepare-package-content.mjs",
+    "build": "npm run prepare-content && tsc && chmod +x dist/index.js",
```
nksazonov: `chmod` does not exist on Windows. Any Windows developer running `npm run build` locally gets a hard failure after compilation succeeds — the binary is present and valid, but the script terminates with an unrecognized command error, making the built output unusable. The execute bit is irrelevant on Windows anyway.

Append `|| true` to the `chmod` call to gracefully no-op on platforms that do not support it, or use a cross-platform equivalent via `node -e "require('fs').chmodSync('dist/index.js', 0o755)""`.

sdk/mcp/src/index.ts
```diff
@@ -23,15 +25,60 @@ const API_YAML = resolve(REPO_ROOT, 'docs/api.yaml');
 const GO_SDK_ROOT = resolve(REPO_ROOT, 'sdk/go');
 const PKG_ROOT = resolve(REPO_ROOT, 'pkg');
 const GO_MODULE_PATH = 'github.com/layer-3/nitrolite';
-const GO_MODULE_VERSION = 'v1.2.0';
+const GO_MODULE_VERSION = 'v1.2.1';
```
nksazonov: `sdkVersion` and `compatVersion` are dynamically read from their respective `package.json` files at startup. `GO_MODULE_VERSION` is a hardcoded string constant returned by `server_info` and embedded in Go scaffolding templates.

When the Go module receives a new release, this constant must be manually updated alongside the normal version-bump files. The inconsistency with how TS SDK versions are sourced means `goModuleVersion` in `server_info` is likely to drift — a user calling `server_info` to diagnose a problem could get a stale value.

Read the version from `go.mod` at startup (parsing the `module` line is a one-liner), or include a generated version file in the content snapshot so all version fields share the same dynamic sourcing strategy.

.github/workflows/publish-sdk-mcp.yml
ihsraham: addressed in [d565102d](https://github.com/layer-3/nitrolite/commit/d565102d7985ace650da5f933b7602e2f5e131eb): npm publish and mcp registry publish are now separate jobs, so the registry job can be rerun after npm succeeds.

.github/workflows/publish-sdk-mcp.yml
ihsraham: addressed in [d565102d](https://github.com/layer-3/nitrolite/commit/d565102d7985ace650da5f933b7602e2f5e131eb): moved this into `verify-package-content.mjs` and added a defense-in-depth comment so it is clear why the leak checks exist.

.github/workflows/publish-sdk-mcp.yml
ihsraham: addressed in [d565102d](https://github.com/layer-3/nitrolite/commit/d565102d7985ace650da5f933b7602e2f5e131eb): renamed it to `SMOKE_DIR` in the smoke job.

.github/workflows/publish-sdk-mcp.yml
ihsraham: i kept this as a ci gate intentionally. for this package we want deterministic strict-mirror releases, and [d565102d](https://github.com/layer-3/nitrolite/commit/d565102d7985ace650da5f933b7602e2f5e131eb) makes that policy explicit in `release.json`, preflight, and the readme. if we need mcp-only patch releases later, we should add a compatibility matrix first.

sdk/mcp/scripts/prepare-package-content.mjs
ihsraham: addressed in [d565102d](https://github.com/layer-3/nitrolite/commit/d565102d7985ace650da5f933b7602e2f5e131eb): copy specs are now required by default and the build fails immediately when a required source is missing or copies zero files.

sdk/mcp/package.json
ihsraham: addressed in [d565102d](https://github.com/layer-3/nitrolite/commit/d565102d7985ace650da5f933b7602e2f5e131eb): replaced shell `chmod` with `node scripts/set-executable.mjs dist/index.js`, which also fails if the binary is missing.

sdk/mcp/src/index.ts
ihsraham: addressed in [d565102d](https://github.com/layer-3/nitrolite/commit/d565102d7985ace650da5f933b7602e2f5e131eb): removed the hardcoded go version and generated `content/release.json`; `server_info` and go scaffolds now read the packaged `goModuleVersion`.

nksazonov (approved):
Well done!

PR #717, opened 2026-04-30
YNU-779: feat(nitronode/metrics): saturation + DB latency + label hygiene

philanton:
## Summary

Extends nitronode's emitted metric set and standardizes its label naming so dashboards and alerts can be authored against a coherent surface. Three commits, each independently reviewable:

1. **`2aba0e3d` feat(nitronode/metrics): add saturation + DB latency observability** — three new emission paths:
   - `nitronode_rpc_inflight{method}` — gauge incremented at the `ObservabilityMiddleware` entry and decremented on exit. Saturation signal that pairs with `rpc_request_duration_seconds` to distinguish "slow handler" from "queued behind contention".
   - `nitronode_db_query_duration_seconds{query_kind}` — histogram observed via gorm callbacks (Create / Query / Update / Delete / Row / Raw). Captures app-side round-trip including pgbouncer hop.
   - `go_sql_*` — standard `collectors.NewDBStatsCollector` registered with db name `nitronode`. Pool gauges + wait counters (`go_sql_wait_duration_seconds_total`, `go_sql_wait_count_total`, etc.) for the underlying `*sql.DB`.

   New `store/database/metrics.go` owns the gorm-callback registration and dbstats collector helpers, exposing a tiny `QueryDurationObserver` interface so the package doesn't depend on `nitronode/metrics` (which would induce a cycle through `pkg/{app,core,rpc}`).

2. **`921a00b0` refactor(nitronode/metrics): standardize label naming + scrub appID** — coordinated **breaking** label renames:
   - `application` → `application_id` on `nitronode_app_sessions_total`, `nitronode_app_sessions_active`, `nitronode_app_state_updates_total`, `nitronode_app_session_update_sig_validations_total` (the holdouts; `user_states_total` / `transactions_total` already used `application_id`).
   - `status` → `result` on `nitronode_rpc_requests_total` and `nitronode_rpc_request_duration_seconds`; `process_result` → `result` on `nitronode_blockchain_events_total`. Settles on `result ∈ {success, failed}` everywhere.
   - `IncAppStateUpdate`, `IncAppSessionUpdateSigValidation`, `SetAppSessions`, `SetActiveAppSessions` now wrap `applicationID` with `getApplicationIDLabelValue` so empty becomes `_DEFAULT` (matches `IncUserState` / `RecordTransaction`).
   - `nitronode_rpc_messages_total` → `nitronode_rpc_messages_emitted_total`. The metric counts both the request and the response message of every RPC; the new name and Help spell out the doubled emission so downstream consumers don't compute QPS at 2× the real rate.

3. **`196190ca` docs(nitronode/metrics): expand Help texts to encode label semantics** — Help string rewrite covering trigger conditions, label semantics with enum value sets, and aggregation guidance for every metric. Notable:
   - `user_balance_underfunded` / `user_balance_releasable` now explain the channel-checkpoint semantics: each is the worst-case potential outflow / inflow from `node_balance` if every underfunded / releasable channel checkpointed simultaneously.
   - `transactions_amount_total` documents the `decimal.Decimal.InexactFloat64` precision loss explicitly.

## Verified

- `go build ./nitronode/...` clean.
- `go test ./nitronode/...` clean (existing test suite — no behaviour-affecting changes; rename-only refactors covered by package-level builds).
- DB instrumentation does not introduce per-call allocation beyond a single `time.Time` stashed in the gorm context dict.

## Migration

Breaking — every dashboard / alert / recording rule that referenced the old label names or `nitronode_rpc_messages_total` must update in lockstep with the deploy. The ops-side changes (dashboard YAML, alert rules) follow on `feat/metrics-extensions` in `layer-3/ops`.

## Deferred

- `D N7` (PodMonitoring `metricRelabeling` guard against `application_id` cardinality blow-up) is captured in `docs/nitronode-env-metrics.md` §7 / §9 in the ops repo — postponed until cardinality becomes a real ingest cost.

## Summary by CodeRabbit

* **New Features**
  * In-flight RPC request tracking added to monitor concurrent requests per RPC method
  * Database query duration metrics and DB pool stats collection registered for deeper DB performance insights

* **Chores**
  * Standardized application metric label name to application_id
  * Renamed and clarified RPC metric names and result semantics for improved observability

### DIFF
--- a/nitronode/store/database/metrics.go
+++ b/nitronode/store/database/metrics.go
@@ -1,0 +1,112 @@
+package database
+
+import (
+	"errors"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
+	"gorm.io/gorm"
+)
+
+// QueryDurationObserver records the time a single gorm DB operation took.
+// Implemented by the runtime metric exporter to avoid an import cycle
+// (the metrics package depends on pkg/{app,core,rpc}, store/database does not).
+type QueryDurationObserver interface {
+	ObserveDBQueryDuration(queryKind string, duration time.Duration)
+}
+
+// metricsStartKey is the gorm.DB context key the before-callback uses to stash
+// the operation start timestamp for the after-callback to read.
+const metricsStartKey = "nitronode:metrics:start"
+
+// callbackName is registered on every gorm callback chain we instrument; if
+// you change it remember to keep the before/after pair in sync.
+const callbackName = "nitronode:metrics"
+
+// queryKinds is the set of gorm callback chains we hook. Order matches the
+// gorm callback registry; "raw" covers Raw / Exec catch-all on the gorm v2
+// callback chain "raw".
+var queryKinds = []string{"create", "query", "update", "delete", "row", "raw"}
+
+// RegisterMetricsCallbacks installs gorm callbacks that observe wall-clock
+// duration of each Create / Query / Update / Delete / Row / Raw operation
+// onto obs. Pass nil to skip registration (test / sqlite-in-memory cases).
+//
+// The callbacks add no per-call allocation beyond a single time.Time stashed
+// in the gorm context dict.
+func RegisterMetricsCallbacks(db *gorm.DB, obs QueryDurationObserver) error {
+	if db == nil {
+		return errors.New("database: nil gorm.DB")
+	}
+	if obs == nil {
+		return nil
+	}
+
+	for _, kind := range queryKinds {
+		kind := kind // pin for closures
+		chain := db.Callback().Query()
+		switch kind {
+		case "create":
+			chain = db.Callback().Create()
+		case "update":
+			chain = db.Callback().Update()
+		case "delete":
+			chain = db.Callback().Delete()
+		case "row":
+			chain = db.Callback().Row()
+		case "raw":
+			chain = db.Callback().Raw()
+		}
+
+		if err := chain.Before("*").Register(callbackName+":before:"+kind, func(tx *gorm.DB) {
+			tx.Set(metricsStartKey, time.Now())
+		}); err != nil {
+			return err
+		}
+		if err := chain.After("*").Register(callbackName+":after:"+kind, func(tx *gorm.DB) {
+			v, ok := tx.Get(metricsStartKey)
+			if !ok {
+				return
+			}
+			start, ok := v.(time.Time)
+			if !ok {
+				return
+			}
+			obs.ObserveDBQueryDuration(kind, time.Since(start))
+		}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// RegisterDBStatsCollector registers the standard go_sql_* collector on reg
+// for the underlying *sql.DB, so pool-state gauges and wait counters become
+// scrapable. Returns an error if the gorm DB doesn't expose a *sql.DB
+// (sqlite-in-memory variants normally do; misconfigured pools don't).
+//
+// Emits, under "nitronode" db name:
+//
+//	go_sql_max_open_connections
+//	go_sql_open_connections
+//	go_sql_in_use_connections
+//	go_sql_idle_connections
+//	go_sql_wait_count_total
+//	go_sql_wait_duration_seconds_total
+//	go_sql_max_idle_closed_total
+//	go_sql_max_idle_time_closed_total
+//	go_sql_max_lifetime_closed_total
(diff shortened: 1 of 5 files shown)


nitronode/store/database/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+		return errors.New("database: nil gorm.DB")
+	}
+	if reg == nil {
+		return errors.New("database: nil prometheus registerer")
+	}
+	sqlDB, err := db.DB()
+	if err != nil {
+		return err
+	}
+	return reg.Register(collectors.NewDBStatsCollector(sqlDB, "nitronode"))
```
nksazonov, line 125: `reg.Register()` returns `prometheus.AlreadyRegisteredError` when a collector with the same descriptor is already registered — which happens in any test that constructs the server component more than once in-process. The caller in `runtime.go` passes all errors straight to `logger.Fatal`, so a second initialisation kills the process with no useful diagnostic.

The standard Prometheus pattern is to treat `AlreadyRegisteredError` as success (the collector is already there and working):
```go
   if err := reg.Register(collectors.NewDBStatsCollector(sqlDB, "nitronode")); err != nil {
       var already prometheus.AlreadyRegisteredError
       if errors.As(err, &already) {
           return nil
       }
       return err
   }
   return nil
```

nitronode/store/database/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+	if db == nil {
+		return errors.New("database: nil gorm.DB")
+	}
+	if obs == nil {
+		return nil
+	}
+
+	for _, kind := range queryKinds {
+		kind := kind // pin for closures
+		chain := db.Callback().Query()
```
nksazonov, line 54: The switch that selects the gorm callback chain pre-initialises chain to `db.Callback().Query()` before the switch, with no case "query": "query" works only because it falls through to the pre-set value. If `queryKinds` ever gains a new string without a matching switch arm, that kind silently hooks the query processor instead of its own chain, reporting the wrong `query_kind` label with no error or warning.

Adding an explicit case "query": and a default: that returns an error would make any future omission loud:
```go
   case "query":
       chain = db.Callback().Query()
   default:
       return fmt.Errorf("database: unknown query kind %q", kind)
```

nitronode/store/database/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+			chain = db.Callback().Update()
+		case "delete":
+			chain = db.Callback().Delete()
+		case "row":
+			chain = db.Callback().Row()
+		case "raw":
+			chain = db.Callback().Raw()
+		}
+
+		if err := chain.Before("*").Register(callbackName+":before:"+kind, func(tx *gorm.DB) {
```
nksazonov: Here and on line 67 (the "after" callback):
gorm's `callback.Register()` appends when a callback with the same name already exists — it logs a Warn but does not replace and does not error. If `RegisterMetricsCallbacks` is ever called twice on the same `*gorm.DB (test helpers, DI restart, future refactor)`, every after-callback fires twice per query. The `db_query_duration_seconds` histogram would observe double the actual duration on every operation, silently producing wrong data.

`chain.Replace(...)` is idempotent by design and is a drop-in substitute for `chain.Register(...)` in both the before and after registrations.

nksazonov (approved):
LGTM with some comments

nitronode/store/database/metrics.go
philanton: Not for us, I don't see any point in having a few registrations per descriptor, we should see them

nitronode/store/database/metrics.go
philanton: Applied in 01f12f10 — `fix(nitronode/metrics): make gorm callback registration explicit + idempotent`.

Added explicit `case "query":` + a `default:` that returns `fmt.Errorf("database: unknown query kind %q", kind)`, so any new entry to `queryKinds` without a matching switch arm now fails loudly at runtime instead of silently hooking the query processor with the wrong `query_kind` label.

Side note: the dummy `chain := db.Callback().Query()` init is still there because gorm's `processor` / `callback` types are unexported — `var chain *gorm.processor` doesn't compile from outside the package. The default branch returning an error is what closes the silent-fallthrough hole.

nitronode/store/database/metrics.go
philanton: Applied in 01f12f10 — `fix(nitronode/metrics): make gorm callback registration explicit + idempotent`.

Switched both the before and the after callback registration to `chain.Replace(...)`. Confirmed by reading `gorm.io/gorm@v1.31.1/callbacks.go` that `Replace` is idempotent on the callback name and a drop-in for `Register` here. No more double-fire risk on a repeat `RegisterMetricsCallbacks` call.

PR #723, opened 2026-05-06
MF-C01: rpc: cap inbound WebSocket frame size and rate-limit per connection

philanton:
## Summary
- Cap inbound WebSocket frame size at the transport layer via `gorilla.SetReadLimit` (default 128 KiB). Frames over the cap close the connection with code 1009 before any allocation.
- Add a pluggable `FrameRateLimiter` interface (`pkg/rpc/rate_limiter.go`) with a `ByteTokenBucket` implementation. Per-connection byte budget (default 256 KiB/s steady, 1 MiB burst) consulted before each frame is dispatched.
- Wire knobs through `WebsocketNodeConfig` and nitronode `runtime.go` (`NITRONODE_WS_MAX_MESSAGE_SIZE`, `NITRONODE_WS_BYTES_PER_SEC`, `NITRONODE_WS_BYTES_BURST`).
- Unit tests cover bucket burst/refill/cap/overflow + connection behavior on oversized frame and limiter rejection.
- stress-v1 chart picks up the new envs and gains NGINX ingress annotations (`limit-connections`, `limit-rps`, `limit-burst-multiplier`, `proxy-body-size`). Chart README documents the layered defense (Cloudflare → ingress-nginx → app).

## Why
Audit finding: the RPC server unmarshals incoming WebSocket frames before any size or rate check. An unauthenticated attacker could send oversized frames or sustained large traffic and exhaust memory/CPU before the per-message rate-limit middleware ran (it executes inside `processRequests`, after the first JSON unmarshal). The new caps stop the work at the transport layer.

## Companion PR
Pairs with [layer-3/ops#15](https://github.com/layer-3/ops/pull/15) which preserves real client IPs through ingress-nginx so the per-IP annotations on the WebSocket Ingress key on real clients.

## Notable
- Latent bug fixed: `readMessages` did not call `handleClosure` on the queue-full path, so `Serve`'s `wg.Wait` could hang on that exit.
- Defaults sized for browser reload reconnect storms (auth + subscribe ≈ 20 KB / tab) — they clear the bucket instantly.
- Disable the byte-rate cap with `NITRONODE_WS_BYTES_PER_SEC=-1` for canary rollout. Frame size cap stays on regardless.

## Summary by CodeRabbit

* **New Features**
  * WebSocket connection hardening with configurable message size limits and per-connection byte-rate limiting to mitigate denial-of-service scenarios.

* **Documentation**
  * Added WebSocket DoS hardening guidance including WAF rate-limiting recommendations and NGINX Ingress configuration examples.

* **Tests**
  * Added comprehensive tests for WebSocket rate limiting and oversized frame handling.

### DIFF
--- a/nitronode/chart/config/stress-v1/nitronode.yaml.gotmpl
+++ b/nitronode/chart/config/stress-v1/nitronode.yaml.gotmpl
@@ -83,8 +83,10 @@ networking:
       nginx.ingress.kubernetes.io/limit-connections: "50"        # concurrent / IP
       nginx.ingress.kubernetes.io/limit-rps: "20"                # new conns/s / IP
       nginx.ingress.kubernetes.io/limit-burst-multiplier: "3"
-      # Belt-and-braces: ingress-level body cap. App enforces frame size too.
-      nginx.ingress.kubernetes.io/proxy-body-size: "256k"
+      # Note: nginx proxy-body-size applies to HTTP request bodies only. After
+      # the WebSocket upgrade, ingress-nginx proxies the TCP stream transparently
+      # and cannot enforce a per-frame size limit. The frame cap is enforced at
+      # the application layer via NITRONODE_WS_MAX_MESSAGE_SIZE / SetReadLimit.
 
 stressTest:
   enabled: false
--- a/nitronode/chart/config/stress-v1/nitronode.yaml.gotmpl
+++ b/nitronode/chart/config/stress-v1/nitronode.yaml.gotmpl
@@ -21,6 +21,13 @@ config:
     NITRONODE_MAX_SESSION_DATA_LEN: "1024"
     NITRONODE_MAX_SIGNED_UPDATES: "0"
     NITRONODE_MAX_SESSION_KEY_IDS: "256"
+    # WebSocket DoS hardening.
+    # Frame cap: largest legit v1 RPC ≈ a few KB; 128 KiB leaves headroom.
+    # Byte budget: 256 KiB/s steady, 1 MiB burst. Comfortably absorbs reload
+    # reconnect storms (auth + subscribe ≈ 20 KB / tab) without throttling.
+    NITRONODE_WS_MAX_MESSAGE_SIZE: "131072"
+    NITRONODE_WS_BYTES_PER_SEC: "262144"
+    NITRONODE_WS_BYTES_BURST: "1048576"
 
 image:
   repository: ghcr.io/layer-3/nitrolite/nitronode
@@ -66,6 +73,18 @@ networking:
       nginx.ingress.kubernetes.io/proxy-send-timeout: "3600"
       nginx.ingress.kubernetes.io/proxy-connect-timeout: "10"
       nginx.ingress.kubernetes.io/proxy-buffering: "off"
+      # ── DoS hardening (per-IP, NGINX-level) ──────────────────────────────
+      # Caps abusive sources before WebSocket upgrade and JSON decode.
+      # Effective only when ingress-nginx sees real client IPs — see chart
+      # README "WebSocket DoS hardening" for required ConfigMap settings.
+      # stress-v1 has no Cloudflare in front; source IP comes from the GCP
+      # LB. Verify externalTrafficPolicy=Local on the ingress-nginx Service
+      # before relying on per-IP limits.
+      nginx.ingress.kubernetes.io/limit-connections: "50"        # concurrent / IP
+      nginx.ingress.kubernetes.io/limit-rps: "20"                # new conns/s / IP
+      nginx.ingress.kubernetes.io/limit-burst-multiplier: "3"
+      # Belt-and-braces: ingress-level body cap. App enforces frame size too.
+      nginx.ingress.kubernetes.io/proxy-body-size: "256k"
 
 stressTest:
   enabled: false
--- a/nitronode/runtime.go
+++ b/nitronode/runtime.go
@@ -205,6 +205,26 @@ func InitBackbone() *Backbone {
 
 	bytesPerSec := conf.WsBytesPerSec
 	bytesBurst := conf.WsBytesBurst
+	// When the per-connection byte budget is enabled, the burst must be at least
+	// the max message size; otherwise every legitimate frame that passes
+	// SetReadLimit would still be rejected by the bucket on arrival, taking the
+	// node out via config alone. Fail fast at startup.
+	if bytesPerSec > 0 {
+		if bytesBurst <= 0 {
+			logger.Fatal(
+				"NITRONODE_WS_BYTES_BURST must be > 0 when NITRONODE_WS_BYTES_PER_SEC is enabled",
+				"ws_bytes_burst", bytesBurst,
+				"ws_bytes_per_sec", bytesPerSec,
+			)
+		}
+		if conf.WsMaxMessageSize > 0 && bytesBurst < float64(conf.WsMaxMessageSize) {
+			logger.Fatal(
+				"NITRONODE_WS_BYTES_BURST must be >= NITRONODE_WS_MAX_MESSAGE_SIZE when byte limiting is enabled",
+				"ws_bytes_burst", bytesBurst,
+				"ws_max_message_size", conf.WsMaxMessageSize,
+			)
+		}
+	}
 	rpcNode, err := rpc.NewWebsocketNode(rpc.WebsocketNodeConfig{
 		Logger:                  logger,
 		ObserveConnections:      runtimeMetrics.SetRPCConnections,
--- a/nitronode/runtime.go
+++ b/nitronode/runtime.go
@@ -80,6 +80,14 @@ type FullConfig struct {
 	RateLimitBurst              float64          `yaml:"rate_limit_burst" env:"NITRONODE_RATE_LIMIT_BURST" env-default:"20"`
 	WsProcessBufferSize         int              `yaml:"ws_process_buffer_size" env:"NITRONODE_WS_PROCESS_BUFFER_SIZE" env-default:"64"`
 	WsWriteBufferSize           int              `yaml:"ws_write_buffer_size" env:"NITRONODE_WS_WRITE_BUFFER_SIZE" env-default:"64"`
+	// WsMaxMessageSize caps inbound WebSocket frame size in bytes. Frames over the
+	// cap close the connection with WebSocket close code 1009 before allocation.
+	// 128 KiB fits any legitimate v1 RPC with substantial headroom.
+	WsMaxMessageSize int64 `yaml:"ws_max_message_size" env:"NITRONODE_WS_MAX_MESSAGE_SIZE" env-default:"131072"`
+	// WsBytesPerSec is the steady-state byte budget per connection. Set <0 to disable.
+	WsBytesPerSec float64 `yaml:"ws_bytes_per_sec" env:"NITRONODE_WS_BYTES_PER_SEC" env-default:"262144"`
+	// WsBytesBurst is the burst capacity of the per-connection byte bucket.
+	WsBytesBurst float64 `yaml:"ws_bytes_burst" env:"NITRONODE_WS_BYTES_BURST" env-default:"1048576"`
 }
 
 type SignerConfig struct {
@@ -195,11 +203,20 @@ func InitBackbone() *Backbone {
 	// RPC Node
 	// ------------------------------------------------
 
+	bytesPerSec := conf.WsBytesPerSec
+	bytesBurst := conf.WsBytesBurst
 	rpcNode, err := rpc.NewWebsocketNode(rpc.WebsocketNodeConfig{
 		Logger:                  logger,
 		ObserveConnections:      runtimeMetrics.SetRPCConnections,
 		WsConnProcessBufferSize: conf.WsProcessBufferSize,
 		WsConnWriteBufferSize:   conf.WsWriteBufferSize,
+		WsConnMaxMessageSize:    conf.WsMaxMessageSize,
+		NewFrameRateLimiter: func() rpc.FrameRateLimiter {
+			if bytesPerSec <= 0 {
+				return rpc.NoopFrameRateLimiter{}
+			}
+			return rpc.NewByteTokenBucket(bytesPerSec, bytesBurst)
+		},
 	})
 	if err != nil {
 		logger.Fatal("failed to initialize RPC node", "error", err)
--- a/nitronode/runtime.go
+++ b/nitronode/runtime.go
@@ -203,6 +203,16 @@ func InitBackbone() *Backbone {
 	// RPC Node
 	// ------------------------------------------------
 
+	// MF-C01 requires a hard frame-size cap; refuse to start without one even if
+	// the operator zeroed the env. The library would substitute its own default,
+	// but we want the misconfiguration surfaced rather than papered over.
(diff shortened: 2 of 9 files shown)


philanton:
@coderabbitai full review please

nitronode/runtime.go
```diff
@@ -80,6 +80,14 @@ type FullConfig struct {
 	RateLimitBurst              float64          `yaml:"rate_limit_burst" env:"NITRONODE_RATE_LIMIT_BURST" env-default:"20"`
 	WsProcessBufferSize         int              `yaml:"ws_process_buffer_size" env:"NITRONODE_WS_PROCESS_BUFFER_SIZE" env-default:"64"`
 	WsWriteBufferSize           int              `yaml:"ws_write_buffer_size" env:"NITRONODE_WS_WRITE_BUFFER_SIZE" env-default:"64"`
+	// WsMaxMessageSize caps inbound WebSocket frame size in bytes. Frames over the
+	// cap close the connection with WebSocket close code 1009 before allocation.
+	// 128 KiB fits any legitimate v1 RPC with substantial headroom.
+	WsMaxMessageSize int64 `yaml:"ws_max_message_size" env:"NITRONODE_WS_MAX_MESSAGE_SIZE" env-default:"131072"`
```
ihsraham, line 86: suggestion: reject `NITRONODE_WS_MAX_MESSAGE_SIZE <= 0` at startup. c-01 asks for a hard frame limit, so the app should not have a config value that turns it off by accident. the default is safe, but failing fast makes the audit closure cleaner.

ihsraham:
separate security note: `go.mod:3` uses go `1.25.7`, and `nitronode/dockerfile:1` builds from `golang:1.25-alpine`. `govulncheck` reports go standard library issues fixed in `1.25.9`. this is not part of c-01, but a small follow-up bump would make ci and images use the patched go version.

nksazonov (commented):
-

nitronode/chart/config/stress-v1/nitronode.yaml.gotmpl
```diff
@@ -66,6 +73,18 @@ networking:
...
+      # Effective only when ingress-nginx sees real client IPs — see chart
+      # README "WebSocket DoS hardening" for required ConfigMap settings.
+      # stress-v1 has no Cloudflare in front; source IP comes from the GCP
+      # LB. Verify externalTrafficPolicy=Local on the ingress-nginx Service
+      # before relying on per-IP limits.
+      nginx.ingress.kubernetes.io/limit-connections: "50"        # concurrent / IP
+      nginx.ingress.kubernetes.io/limit-rps: "20"                # new conns/s / IP
+      nginx.ingress.kubernetes.io/limit-burst-multiplier: "3"
+      # Belt-and-braces: ingress-level body cap. App enforces frame size too.
+      nginx.ingress.kubernetes.io/proxy-body-size: "256k"
```
nksazonov: `proxy-body-size` maps to nginx's `client_max_body_size`, which applies only to HTTP request bodies. After the WebSocket upgrade handshake, nginx proxies the TCP stream transparently — this directive has no effect on WebSocket frames. Both the inline comment and `chart/README.md:130` describe this as a defense layer for frame sizes, which is inaccurate. The app-level `SetReadLimit` is the only effective frame cap. Consider removing the annotation or correcting the comment to avoid the misconception that ingress enforces a frame-size limit.

pkg/rpc/connection.go
```diff
@@ -178,17 +197,25 @@ func NewWebsocketConnection(config WebsocketConnectionConfig) (*WebsocketConnect
 	if config.PongTimeout <= 0 {
 		config.PongTimeout = defaultWsConnPongTimeout
 	}
+	if config.MaxMessageSize == 0 {
+		config.MaxMessageSize = defaultWsConnMaxMessageSize
```
nksazonov: The defaulting for `MaxMessageSize` is asymmetric with every other field in this function. All others use `<= 0` → apply default, so a caller passing `-1` expecting a reset gets the default. Here, `-1` permanently disables the frame cap: `== 0` is the only trigger and the guard in `Serve` is `> 0`. A caller following the established pattern in this struct will silently remove all frame-size protection. Align to `<= 0` for the default case, or add a `DisableFrameSizeLimit bool` field to make disabling explicit. Also, the struct-field comment at line 115 ("Zero or negative disables") is inaccurate — after `NewWebsocketConnection`, zero is never stored; it gets replaced with 131072.

nitronode/runtime.go
```diff
@@ -195,11 +203,40 @@ func InitBackbone() *Backbone {
...
+				"ws_bytes_per_sec", bytesPerSec,
+			)
+		}
+		if conf.WsMaxMessageSize > 0 && bytesBurst < float64(conf.WsMaxMessageSize) {
+			logger.Fatal(
+				"NITRONODE_WS_BYTES_BURST must be >= NITRONODE_WS_MAX_MESSAGE_SIZE when byte limiting is enabled",
+				"ws_bytes_burst", bytesBurst,
+				"ws_max_message_size", conf.WsMaxMessageSize,
+			)
+		}
```
nksazonov: The condition `conf.WsMaxMessageSize > 0` skips burst validation when `NITRONODE_WS_MAX_MESSAGE_SIZE=0`. In that case `NewWebsocketConnection` substitutes the 131072-byte default, but the startup check has already passed. If `NITRONODE_WS_BYTES_BURST` is set below 131072 at the same time, every legitimate frame passes `SetReadLimit` and is then rejected by the bucket — the node appears healthy at startup but drops all traffic. The condition should also handle the `== 0` case, substituting the effective default (131072) in the comparison.

pkg/rpc/connection_test.go
```diff
@@ -92,6 +92,144 @@ func TestWebsocketConnection_Serve(t *testing.T) {
...
+
+	// Read limit must be set before any read happens.
+	require.Eventually(t, func() bool {
+		return wsConnMock.getReadLimit() == 64*1024
+	}, 200*time.Millisecond, 5*time.Millisecond)
+}
+
+// TestWebsocketConnection_LocalReadLimit_GracefulClose simulates the local
+// SetReadLimit hitting on an inbound frame. Gorilla returns ErrReadLimit to
+// the application (and best-effort sends close 1009 to the peer over the wire).
```
nksazonov, line 119: `SetReadLimit` is called synchronously in `Serve` before any goroutines are spawned (`connection.go:252-254`). The read limit is already set when `Serve` returns, so `require.Eventually` polling is unnecessary — a direct `require.Equal` immediately after the call is sufficient. The polling implies a race that does not exist and adds ~5 ms of wall-clock overhead per test run.

pkg/rpc/connection.go
```diff
@@ -336,6 +387,19 @@ func (conn *WebsocketConnection) readMessages(handleClosure func(error)) {
...
+				"origin", conn.origin,
+				"frame_bytes", len(messageBytes),
+			)
+			select {
+			case conn.closeConnCh <- struct{}{}:
+			default:
+			}
+			handleClosure(nil)
+			return
+		}
```
nksazonov, line 403: By the time the rate limiter runs, the frame is already allocated in the Go heap. This is intentional (the frame-size cap via `SetReadLimit` prevents any single frame from being large), but it means a burst of back-to-back max-sized frames (128 KiB each) can briefly consume up to `burst` bytes (1 MiB by default) of per-connection memory before the connection is dropped. This trade-off is worth documenting explicitly in the `FrameRateLimiter` interface comment, since implementers may assume allocation is prevented before `Allow` is called.

pkg/rpc/rate_limiter.go
```diff
@@ -0,0 +1,63 @@
...
+
+// FrameRateLimiter decides whether an inbound WebSocket frame is admitted.
+// Implementations attached to a single connection may assume serial access
+// from the connection's read goroutine; implementations shared across
+// connections must be safe for concurrent use.
+//
+// Returning false causes the connection to close.
+type FrameRateLimiter interface {
+	// Allow reports whether a frame of size bytes is permitted at now.
+	Allow(now time.Time, size int) bool
```
nksazonov: `Allow` reads as an imperative command ("allow this frame") rather than a query. A reader calling `if !conn.frameRateLimiter.Allow(...)` has to mentally invert the name. `Admit` is the standard term in admission control contexts, makes the call site read naturally (`if !limiter.Admit(now, size)`), and better matches the domain — this is an admission control interface, not a permission-granting one. `golang.org/x/time/rate` uses `Allow` too, but new interfaces do not need to inherit the ambiguity.

nksazonov (commented):
The core implementation — `SetReadLimit`, `ByteTokenBucket`, startup validation, and the latent `wg` hang fix — is correct. Control flow through all close paths is sound (exactly 3 `childHandleClosure` calls for `wg(3)`). Tests cover the important paths. Issues below are scoped to a documentation/configuration correctness gap (`proxy-body-size`), a design consistency gap (`MaxMessageSize` defaulting), and one startup validation edge case.

nitronode/runtime.go
```diff
@@ -195,11 +203,20 @@ func InitBackbone() *Backbone {
...
 		ObserveConnections:      runtimeMetrics.SetRPCConnections,
 		WsConnProcessBufferSize: conf.WsProcessBufferSize,
 		WsConnWriteBufferSize:   conf.WsWriteBufferSize,
+		WsConnMaxMessageSize:    conf.WsMaxMessageSize,
+		NewFrameRateLimiter: func() rpc.FrameRateLimiter {
+			if bytesPerSec <= 0 {
+				return rpc.NoopFrameRateLimiter{}
+			}
+			return rpc.NewByteTokenBucket(bytesPerSec, bytesBurst)
+		},
```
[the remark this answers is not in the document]
philanton: Already addressed in 83c3b54a — the fail-fast logic now lives at `nitronode/runtime.go:212-227`. Rejects `bytesBurst <= 0` and `bytesBurst < WsMaxMessageSize` when `WsBytesPerSec > 0`. Subsequent commit 6a7556a3 also rejects `WsMaxMessageSize <= 0` upstream so the `> 0` guard on the second check could be dropped.

pkg/rpc/connection.go
```diff
@@ -322,10 +355,24 @@ func (conn *WebsocketConnection) readMessages(handleClosure func(error)) {
...
+				// Expected attacker / misconfigured-client path. Frame exceeded
+				// SetReadLimit; gorilla already sent close 1009. Treat as
+				// graceful close, not abnormal termination.
+				conn.logger.Warn("inbound frame exceeded MaxMessageSize; closing connection",
+					"error", err,
+					"origin", conn.origin,
+					"max_message_size", conn.maxMessageSize,
+				)
+				handleClosure(nil)
+			case websocket.IsUnexpectedCloseError(err,
```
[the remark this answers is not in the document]
philanton: Already addressed in 83c3b54a. The switch at `pkg/rpc/connection.go:360-361` now matches both `errors.Is(err, websocket.ErrReadLimit)` (local read-limit hit) and `websocket.IsCloseError(err, websocket.CloseMessageTooBig)` (peer-initiated 1009). `errors` added to imports. Test split into `TestWebsocketConnection_LocalReadLimit_GracefulClose` (injects `ErrReadLimit`) and `TestWebsocketConnection_PeerInitiated1009_GracefulClose` to cover both paths.

nitronode/runtime.go
philanton: Done in 6a7556a3. `runtime.go` now `logger.Fatal` on `conf.WsMaxMessageSize <= 0` before RPC node init. Matches C-01 intent — frame cap cannot be disabled via env, even though `NewWebsocketConnection` would still substitute its 128 KiB default.

nitronode/chart/config/stress-v1/nitronode.yaml.gotmpl
philanton: Fixed in f91c1ed0. Annotation removed from `stress-v1` gotmpl, replaced with an inline comment explaining the scope (HTTP request body only, no effect on WS frames after the upgrade). `chart/README.md` updated likewise — readers redirected to `NITRONODE_WS_MAX_MESSAGE_SIZE` as the sole effective frame cap.

pkg/rpc/connection.go
philanton: Fixed in 7c091057. Defaulting now uses `<= 0` to match every other field in the struct. Both struct-field comments corrected to: "Non-positive values fall back to the default; the cap cannot be disabled at this layer." The asymmetric `== 0` trap is gone.

nitronode/runtime.go
philanton: Fixed in 6a7556a3. Took the upstream-reject route rather than substituting the effective default in the comparison: `runtime.go` now `logger.Fatal` on `WsMaxMessageSize <= 0` before the burst check runs, so by the time the comparison happens `WsMaxMessageSize` is guaranteed positive. Dropped the `> 0` guard on the burst-vs-message-size check accordingly. Closes the silent-default trap.

pkg/rpc/connection_test.go
philanton: Fixed in 136233c4. Replaced with a direct `require.Equal(t, int64(64*1024), wsConnMock.getReadLimit())` — `Serve` calls `SetReadLimit` synchronously before spawning any goroutine (`connection.go:254`), so the limit is observable the moment `Serve` returns. Comment updated to make the synchronous contract explicit.

pkg/rpc/connection.go
philanton: Fixed in 7c091057. Added an "Allocation note" paragraph to the `FrameRateLimiter` interface comment: `Admit` runs after the frame is heap-allocated; per-frame size is bounded by `SetReadLimit`, but a burst of N back-to-back max-sized frames can briefly hold up to N * MaxMessageSize bytes per connection. Implementers are pointed at that ceiling when sizing burst capacity.

pkg/rpc/rate_limiter.go
philanton: Done in 512a513d. Renamed `FrameRateLimiter.Allow` → `Admit` on the interface, both implementations (`NoopFrameRateLimiter`, `ByteTokenBucket`), the call site at `connection.go:392`, and the test stubs. Agreed on the framing — admission control reads naturally as a query (`if !limiter.Admit(now, size)`) and avoids the mental inversion that `Allow` invites. `x/time/rate` precedent noted but not load-bearing for a new interface.

philanton:
@ihsraham — agreed, tracking separately. Out of C-01 scope, will cut a follow-up PR bumping `go.mod` toolchain (1.25.7 → 1.25.9) and the `nitronode/Dockerfile` base image (`golang:1.25-alpine` → `1.25.9-alpine`) with `govulncheck` verification before/after.

nksazonov (commented):
[TEST] Ignore this review — testing GitHub review API functionality.

ihsraham (approved):
approved. the c-01 boundary looks addressed: the frame cap is applied before json parsing, nitronode rejects a disabled frame cap, and the byte limiter config fails fast when burst is too small.

two small cleanup notes, not blockers:
- `pkg/rpc/node.go:137` still says `<0` disables the cap, but the connection layer now defaults non-positive values and nitronode rejects `<= 0`.
- `pkg/rpc/rate_limiter.go:32` still says the sole caller is `allow`; the method is now `admit`.

nksazonov (approved):

PR #724, opened 2026-05-06
MF-H01: fix(nitronode): paginate get_last_key_states endpoints

philanton:
## Summary
- Audit finding **MF-H01**: `channels.v1.get_last_key_states` and `app_sessions.v1.get_last_key_states` returned every session-key state for a wallet in a single response. A user who registers many keys for their own wallet could repeatedly trigger expensive DB reads and large payloads (DB / CPU / memory exhaustion).
- Add mandatory pagination with a hard maximum page size of **10** to both endpoints. Default limit also **10**; values above the cap are silently clamped.
- Store layer now returns `(states, totalCount, err)`, computed via a `COUNT()` subquery against the same join. Handlers populate the standard `PaginationMetadataV1` response field used elsewhere in the v1 API.
- `ORDER BY (created_at DESC, id ASC)` for stable pagination — `id` is the primary key, used as a deterministic tiebreaker so concurrent inserts in the same `created_at` tick do not cause overlapping or missed rows across pages.

## Changes
- `pkg/rpc/api.go`: add `Pagination` to `*GetLastKeyStatesRequest`, add `Metadata` to `*GetLastKeyStatesResponse` for both `channels.v1` and `app_sessions.v1`.
- `nitronode/store/database/{channel,app_session}_session_key_state.go`: new signature `(wallet, sessionKey, limit, offset) -> (states, totalCount, err)`. Adds COUNT subquery + `LIMIT`/`OFFSET` + stable order.
- `nitronode/store/database/interface.go`: signatures updated.
- `nitronode/api/{channel_v1,app_session_v1}/`: handler `Store` interfaces, handlers, and mocks updated. Handlers parse pagination params, clamp to default 10 / max 10, build metadata.
- `docs/api.yaml`: document new `pagination` request field and `metadata` response field for both methods.
- DB store tests: bulk-updated existing call sites; added a new pagination subtest (5 keys, page 2/2/1, no overlap, `totalCount == 5`).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — all packages pass, including new pagination tests in `nitronode/store/database`.

## Summary by CodeRabbit

* **New Features**
  * Added pagination to get_last_key_states for channels and apps: requests accept optional pagination params; responses include pagination metadata (page, per-page, total count, page count). Maximum page size is capped at 10.


nitronode/store/database/app_session_key_state.go
```diff
@@ -113,26 +114,32 @@ func (s *DBStore) GetLastAppSessionKeyStates(wallet string, sessionKey *string)
 		subQuery = subQuery.Where("session_key = ?", strings.ToLower(*sessionKey))
 	}
 
-	query := s.db.
-		Joins("JOIN (?) AS latest ON app_session_key_states_v1.user_address = latest.user_address AND app_session_key_states_v1.session_key = latest.session_key AND app_session_key_states_v1.version = latest.max_version", subQuery).
+	baseQuery := s.db.Model(&AppSessionKeyStateV1{}).
+		Joins("JOIN (?) AS latest ON app_session_key_states_v1.user_address = latest.user_address AND app_session_key_states_v1.session_key = latest.session_key AND app_session_key_states_v1.version = latest.max_version", subQuery)
+
+	var totalCount int64
```
ihsraham, line 120: same issue here. the handler caps how many rows come back, but this exact count plus a large offset can still make one request expensive when a wallet has many keys. to fully close h-01, i would make the db work bounded too, for example with limit + 1 and has_more, or cursor pagination.

nitronode/store/database/channel_session_key_state.go
```diff
@@ -95,25 +96,31 @@ func (s *DBStore) GetLastChannelSessionKeyStates(wallet string, sessionKey *stri
 		subQuery = subQuery.Where("session_key = ?", strings.ToLower(*sessionKey))
 	}
 
-	query := s.db.
-		Joins("JOIN (?) AS latest ON channel_session_key_states_v1.user_address = latest.user_address AND channel_session_key_states_v1.session_key = latest.session_key AND channel_session_key_states_v1.version = latest.max_version", subQuery).
+	baseQuery := s.db.Model(&ChannelSessionKeyStateV1{}).
+		Joins("JOIN (?) AS latest ON channel_session_key_states_v1.user_address = latest.user_address AND channel_session_key_states_v1.session_key = latest.session_key AND channel_session_key_states_v1.version = latest.max_version", subQuery)
+
+	var totalCount int64
```
ihsraham, line 102: i think this still leaves a smaller dos path open. the response is now capped, which is good, but an exact count plus a large offset can still make the db walk a lot of rows before it returns one page. since h-01 is about per-request cost, i would avoid exact counts here and use limit + 1 with has_more, or cursor pagination.

docs/api.yaml
```diff
@@ -648,12 +648,19 @@ api:
...
                   optional: true
+                - field_name: pagination
+                  type: pagination_params
+                  description: Pagination parameters (offset, limit, sort). Default limit 10, max 10.
+                  optional: true
               response:
                 - field_name: states
                   type: array
                   items:
                     type: app_session_key_state
```
ihsraham: this looks like the channel endpoint, but the response item type is still app_session_key_state. that can make clients built from this spec expect the wrong shape. i think this should be channel_session_key_state.

docs/api.yaml
```diff
@@ -827,7 +834,7 @@ api:
                 - message: invalid_session_key_state
                   description: The session key state is invalid
             - name: get_last_key_states
-              description: Retrieve the latest session key states for a user with optional filtering by session key
+              description: Retrieve the latest session key states for a user with optional filtering by session key. Mandatory pagination caps response size (max page size 10).
               request:
                 - field_name: wallet
```
ihsraham: this field name does not match the wire type in go and ts. both use user_address here, not wallet, so the spec can lead clients to send the wrong key.

ihsraham:
one api contract gap is still outside this diff. the low-level rpc types now accept pagination, but the high-level ts and go sdk helpers still do not expose pagination and still return only states, not metadata.

this means existing sdk users will now silently get only the first 10 keys and have no way to see that more pages exist. i would add paginated helper methods, or extend the options and return metadata.

nksazonov (commented):
test review body only - no inline comments

nksazonov:
**⚠️ WARNING: `page=0` / `page_count=0` returned for empty result sets**

https://github.com/layer-3/nitrolite/blob/2cf3509fbc997209a6056aebf11e867ee6b6550f/nitronode/api/channel_v1/get_last_key_states.go#L73-L80

`page` and `pageCount` are left at their zero values when `totalCount == 0`, so the metadata serialises as `{"page": 0, "per_page": 10, "total_count": 0, "page_count": 0}`. Non-empty results use 1-based pages, so empty results produce an inconsistent `page: 0`. Clients checking `page >= 1` or rendering "Page 0 of 0" need a special case the type doesn't hint at.

Set `page = 1` unconditionally; only compute `pageCount` inside the `totalCount > 0` guard.

Same issue at `nitronode/api/app_session_v1/get_last_key_states.go` L73–L80 (identical block).

nksazonov:
**⚠️ WARNING: SDK client silently caps at 10 results; callers cannot paginate or detect truncation**

https://github.com/layer-3/nitrolite/blob/2cf3509fbc997209a6056aebf11e867ee6b6550f/sdk/ts/src/client.ts#L1707-L1713

`resp.metadata` (containing `total_count` and `page_count`) is discarded here. The request is also built without a `pagination` field. Callers who previously received all session key states will now silently receive at most 10 with no indication that more exist — the return type `Promise<ChannelSessionKeyStateV1[]>` gives no hint of truncation.

At minimum, return the metadata alongside the states (e.g. `{ states, metadata }`) so consumers can detect truncation. The same issue applies to `getLastKeyStates` (app session variant) at L1762–L1768.

nksazonov:
**⚠️ WARNING: `Sort` field accepted but never read**

https://github.com/layer-3/nitrolite/blob/2cf3509fbc997209a6056aebf11e867ee6b6550f/nitronode/api/channel_v1/get_last_key_states.go#L31-L38

Only `Limit` and `Offset` are extracted from `req.Pagination`; `Sort` is silently ignored and the order is hardcoded as `created_at DESC, id ASC`. The API YAML describes this field as `"offset, limit, sort"` and `PaginationParamsV1.Sort` is a defined `*string`, so a client passing `sort: "asc"` gets silently ignored behaviour with no error returned.

Either drop `Sort` from `PaginationParamsV1` for these endpoints and document the fixed order, or return an error when an unsupported sort value is supplied.

Same in `nitronode/api/app_session_v1/get_last_key_states.go` L31–L38.

nksazonov:
**💡 SUGGESTION: Page formula silently misbehaves for non-aligned offsets**

https://github.com/layer-3/nitrolite/blob/2cf3509fbc997209a6056aebf11e867ee6b6550f/nitronode/api/channel_v1/get_last_key_states.go#L77-L79

`page = (offset / limit) + 1` only gives a meaningful result when `offset` is a multiple of `limit`. With `limit` clamped to 10, `offset=3` produces `page=1` — identical to `offset=0`. Since no validation enforces aligned offsets, a client using a non-standard offset receives misleading `page` metadata. Add a comment stating the aligned-offset assumption, or validate and reject non-aligned values.

nksazonov:
**💡 SUGGESTION: Identical constants duplicated across two packages**

https://github.com/layer-3/nitrolite/blob/2cf3509fbc997209a6056aebf11e867ee6b6550f/nitronode/api/channel_v1/get_last_key_states.go#L9-L11

`getLastKeyStatesDefaultLimit` and `getLastKeyStatesMaxLimit` (both `10`) are defined identically here and in `nitronode/api/app_session_v1/get_last_key_states.go`. If the cap ever changes it needs updating in both places with no compile-time check that they stay in sync. Consider moving them to a shared location (e.g. alongside `PaginationParamsV1` in `pkg/rpc`) and referencing from both packages.

nitronode/store/database/app_session_key_state.go
philanton: Yeah, fair point. Went with a pointer table rather than `limit + 1` / cursor: added `current_session_key_states_v1` keyed on `(user_address, session_key, kind)` with a single `version` column. `get_last_key_states` now filters that pointer table by `(user, kind)` and JOINs history on `(user, session_key, version)`, so both the COUNT and the page query are bounded by the number of distinct keys for the user, not by history churn. Also added `NITRONODE_MAX_SESSION_KEYS_PER_USER` (default 100) to cap that set when registering a new key. Fixed in [2be4a4b](https://github.com/layer-3/nitrolite/pull/724/changes/2be4a4b077d01c2f97a8f7d46ceeb9e5dea46cda).

nitronode/store/database/channel_session_key_state.go
philanton: Same approach for both endpoints — see the reply on the app_session thread for the rationale. Both readers are now bounded by the pointer table. Fixed in [2be4a4b](https://github.com/layer-3/nitrolite/pull/724/changes/2be4a4b077d01c2f97a8f7d46ceeb9e5dea46cda).

docs/api.yaml
philanton: Yep, that one is wrong — channel response items should be `channel_session_key_state`. Fixed in [d4e02d5](https://github.com/layer-3/nitrolite/pull/724/changes/d4e02d511429bc7a35dc306394e1c18c00c1054e).

docs/api.yaml
philanton: Yep, the wire types use `user_address` in both Go and TS — the spec was the odd one out. Fixed in [d4e02d5](https://github.com/layer-3/nitrolite/pull/724/changes/d4e02d511429bc7a35dc306394e1c18c00c1054e).

nitronode/store/database/app_session_key_state.go
```diff
@@ -113,26 +114,35 @@ func (s *DBStore) GetLastAppSessionKeyStates(wallet string, sessionKey *string)
...
 
 	var dbStates []AppSessionKeyStateV1
 	if err := query.Find(&dbStates).Error; err != nil {
 		if err == gorm.ErrRecordNotFound {
-			return []app.AppSessionKeyStateV1{}, nil
+			return []app.AppSessionKeyStateV1{}, uint32(totalCount), nil
 		}
-		return nil, fmt.Errorf("failed to get session key states: %w", err)
+		return nil, 0, fmt.Errorf("failed to get session key states: %w", err)
 	}
```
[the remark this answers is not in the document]
philanton: Already done in [6cc713a](https://github.com/layer-3/nitrolite/pull/724/changes/6cc713ae840fcad96836301fc32287a0d24bd9b5) — dropped the unreachable `ErrRecordNotFound` branch from both the channel and app_session `Find` calls.

philanton:
Replies inline for the issue-level threads:

@ihsraham — re. SDK helpers ([comment](https://github.com/layer-3/nitrolite/pull/724#issuecomment-4388846291)): you're right, the high-level wrappers in `sdk/ts/src/client.ts` and the Go SDK still don't forward `pagination` or surface `metadata`. Existing 0.5.x callers will silently truncate at 10 keys with no signal that more exist. I'd track that as a follow-up — it touches both SDKs and changes the helper return shape, so it deserves its own PR rather than piling onto this one.

@nksazonov:
1) `page=0` / `page_count=0` on empty results — fixed in [d4e02d5](https://github.com/layer-3/nitrolite/pull/724/changes/d4e02d511429bc7a35dc306394e1c18c00c1054e). Pulled the metadata builder out into `buildPageMetadata` and made `page` always 1-based; `pageCount` stays at 0 only when `totalCount == 0`.
2) SDK client.ts discarding metadata — same SDK gap as above, follow-up.
3) `Sort` silently ignored — fixed in [d4e02d5](https://github.com/layer-3/nitrolite/pull/724/changes/d4e02d511429bc7a35dc306394e1c18c00c1054e). Non-empty `sort` is now rejected with `invalid_pagination`. Kept `Sort` on `PaginationParamsV1` since it's shared across the v1 API and other endpoints honor it; validation lives in the handler.
4) Page formula on non-aligned offsets — added a comment on `buildPageMetadata` documenting that for unaligned offsets the metadata reports the page containing row `offset+1`. I didn't reject outright since callers can legitimately resume from an item count rather than a page number.
5) Duplicated constants — fixed in [d4e02d5](https://github.com/layer-3/nitrolite/pull/724/changes/d4e02d511429bc7a35dc306394e1c18c00c1054e). Moved to `pkg/rpc.GetLastKeyStatesPageLimit` so both handler packages reference one source of truth.

Separately, addressed the COUNT/offset DOS angle that @ihsraham raised on the line-comment threads: added `current_session_key_states_v1`, a per-`(user_address, session_key, kind)` pointer table. `get_last_key_states` now filters that pointer table by `(user, kind)` and JOINs history on the version, so per-request DB work is bounded by the number of distinct keys for the user, regardless of version churn. Also added `NITRONODE_MAX_SESSION_KEYS_PER_USER` (default 100) to bound that set on registration. See [2be4a4b](https://github.com/layer-3/nitrolite/pull/724/changes/2be4a4b077d01c2f97a8f7d46ceeb9e5dea46cda).

nitronode/api/channel_v1/submit_session_key_state.go
```diff
@@ -69,10 +70,24 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
...
+		// proper "expected version" error rather than racing on the history UNIQUE constraint.
+		latestVersion, err := tx.LockSessionKeyState(coreState.UserAddress, coreState.SessionKey, database.SessionKeyKindChannel)
 		if err != nil {
-			return rpc.Errorf("failed to check existing session key state: %v", err)
+			return rpc.Errorf("failed to lock session key state: %v", err)
+		}
+
+		// Enforce the per-user cap when registering a new session key. Existing keys (latestVersion > 0)
+		// can always be updated regardless of the cap so that legitimate rotation is never blocked.
+		if latestVersion == 0 && h.maxSessionKeysPerUser > 0 {
```
ihsraham, line 91: i think this cap can still race under parallel new key submits. this transaction locks only the specific `(user, session_key, kind)` row, so two different new keys for the same user do not wait on each other. they can both read the same count and both pass before either commit lands. since this cap is part of the h-01 fix, i would lock per `user_address` before the count and insert, for example with a user lock row or a postgres advisory transaction lock. same issue applies in the app session submit path.

philanton: Yep, you're right that the row lock leaves this open — two concurrent new-key submits for the same user both pass the count. I'd treat the cap as a soft DOS bound rather than a hard quota though: the over-shoot is at most `(concurrent new-key writers - 1)` keys per cap-check window, so an attacker can't drive the count arbitrarily high without also driving concurrency arbitrarily high, and the bounded read path on the get/store side is unchanged.

Kept the simpler shape for now and captured the race + the concrete fix (per-user `pg_advisory_xact_lock(hashtext(user_address))`) as a TODO at the call site in [cfdbbbc](https://github.com/layer-3/nitrolite/pull/724/changes/cfdbbbc830ed8d28aebd17b1140cb47ed79c2b5e), so when/if we ever need a hard quota the fix is one lock acquisition away.

ihsraham (approved):
approving for h-01 closure. the finding asked for hard max pagination on both `get_last_key_states` endpoints, and this now does that with a max page size of 10.

`current_session_key_states_v1` also bounds the db read path, so the original unbounded response and history-scan dos path is addressed. the soft per-user cap race can stay as a follow-up since it does not reopen this finding.

PR #726, opened 2026-05-07
MF-L01: fix(contracts/ChannelHub): cap ERC20 transfer returndata copy to 32 bytes

nksazonov:
## Security audit finding

**Severity:** Low  
**Location:** `ChannelHub._trySafeTransfer`

The original implementation used the high-level Solidity call form:

```solidity
(bool success, bytes memory returnData) =
    address(token).call{gas: TRANSFER_GAS_LIMIT}(abi.encodeCall(IERC20.transfer, (to, amount)));
```

`TRANSFER_GAS_LIMIT` correctly caps the gas forwarded to the callee, but the returndata copy happens in the **caller's** context after the call returns. Solidity compiles the `bytes memory returnData` capture into three steps — `RETURNDATASIZE`, memory allocation of `returndata_size + 32` bytes, and `RETURNDATACOPY` over the full buffer — all paid from the caller's (ChannelHub's) remaining gas, not from the capped budget.

A malicious or non-compliant token can return an arbitrarily large buffer from `transfer()`. This triggers quadratic memory expansion in ChannelHub before `returnData.length` is ever checked, causing `_trySafeTransfer` to OOG and revert rather than return `false`. The revert propagates through `_nonRevertingPushFunds`, breaking its non-reverting guarantee and blocking channel settlement paths (close, challenge response, escrow finalization).

## Fix

Replaced the high-level call with an assembly block that writes at most 32 bytes of output to Solidity scratch space (`0x00–0x1f`):

```solidity
bytes memory callData = abi.encodeCall(IERC20.transfer, (to, amount));

assembly ("memory-safe") {
    // Output to scratch space (0x00-0x1f); returndatasize() still reflects the full returndata length.
    success := call(TRANSFER_GAS_LIMIT, token, 0, add(callData, 0x20), mload(callData), 0x00, 0x20)
    rdsize := returndatasize()
    retval := mload(0x00)
}
```

Key properties:
- `retLen = 0x20` in the `call` opcode limits the output buffer to 32 bytes; EVM memory expansion is bounded by `retOffset + retLen = 0x20`, not by actual returndata size.
- `returndatasize()` still reflects the full length of the token's response, used to distinguish no-return-value tokens (`rdsize == 0`) from truncated responses (`rdsize < 32`).
- The scratch space range `0x00–0x1f` is already within the memory expanded by the preceding `abi.encodeCall()`, so no additional expansion occurs.
- `abi.encodeCall` is used for calldata encoding (type-checked selector, no hardcoded magic constant).

## Tests added

Two new tests cover the previously untested branches, plus two regression pins:

| Test | Branch |
|---|---|
| `test_succeeds_whenERC20ReturnsOversizedDataWithNonzeroValue` | `rdsize > 32`, nonzero first word → success |
| `test_accumulatesReclaims_whenERC20ReturnsOversizedDataWithZeroValue` | `rdsize > 32`, zero first word → reclaim |
| `test_accumulatesReclaims_whenERC20ReturnsShortData` | `rdsize ∈ [1, 31]` → failure regardless of content |
| `test_succeeds_whenERC20ReturnsNonCanonicalBoolValue` | `rdsize == 32`, value `2` → treated as success (not ABI decode revert) |

Also fixed a pre-existing bug in `MalformedReturningERC20`: `mstore(0, 1); return(0, 1)` was returning byte `0x00` (the MSB of the big-endian word) instead of `0x01`. Replaced with `mstore8(0, 1)`.

### DIFF
--- a/contracts/test/ChannelHub_nonRevertingPushFunds.t.sol
+++ b/contracts/test/ChannelHub_nonRevertingPushFunds.t.sol
@@ -204,6 +204,33 @@ contract ChannelHubTest_nonRevertingPushFunds is Test {
         _verifyBalancesNotChanged(recipient, address(oversizedToken), TRANSFER_AMOUNT);
     }
 
+    // Covers the rdsize ∈ [1, 31] branch of _trySafeTransfer.
+    // Any response shorter than 32 bytes is rejected as failure regardless of content —
+    // a valid ERC20 bool is always a full 32-byte ABI word.
+    function test_accumulatesReclaims_whenERC20ReturnsShortData() public {
+        OversizedReturnERC20 shortToken = new OversizedReturnERC20(16, 0);
+        shortToken.mint(address(cHub), BALANCE_AMOUNT);
+
+        vm.expectEmit(true, true, false, true);
+        emit ChannelHub.TransferFailed(recipient, address(shortToken), TRANSFER_AMOUNT);
+
+        cHub.exposed_nonRevertingPushFunds(recipient, address(shortToken), TRANSFER_AMOUNT);
+
+        _verifyBalancesNotChanged(recipient, address(shortToken), TRANSFER_AMOUNT);
+    }
+
+    // Covers rdsize == 32 with a non-canonical bool value (2).
+    // abi.decode(..., (bool)) reverts on value 2 — Solidity 0.8 only accepts 0 or 1.
+    // Decoding as uint256 and checking != 0 handles this correctly without reverting.
+    function test_succeeds_whenERC20ReturnsNonCanonicalBoolValue() public {
+        OversizedReturnERC20 nonCanonicalToken = new OversizedReturnERC20(32, 2);
+        nonCanonicalToken.mint(address(cHub), BALANCE_AMOUNT);
+
+        cHub.exposed_nonRevertingPushFunds(recipient, address(nonCanonicalToken), TRANSFER_AMOUNT);
+
+        _verifyTransferSuccess(recipient, address(nonCanonicalToken), TRANSFER_AMOUNT);
+    }
+
     // ========== ERC777 Donation-Back Tests ==========
 
     function test_succeeds_whenERC777DonatesBack() public {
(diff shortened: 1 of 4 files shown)


philanton (approved):
Solid fix. Assembly correctly caps returndata copy to 32 bytes via `retLen=0x20`; `memory-safe` annotation valid since scratch space (0x00-0x3f) writes are permitted. Memory expansion bounded by abi.encodeCall's prior allocation. Branch coverage good: rdsize ∈ {0, [1,31], 32, >32} all exercised, plus non-canonical bool regression pin.

Minor notes (non-blocking):
- The `MalformedReturningERC20` `mstore→mstore8` change is an unrelated mock bugfix; behavior-neutral for both old and new `_trySafeTransfer` (both fall into the `length < 32` failure path), but worth flagging since it's bundled in the same commit.
- `OversizedReturnERC20` skips `_transfer` when `firstWord == 0`, so the "token returns oversized-failure-but-actually-moved-funds" combo isn't covered. Edge case, likely out of scope.
- A fuzz test across `rdsize` boundaries (31/32/33) would harden the regression pin, but not required.

contracts/test/ChannelHub_nonRevertingPushFunds.t.sol
```diff
@@ -173,6 +174,63 @@ contract ChannelHubTest_nonRevertingPushFunds is Test {
         _verifyBalancesNotChanged(recipient, address(malformedToken), TRANSFER_AMOUNT);
     }
 
+    // ========== Oversized Return Data ERC20 Tests ==========
+
+    // Covers the rdsize > 32, retval != 0 branch of _trySafeTransfer.
+    // With the old high-level (bool, bytes memory) call form, a token returning a large buffer
+    // would exhaust caller gas during returndata copy before we could inspect the length.
+    // The assembly fix caps the copy to 32 bytes, so the first word is read and the transfer succeeds.
+    function test_succeeds_whenERC20ReturnsOversizedDataWithNonzeroValue() public {
```
ihsraham, line 210: this test is useful, but it still passes if the old high-level copy is restored. since this is audit remediation, can we add one case that fails on the old code and passes on this fix, maybe by using very large return data with a tight gas limit? that makes l-01 pinned for real.

ihsraham (approved):
approved

contracts/test/ChannelHub_nonRevertingPushFunds.t.sol
nksazonov: 6bfedda1bb96481bba7db74ef89d08a4649436dd

PR #728, opened 2026-05-07
MF-I01-I02: fix(contracts): address security audit findings I-01 and I-02

nksazonov:
### I-01 — Wrong error selector in `withdrawFromNode`

`withdrawFromNode()` guarded `msg.sender == NODE` with `IncorrectNode()`. The contract's convention is:

- `IncorrectNode()` — data validation: `def.node != NODE` in `_requireValidDefinition`
- `IncorrectMsgSender()` — caller auth: `msg.sender != NODE`

Using the wrong selector made the revert misleading to integrators doing error-based retry logic or monitoring, and was inconsistent with all other `msg.sender` guards in the contract (`challengeChannel`, `initiateEscrowDeposit`).

**Fix:** `IncorrectNode()` → `IncorrectMsgSender()` in `withdrawFromNode()`.
**Test added:** `test_reverts_whenCallerIsNotNode` asserts the correct selector.

---

### I-02 — Misleading NatSpec on `registerNodeValidator`

The `@dev` comment stated *"NODE's private key only signs the registration data, never sends transactions directly"* — a property true only for `registerNodeValidator` (relayer-friendly via off-chain signature), but not for the protocol as a whole. Three other functions require `NODE` as `msg.sender`:

- `withdrawFromNode`
- home-chain path of `initiateEscrowDeposit`
- new-version path of `challengeChannel` with `INITIATE_ESCROW_DEPOSIT`

Reading the comment as a protocol-wide guarantee produced a wrong trust model for integrators building relayer infrastructure.

**Fix:** Narrowed the NatSpec to scope the relayer-friendly claim to `registerNodeValidator` only; listed the three NODE-gated operations explicitly; clarified that the `challengeChannel` guard applies only to the new-version (state-advancing) path, not the same-version (already-processed) path.

---

## Additional changes

- **`perf`** `Utils.sol`: replaced `keccak256(abi.encode(...))` with inline assembly in `getChannelId` and `getEscrowId`, eliminating heap allocations. ~350–580 gas saved avg on `createChannel`, `initiateEscrow*`, `initiateMigration`; −170 bytes deployed bytecode. Added `INVARIANT` warning at `ChannelDefinition` in `Types.sol`: the hardcoded struct size `0xC0` must be updated if fields are added or removed.
- **`fix`** Suppressed three `forge-lint incorrect-shift` false positives in test files (bitmask construction `1 << N` is intentional; linter cannot infer semantic intent).
- **`fix`** Resolved `/review` findings: struct-size invariant comment, NatSpec same-version-path clarification, `test_reverts_whenCallerIsNotNode` test.

## Test plan

- [x] `forge test` — 269/269 pass
- [x] `forge lint` — zero warnings, zero notes
- [x] `withdrawFromNode` non-NODE caller reverts with `IncorrectMsgSender` (new test)

### DIFF
--- a/contracts/src/Utils.sol
+++ b/contracts/src/Utils.sol
@@ -33,9 +33,12 @@ library Utils {
     function getEscrowId(bytes32 channelId, uint64 version) internal pure returns (bytes32 escrowId) {
         // "channelId, (state-)version" pair is unique as long as participants do not reuse versions
         // Uses the 64-byte scratch space (0x00–0x3f) to avoid a heap allocation.
+        // Mask version to 64 bits: mstore writes the full 256-bit stack word, and Solidity does
+        // not guarantee that internal call arguments are cleaned for sub-256-bit types. Without
+        // the mask, dirty high bits would produce a hash that diverges from abi.encode(channelId, version).
         assembly ("memory-safe") {
             mstore(0x00, channelId)
-            mstore(0x20, version)
+            mstore(0x20, and(version, 0xffffffffffffffff))
             escrowId := keccak256(0x00, 0x40)
         }
     }
--- a/contracts/src/Utils.sol
+++ b/contracts/src/Utils.sol
@@ -17,11 +17,12 @@ library Utils {
     error FailedToFetchDecimals();
 
     function getChannelId(ChannelDefinition memory def, uint8 version) internal pure returns (bytes32 channelId) {
-        bytes32 baseId = keccak256(abi.encode(def));
-
         assembly ("memory-safe") {
+            // ChannelDefinition has 6 static fields × 32 bytes = 192 (0xC0) bytes in memory.
+            // Memory layout is identical to abi.encode for structs with only value types, so we
+            // hash the struct pointer directly, avoiding the abi.encode allocation.
+            let baseId := keccak256(def, 0xC0)
             // Store the version in the first byte (most significant byte) of the channelId
-            // Clear the first byte of baseId, then set it to version
             channelId := or(
                 and(baseId, 0x00ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff),
                 shl(248, version)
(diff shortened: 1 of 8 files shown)


philanton (approved):
I-01 fix correct (selector matches contract's data-validation vs caller-auth convention), I-02 NatSpec accurate. Test pin for non-NODE caller good.

Worth raising on the bundled perf change (non-blocking):

- **`getEscrowId` dirty-bits**: `mstore(0x20, version)` reads `uint64 version` directly in assembly. Per Solidity docs, local variables accessed via inline assembly are not guaranteed to be cleaned — the old `abi.encode(channelId, version)` cleaned implicitly. If a caller ever passes a `uint64` with dirty high bits (e.g. from a narrowed `uint256` cast or an arithmetic temporary), the resulting hash differs from the old behavior. Recommend explicit mask: `mstore(0x20, and(version, 0xffffffffffffffff))`. Not exploitable with current call sites, but compiler-fragile.

- **`getChannelId` raw-memory hash**: relies on Solidity memory layout matching `abi.encode` output for a struct of value types — true today, and the `INVARIANT` comment at `ChannelDefinition` covers field count. Less risky than `getEscrowId` since fields are written through Solidity (always clean).

- forge-lint suppressions are reasonable; `1 << ID` is an intentional bitmask, linter can't infer semantic intent.

No migration concern since this lands pre-deployment.

nksazonov:
@philanton , addressed the dirty-bits possible bug surface: 740ea005

contracts/src/Utils.sol
```diff
@@ -17,21 +17,30 @@ library Utils {
     error FailedToFetchDecimals();
 
     function getChannelId(ChannelDefinition memory def, uint8 version) internal pure returns (bytes32 channelId) {
-        bytes32 baseId = keccak256(abi.encode(def));
-
         assembly ("memory-safe") {
+            // ChannelDefinition has 6 static fields × 32 bytes = 192 (0xC0) bytes in memory.
+            // Memory layout is identical to abi.encode for structs with only value types, so we
+            // hash the struct pointer directly, avoiding the abi.encode allocation.
+            let baseId := keccak256(def, 0xC0)
```
ihsraham, line 24: this looks ok, but this pr is for audit cleanup and this change also changes how channel ids are made. please add one small test that checks this returns the same id as the old abi.encode path. that makes the audit proof easier to trust.

ihsraham (approved):
approved. the two audit fixes look correct. i left one note asking for a small proof test because the channel id hash change is outside the audit items.

contracts/src/Utils.sol
nksazonov: 8cf5f5bc484ff4d2a59c3ced4f2a0b644a9fe32d

PR #730, opened 2026-05-08
MF-I06: fix(nitronode): gate escrow transitions on home channel onchain materialization

philanton:
## Summary

- Rename `CheckOpenChannel` → `CheckActiveChannel`. Return `(string, *core.ChannelStatus, error)` instead of `(string, bool, error)` so callers can distinguish `Void` (DB-only, awaiting onchain confirmation) from `Open` (materialized onchain). Nil status pointer means no active channel.
- Document why the permissive `Void+Open` filter is intentional: non-escrow offchain transitions (transfers, etc.) are allowed before onchain confirmation lands.
- Gate cross-chain escrow transitions (`MutualLock`, `EscrowLock`) on `Status == Open` with an executable guard placed **before** the stub `return rpc.Errorf("transition is not supported yet")`. Single-step unstub later: drop the stub return; guard stays active. Today the guard already runs — `Void` inputs get `"home channel is not materialized onchain"`, `Open` inputs fall through to `"not supported yet"`.
- Add `TestSubmitState_MutualLock_VoidHomeChannel_Rejected` and `TestSubmitState_EscrowLock_VoidHomeChannel_Rejected` to lock the contract: future PRs that unstub these transitions without preserving the guard get red CI.

## Why

Audit finding (MF-I06): `request_creation()` inserts the home channel into the DB as `ChannelStatusVoid` before the channel is materialized onchain. `CheckOpenChannel` treated `status <= Open` as usable, so a future cross-chain `MutualLock` / `INITIATE_ESCROW_DEPOSIT` flow could be co-signed before the prior creation/checkpoint state was submitted via `createChannel()`. Onchain, `_isChannelHomeChain()` returns false while status == VOID, so `initiateEscrowDeposit()` would not take the home-chain path. The fix is to surface the actual status to callers and gate cross-chain escrow transitions on `Status == Open`.

The escrow branches are currently stubbed, so the guard has no live-traffic impact today (the stub `return` would fire regardless, and `nodeSig` is never returned via `c.Succeed` on the error path). The value of shipping the guard now is the CI contract: the Void-rejection tests pin the behavior so an unstub-PR that forgets the guard cannot land green.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./nitronode/store/database/... ./nitronode/api/channel_v1/... ./nitronode/api/app_session_v1/...`
- [x] New Void-rejection tests pass; existing `_Success` tests (skipped) unaffected.

## Summary by CodeRabbit

* **Refactor**
  * Improved channel status validation to distinguish between in-progress and on-chain channels across the application. The system now provides more granular channel state information to ensure accurate eligibility checks during deposit submissions and state transitions.


nitronode/api/channel_v1/submit_state.go
```diff
@@ -49,12 +49,12 @@ func (h *Handler) SubmitState(c *rpc.Context) {
...
 
-		approvedSigValidators, userHasOpenChannel, err := tx.CheckOpenChannel(incomingState.UserWallet, incomingState.Asset)
+		approvedSigValidators, channelStatus, err := tx.CheckActiveChannel(incomingState.UserWallet, incomingState.Asset)
 		if err != nil {
-			return rpc.Errorf("failed to check open channel: %v", err)
+			return rpc.Errorf("failed to check active channel: %v", err)
 		}
-		if !userHasOpenChannel {
-			return rpc.Errorf("user has no open channel")
+		if channelStatus == nil {
```
ihsraham, line 56: for i-06 closure, can we add the escrow materialization check next to this one, before the shared signing path?

the issue is about not co-signing `MutualLock` or `EscrowLock` while the home channel is still only db-side `Void`. returning unsupported later keeps the response and storage safe today, but a small pre-sign `channelStatus == core.ChannelStatusOpen` gate would make this boundary clear.

nitronode/api/channel_v1/submit_state.go
```diff
@@ -168,6 +168,15 @@ func (h *Handler) SubmitState(c *rpc.Context) {
...
+				// while status == VOID, so initiateEscrowDeposit() on the home chain
+				// would not take the home-chain path until the prior creation/checkpoint
+				// state is submitted via createChannel().
+				//
+				// if *channelStatus != core.ChannelStatusOpen {
+				// 	return rpc.Errorf("home channel is not materialized onchain")
+				// }
 				// if err := h.createEscrowChannel(tx, incomingState); err != nil {
 				// 	return err
 				// }
```
nksazonov, line 189: The guard is unreachable: it sits after the unconditional `return` on line 170. When this transition is unstubbed, a developer must simultaneously remove the `return` *and* uncomment the guard — missing either step silently ships an unprotected path. The comment alone is not a reliable safety net.

**Suggestion:** Invert the structure — place the guard check (or a `// TODO: assert *channelStatus == ChannelStatusOpen here`) *before* the `return rpc.Errorf("transition is not supported yet")`, so unstubbing becomes a single-step operation and the guard is in execution order.

nitronode/api/channel_v1/submit_state.go
```diff
@@ -178,6 +187,14 @@ func (h *Handler) SubmitState(c *rpc.Context) {
...
+				// co-signing an EscrowLock. Same reason as MutualLock — the onchain
+				// home-chain path depends on the home channel having been created via
+				// createChannel() first.
+				//
+				// if *channelStatus != core.ChannelStatusOpen {
+				// 	return rpc.Errorf("home channel is not materialized onchain")
+				// }
 				// if err := h.createEscrowChannel(tx, incomingState); err != nil {
 				// 	return err
 				// }
```
nksazonov, line 205: Same fragility as the `MutualLock` branch above: the onchain-materialization guard is dead code after an unconditional `return`, so the two-step unstub risk applies identically here.

**Suggestion:** Move the guard check before the stub `return` so that removing the stub unconditionally activates the guard.

nitronode/api/channel_v1/submit_state_test.go
```diff
@@ -1179,7 +1179,7 @@ func TestSubmitState_MutualLock_Success(t *testing.T) {
 	mockAssetStore.On("GetAssetDecimals", asset).Return(uint8(6), nil)
 	mockAssetStore.On("GetTokenDecimals", uint64(1), "0xTokenAddress").Return(uint8(6), nil)
 	mockAssetStore.On("GetTokenDecimals", uint64(2), "0xTokenAddress").Return(uint8(6), nil)
-	mockTxStore.On("CheckOpenChannel", userWallet, asset).Return("0x03", true, nil)
+	mockTxStore.On("CheckActiveChannel", userWallet, asset).Return("0x03", core.ChannelStatusOpen, nil)
 	mockTxStore.On("GetLastUserState", userWallet, asset, false).Return(currentState, nil)
 	mockStatePacker.On("PackState", mock.Anything).Return(packedState, nil)
 	mockTxStore.On("GetChannelByID", homeChannelID).Return(&homeChannel, nil)
```
nksazonov, line 1185: Every test in this file mocks `CheckActiveChannel` returning `core.ChannelStatusOpen`. There is no test asserting that a `ChannelStatusVoid` home channel is *rejected* for `MutualLock` or `EscrowLock` transitions — the core of the audit finding. Without it, the guard could be accidentally omitted or inverted when unstubbing and CI would not catch it.

**Suggestion:** Add a `TestSubmitState_MutualLock_VoidChannel_Rejected` (and equivalent for `EscrowLock`) that mocks `CheckActiveChannel` returning `core.ChannelStatusVoid` and asserts an error response. This can be a failing/skipped placeholder now, structured to pass once the stub `return` is removed and the guard is active.

nksazonov (commented):
The PR correctly identifies the audit finding and surfaces the actual channel status to callers — the infrastructure for the fix is sound. Two concerns worth addressing before the escrow transitions are unstubbed:

1. The `ChannelStatusOpen` guards are placed after unconditional `return` statements, making them dead code with a two-step activation risk.
2. No test coverage exists for the `ChannelStatusVoid` rejection case, which is the exact scenario the finding describes.

Everything else (rename, signature change, documentation, app_session path) looks correct.

nitronode/api/channel_v1/submit_state.go
philanton: Yep, added the gate. Placed it per-branch at the `MutualLock`/`EscrowLock` cases rather than upstream right after `CheckActiveChannel` — an upstream gate would also block the default case (transfers etc.) on `Status == Open`, which would break the intentional `Void+Open` permissive filter for non-escrow offchain transitions. Per-branch keeps the boundary at the escrow flows where it actually matters.

Fixed in [e017daa](https://github.com/layer-3/nitrolite/pull/730/changes/e017daa2f1a4f03ef7b23ea573705b9c895e4ff5).

nitronode/api/channel_v1/submit_state.go
philanton: Yep, you're right — moved the guard before the stub `return` so unstubbing is single-step. Also added `TestSubmitState_MutualLock_VoidHomeChannel_Rejected` so this can't regress silently.

Fixed in [e017daa](https://github.com/layer-3/nitrolite/pull/730/changes/e017daa2f1a4f03ef7b23ea573705b9c895e4ff5).

nitronode/api/channel_v1/submit_state.go
philanton: Same fix applied here. Fixed in [e017daa](https://github.com/layer-3/nitrolite/pull/730/changes/e017daa2f1a4f03ef7b23ea573705b9c895e4ff5).

nitronode/api/channel_v1/submit_state_test.go
philanton: Added `TestSubmitState_MutualLock_VoidHomeChannel_Rejected` and `TestSubmitState_EscrowLock_VoidHomeChannel_Rejected`. Both run today (not skipped placeholders) — the guard is now executable, so they assert `"home channel is not materialized onchain"` for `ChannelStatusVoid` inputs. Future unstub-PR that drops the guard gets red CI.

Fixed in [e017daa](https://github.com/layer-3/nitrolite/pull/730/changes/e017daa2f1a4f03ef7b23ea573705b9c895e4ff5).

nitronode/api/channel_v1/submit_state.go
ihsraham: agreed that this should not be a global `Status == Open` gate, since transfers still need the void/open path. the missing piece is only the order: for `MutualLock` and `EscrowLock`, the check still runs after `h.nodeSigner.Sign(...)`.

the error path does not return the signature, so i do not see a practical leak today. for i-06 closure, though, the node should reject before signing. could we add a small pre-sign switch for those two transition types right after `CheckActiveChannel`?

nksazonov (approved):

nitronode/api/channel_v1/submit_state.go
philanton: Yep, fair — moved the gate ahead of `nodeSigner.Sign(...)` so MutualLock/EscrowLock get rejected pre-sign right after `CheckActiveChannel`. Kept the per-branch guards below the stub returns as belt-and-suspenders until the escrow transitions are unstubbed.

Fixed in [ea08a99](https://github.com/layer-3/nitrolite/pull/730/changes/ea08a99410dec8addb9a39d1027f06e0d5a92128).

PR #731, opened 2026-05-08
MF-M01: backfill state user_sig from on-chain events

philanton:
## Summary
- Plumb the on-chain user signature from the parsed `State` payload through the reactor into `channelEvent` / `channelChallengedEvent`, then forward it to the matching `EventHandlerService` handlers.
- New `ChannelHubEventHandlerStore.UpdateStateUserSigIfMissing` populates `channel_states.user_sig` when NULL; idempotent on replay via `user_sig IS NULL` guard.
- Closes the wedge where a unilateral on-chain checkpoint of a node-only state (e.g. recipient submits the `TransferReceive` state directly) leaves the local row partially signed, causing `EnsureNoOngoingStateTransitions` to permanently mismatch versions and block the channel until an on-chain challenge.

## Threat addressed
Recipient signs the node-issued `TransferReceive` state and submits it on chain. `HomeChannelCheckpointed` bumps `channels.state_version` to N+1, but `channel_states` row at N+1 still has `user_sig = NULL`. The gate query in `EnsureNoOngoingStateTransitions` filters on both signatures, picks the prior bilateral row at N, and reports a version mismatch on every subsequent state transition. Channel wedged offchain; recovery requires on-chain challenge — costly griefing vector with a one-tx attack surface.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] New unit tests:
  - `nitronode/event_handlers/service_test.go` — `BackfillsUserSig`, `BackfillError`; existing handler tests updated to expect the backfill mock call (empty sig).
  - `nitronode/store/database/db_store_test.go` — wedge-resolution end-to-end (gate fails → backfill → gate passes), idempotent replay (existing sig preserved), empty-sig no-op, unknown-version no-op.
  - `pkg/blockchain/evm/channel_hub_reactor_test.go` — reactor hex-encodes `Candidate.UserSig` into the event.

### DIFF
--- a/nitronode/event_handlers/service.go
+++ b/nitronode/event_handlers/service.go
@@ -64,6 +64,10 @@ func (s *EventHandlerService) HandleHomeChannelCreated(ctx context.Context, tx c
 		return err
 	}
 
+	if err := tx.UpdateStateUserSigIfMissing(event.ChannelID, event.StateVersion, event.UserSig); err != nil {
+		return err
+	}
+
 	logger.Info("handled HomeChannelCreated event", "channelId", event.ChannelID, "stateVersion", event.StateVersion, "userWallet", channel.UserWallet)
 	return nil
 }
@@ -107,6 +111,10 @@ func (s *EventHandlerService) HandleHomeChannelCheckpointed(ctx context.Context,
 	}
 
 	if err := tx.RefreshUserEnforcedBalance(channel.UserWallet, channel.Asset); err != nil {
+		return err
+	}
+
+	if err := tx.UpdateStateUserSigIfMissing(event.ChannelID, event.StateVersion, event.UserSig); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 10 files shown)


ihsraham (approved):
approved for the m-01 closure. the fix carries the user signature from the accepted on-chain state into the local row, and it only fills the field when it is still missing. that closes the blocked-channel path without widening the rpc trust boundary.

the local build, tests, vet, targeted race tests, and pr snyk checks are clean for this fix. the remaining govulncheck output is the separate go toolchain bump item, not a blocker for m-01.

nksazonov (approved):
Fix is correct and complete for the reported attack vector.

Minor gap: `HandleHomeChannelMigrated` (service.go) is a stub and the only handler without an `UpdateStateUserSigIfMissing` call — even though the reactor already extracts `UserSig` for that event. Low risk today (handler is unimplemented), but worth adding the backfill call alongside the TODO before the handler is fleshed out.

PR #733, opened 2026-05-08
MF-M05: fix(nitronode): enforce TLS by default for Postgres

philanton:
## Summary
- Replace hardcoded `sslmode=disable` in `postgresqlDbUrl` with configurable `NITRONODE_DATABASE_SSLMODE` (binary default: `require`, validated against the libpq list).
- Honor `NITRONODE_DATABASE_URL` verbatim when set so operators can supply a fully-qualified DSN; individual host/user/password/sslmode fields are ignored in that case.
- Wire helm chart helper to propagate `config.database.sslmode` as `NITRONODE_DATABASE_SSLMODE`. Chart default is `require` so helm and the binary agree; deployments where Postgres is only reachable on a private network (e.g. cluster-internal pgbouncer, VPC-only Cloud SQL) can opt out by setting `sslmode: disable`. The stress-v1 profile does so explicitly.

## Why
Audit finding MF-M05: every Postgres connection (schema create, migrations, runtime) used `sslmode=disable`, so credentials and protocol traffic could be observed or tampered with by anyone on the path. The chart documented an `sslmode` value but never propagated it.

## Notes
- Binary default is `require` (TLS without cert verification). Operators wanting strict cert checking should set `verify-full`.
- Local dev / testcontainers paths still use `sslmode=disable` intentionally.
- Release note: direct binary users without TLS-capable Postgres must set `NITRONODE_DATABASE_SSLMODE=disable` explicitly. Helm users on a private-network Postgres must set `config.database.sslmode: disable` to keep the previous behavior.

## Test plan
- [x] `go test ./nitronode/store/database/ -run TestPostgresqlDbUrl` — covers default, explicit, invalid, schema append, URL precedence, unsupported driver
- [x] `go vet ./nitronode/store/database/` clean
- [x] `go build ./nitronode/...` clean
- [x] Helm template renders correctly with `sslmode` set / unset (default → `require`, override → `disable`)
- [x] Manual smoke against TLS-enabled Postgres

### DIFF
--- a/nitronode/chart/templates/helpers/_common.tpl
+++ b/nitronode/chart/templates/helpers/_common.tpl
@@ -87,6 +87,8 @@ Returns common environment variables
   value: "{{ print .port }}"
 - name: NITRONODE_DATABASE_USERNAME
   value: {{ .user }}
+- name: NITRONODE_DATABASE_SSLMODE
+  value: {{ .sslmode | default "disable" | quote }}
 {{- end }}
 {{- range $key, $value := .Values.config.extraEnvs }}
 - name: {{ $key | upper }}
--- a/nitronode/chart/templates/helpers/_common.tpl
+++ b/nitronode/chart/templates/helpers/_common.tpl
@@ -88,7 +88,7 @@ Returns common environment variables
 - name: NITRONODE_DATABASE_USERNAME
   value: {{ .user }}
 - name: NITRONODE_DATABASE_SSLMODE
-  value: {{ .sslmode | default "disable" | quote }}
+  value: {{ .sslmode | default "require" | quote }}
 {{- end }}
 {{- range $key, $value := .Values.config.extraEnvs }}
 - name: {{ $key | upper }}
--- a/nitronode/README.md
+++ b/nitronode/README.md
@@ -85,7 +85,8 @@ assets:
 |----------|-------------|---------|
 | `NITRONODE_SIGNER_KEY` | Private key for signing node state updates | (Required) |
 | `NITRONODE_DATABASE_DRIVER` | `sqlite` or `postgres` | `sqlite` |
-| `NITRONODE_DATABASE_URL` | Connection string or file path | `nitronode.db` |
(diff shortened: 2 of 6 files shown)


nitronode/chart/templates/helpers/_common.tpl
```diff
@@ -87,6 +87,8 @@ Returns common environment variables
   value: "{{ print .port }}"
 - name: NITRONODE_DATABASE_USERNAME
   value: {{ .user }}
+- name: NITRONODE_DATABASE_SSLMODE
+  value: {{ .sslmode | default "disable" | quote }}
```
ihsraham: this still leaves the helm postgres path on plaintext by default. `values.yaml` keeps `config.database.sslmode: disable`, and this line sends that into `NITRONODE_DATABASE_SSLMODE`, so a helm user who only switches the driver to postgres will override the new go default of `require`.

for m-05 closure, please default this to `require` for postgres, or make the chart value default to `require`. if keeping `disable` is intentional for compatibility, then we should call the helm path out as opt-in instead of saying postgres tls is enforced by default.

nksazonov (commented):
The Go-layer fix is solid: the `SSLMode` field, URL-override path, validation map, and tests are all correct. One gap remains on the chart side — the template default in `_common.tpl` still ships `disable`, which means TLS is opt-in rather than opt-out for any deployment that doesn't explicitly set the value. Everything else looks good.

nitronode/chart/templates/helpers/_common.tpl
philanton: Yeah, good catch. Flipped the chart default to `require` in [3fc1c86](https://github.com/layer-3/nitrolite/pull/733/commits/3fc1c86ffe1b769afba58ee72d79917b3c9c1ede) so helm and the binary agree. stress-v1 overrides back to `disable` explicitly because its pgbouncer is cluster-internal — I left a comment in `values.yaml` calling that case out, so private-network deployments know they can opt out without losing the default-secure posture.

ihsraham (approved):
approved for m-05 closure. the postgres path now defaults to tls in both the binary and helm, with explicit opt-out documented for private-network setups.

nksazonov (approved):

PR #734, opened 2026-05-08
Fix SDK acknowledgement before home channel creation

ihsraham:
## Summary
- Fixes `Client.acknowledge(asset)` so received off-chain funds can be acknowledged before a home channel is open.
- Routes latest states without `homeChannelId` through the existing channel creation acknowledgement path.
- Bumps `@yellow-org/sdk` to `1.2.2` and adds unit coverage for acknowledgement path selection.

## Context
While adding channel-readiness UX to the Nitrolite Store example, we found that the app can detect received off-chain funds, but the browser SDK needs to acknowledge those funds through channel creation when no home channel exists yet. This aligns the TS SDK behavior with the Go SDK path for `state exists but no HomeChannelID`.

## Validation
- `npm --prefix sdk/ts run typecheck`
- `npm --prefix sdk/ts test -- --runTestsByPath test/unit/client.test.ts`
- `npm --prefix sdk/ts test`
- `npm --prefix sdk/ts run lint`
- `npm --prefix sdk/ts run build:ci`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Broadened handling for cases where an on-chain channel is not present, improving acknowledgement and transfer flows.

* **Tests**
  * Expanded unit tests to cover acknowledgement and transfer edge cases and state transitions.

* **Chores**
  * Package version bumped to 1.2.2.

[Review Change Stack](https://app.coderabbit.ai/change-stack/layer-3/nitrolite/pull/734)

### DIFF
--- a/sdk/ts/src/client.ts
+++ b/sdk/ts/src/client.ts
@@ -639,7 +639,8 @@ export class Client {
    * or to acknowledge channel creation without a deposit.
    *
    * This method handles two scenarios automatically:
-   * 1. If no channel exists: Creates a new channel with the acknowledgement transition
+   * 1. If no channel exists, including received off-chain funds before channel creation:
+   *    Creates a new channel with the acknowledgement transition
    * 2. If channel exists: Advances the state with an acknowledgement transition
    *
    * The returned state is signed by both the user and the node.
@@ -663,12 +664,10 @@ export class Client {
       // No state exists
     }
 
-    // homeChannelId is intentionally not checked here: a non-null state with
-    // an undefined homeChannelId is valid — it represents a user who received
-    // funds but has not yet opened a channel. Only enter the creation path when
-    // there is truly no prior state at all.
-    // No channel path - create channel with acknowledgement
-    if (!state) {
+    // No open channel path - create channel with acknowledgement. A non-null
+    // state without a homeChannelId represents received off-chain funds before
+    // the user has opened a home channel.
+    if (!state || !state.homeChannelId) {
       // Get supported sig validators bitmap from node config
       const bitmap = await this.getSupportedSigValidatorsBitmap();
 
--- a/sdk/ts/src/client.ts
+++ b/sdk/ts/src/client.ts
@@ -540,7 +540,8 @@ export class Client {
   /**
    * Transfer prepares a transfer state to send funds to another wallet address.
    * This method handles two scenarios automatically:
-   * 1. If no channel exists: Creates a new channel with the transfer transition
+   * 1. If no open channel exists, including received off-chain funds before channel creation:
+   *    Creates a new channel with the transfer transition
    * 2. If channel exists: Advances the state with a transfer send transition
    *
    * The returned state is signed by both the user and the node. For existing channels,
@@ -569,11 +570,10 @@ export class Client {
       // Channel doesn't exist
     }
 
-    // homeChannelId is intentionally not checked here: a non-null state with
-    // an undefined homeChannelId is valid — it represents a user who received
-    // funds but has not yet opened a channel. Only enter the creation path when
-    // there is truly no prior state at all.
-    if (!state) {
+    // No open channel path - create channel with transfer. A non-null state
+    // without a homeChannelId represents received off-chain funds before the
+    // user has opened a home channel.
+    if (!state || !state.homeChannelId) {
       // Get supported sig validators bitmap from node config
       const bitmap = await this.getSupportedSigValidatorsBitmap();
 
@@ -664,6 +664,10 @@ export class Client {
       // No state exists
     }
 
+    if (state?.userSig) {
+      throw new Error('state already acknowledged by user');
+    }
+
     // No open channel path - create channel with acknowledgement. A non-null
     // state without a homeChannelId represents received off-chain funds before
     // the user has opened a home channel.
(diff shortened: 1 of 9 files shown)


sdk/ts/test/unit/client.test.ts
```diff
@@ -82,3 +138,58 @@ describe('Client.getOnChainBalance', () => {
...
+        const state = await client.acknowledge('usdc');
+
+        expect(client.requestChannelCreation).toHaveBeenCalledTimes(1);
+        expect(client.signAndSubmitState).not.toHaveBeenCalled();
+        expect(state.version).toBe(latestState.version + 1n);
+        expect(state.homeChannelId).toBeDefined();
+        expect(state.transition.type).toBe(core.TransitionType.Acknowledgement);
+        expect(state.userSig).toBe(USER_SIGNATURE);
+        expect(state.nodeSig).toBe(NODE_SIGNATURE);
+    });
```
nksazonov, line 169: This test confirms routing, version increment, `homeChannelId`, transition type, and signatures — but never asserts that the user's existing off-chain balance was carried into the new channel state. The central correctness property of this fix — preserving received funds through channel creation — goes unverified. A regression that zeros out `homeLedger.userBalance`/`userNetFlow` would pass all assertions here.

Suggestion: add after the existing assertions:
```ts
expect(state.homeLedger.userBalance.equals(latestState.homeLedger.userBalance)).toBe(true);
expect(state.homeLedger.userNetFlow.equals(latestState.homeLedger.userNetFlow)).toBe(true);
```

sdk/ts/test/unit/client.test.ts
```diff
@@ -1,6 +1,62 @@
...
+    client.assetStore = {
+        getSuggestedBlockchainId: jest.fn().mockResolvedValue(11155111n),
+        getTokenAddress: jest.fn().mockResolvedValue(TOKEN_ADDRESS),
+    };
+    client.getNodeAddress = jest.fn().mockResolvedValue(NODE_ADDRESS);
+    client.signState = jest.fn().mockResolvedValue(USER_SIGNATURE);
+    client.requestChannelCreation = jest.fn().mockImplementation(async (state: core.State) => {
+        state.nodeSig = NODE_SIGNATURE;
+        return NODE_SIGNATURE;
+    });
```
nksazonov: The mock sets `state.nodeSig` directly on the incoming state object, but the real `requestChannelCreation` (`client.ts:1948`) only returns a signature string — it never mutates the state. The caller at `client.ts:711` then assigns `newState.nodeSig = nodeSig`. Because the mock both mutates and returns the same value, the `nodeSig` assertions pass regardless of whether the assignment at line 711 actually executes.

Suggestion: simplify to `jest.fn().mockResolvedValue(NODE_SIGNATURE)` so the mock only simulates what the real function does.

nksazonov (commented):
Good fix!

Two things to be aware of beyond the inline comments:

**`transfer()` has the same pre-fix pattern** (`client.ts:576` still uses `if (!state)` without a `!state.homeChannelId` guard). The Go SDK's `Transfer()` already handles `state.HomeChannelID == nil` by routing to channel creation. In the TS SDK, a user with received off-chain funds calling `transfer()` before `acknowledge()` will reach `signAndSubmitState` — and the node will reject with "user has no open channel" (`submit_state.go:56–58`). The existing comment at `client.ts:572` saying "homeChannelId is intentionally not checked here" is now misleading given this PR's rationale. Consider either fixing `transfer()` the same way or updating the comment to document that `acknowledge()` must be called first.


The fix to the aforementioned issue should probably be done in a separate PR. Let's wait for @philanton to decide together.

nksazonov:
@ihsraham , after speaking to @philanton , could you please make the same change for `transfer(...)` logic in the TS SDK?

sdk/ts/test/unit/client.test.ts
ihsraham: fixed in [`2052e6af`](https://github.com/layer-3/nitrolite/commit/2052e6af). added balance preservation checks for `userBalance` and `userNetFlow`.

sdk/ts/test/unit/client.test.ts
ihsraham: fixed in [`2052e6af`](https://github.com/layer-3/nitrolite/commit/2052e6af). the mock now only returns the signature, matching the real method.

sdk/ts/src/client.ts
```diff
@@ -663,12 +664,10 @@ export class Client {
...
-    // homeChannelId is intentionally not checked here: a non-null state with
-    // an undefined homeChannelId is valid — it represents a user who received
-    // funds but has not yet opened a channel. Only enter the creation path when
-    // there is truly no prior state at all.
-    // No channel path - create channel with acknowledgement
-    if (!state) {
+    // No open channel path - create channel with acknowledgement. A non-null
+    // state without a homeChannelId represents received off-chain funds before
+    // the user has opened a home channel.
+    if (!state || !state.homeChannelId) {
```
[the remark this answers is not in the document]
ihsraham: fixed in [`2052e6af`](https://github.com/layer-3/nitrolite/commit/2052e6af). moved the already-acknowledged guard before choosing the channel creation path.

ihsraham:
@nksazonov done in [`2052e6af`](https://github.com/layer-3/nitrolite/commit/2052e6af). `transfer()` now uses the same no-home-channel creation path as `acknowledge()`.

philanton (approved):

nksazonov (approved):

PR #735, opened 2026-05-08
MF-M07: Unblock receiver states after finalized escrow operations

philanton:
## Summary
- Replace the `lastSignedState.EscrowChannelID != nil` guard in `issueTransferReceiverState` (`channel_v1`) and `issueReleaseReceiverState` (`app_session_v1`) with a new `EnsureNoOngoingEscrowOperation(wallet, asset)` store method.
- Predicate keys on the latest signed transition type instead of the raw `EscrowChannelID` field: `EscrowLock` / `MutualLock` always block; `EscrowDeposit` / `EscrowWithdraw` block only when the on-chain escrow channel `state_version` hasn't caught up; everything else allows.
- Adds DB-store tests covering all branches (no prior state, non-escrow, lock-types, finalize synced/unsynced, unsigned).

## Background (audit MF-M07)
The previous guard rejected any receiver-side state whenever the receiver's latest signed state had a non-nil `EscrowChannelID`. The field is cleared by `State.NextState()` only when the previous transition was `EscrowDeposit` or `EscrowWithdraw`, so the signed state itself retains it after a finalized escrow flow. As a result, a user who completed a deposit/withdrawal could be unable to receive transfers or app-session releases until an unrelated off-chain action advanced past the finalized state.

The replacement matches the existing `EnsureNoOngoingStateTransitions` pattern (transition type + state.version vs channel.state_version) but is scoped to escrow concerns only, so home-deposit / home-withdrawal logic is untouched.

## Test plan
- [x] `go test ./nitronode/store/database/... -run EnsureNoOngoingEscrowOperation -v`
- [x] `go test ./nitronode/...`
- [x] `go test ./pkg/...`

## Summary by CodeRabbit

* **Bug Fixes**
  - Enhanced escrow operation validation to detect and block state transitions when escrow activities are in-flight, including locks, deposits, and withdrawals for wallet-asset pairs.
  - Strengthened operational safeguards by preventing state changes during pending escrow operations not yet finalized on-chain.
  - Improved consistency of escrow-related state management across payment channel and application session handlers.

### DIFF
--- a/nitronode/store/database/db_store.go
+++ b/nitronode/store/database/db_store.go
@@ -237,3 +237,65 @@ func (s *DBStore) EnsureNoOngoingStateTransitions(wallet, asset string) error {
 
 	return nil
 }
+
+// EnsureNoOngoingEscrowOperation validates that the user has no in-flight escrow
+// operation that would prevent the node from issuing receiver-side states (transfer
+// receive, app-session release).
+//
+// Validation logic by latest signed transition type:
+//   - escrow_lock / mutual_lock: always considered ongoing (no finalization yet)
+//   - escrow_deposit / escrow_withdraw: only considered ongoing when the on-chain
+//     escrow channel state version has not caught up with the signed state version
+//   - any other transition: not an escrow operation, allow
+func (s *DBStore) EnsureNoOngoingEscrowOperation(wallet, asset string) error {
+	wallet = strings.ToLower(wallet)
+
+	type escrowCheck struct {
+		TransitionType       core.TransitionType
+		StateVersion         uint64
+		EscrowChannelVersion *uint64
+	}
+
+	var result escrowCheck
+	tx := s.db.Raw(`
+		SELECT
+			s.transition_type as transition_type,
+			s.version as state_version,
+			ec.state_version as escrow_channel_version
+		FROM channel_states s
+		LEFT JOIN channels ec ON ec.channel_id = s.escrow_channel_id
+		WHERE s.user_wallet = ?
+			AND s.asset = ?
+			AND s.user_sig IS NOT NULL
+			AND s.node_sig IS NOT NULL
+		ORDER BY s.epoch DESC, s.version DESC
+		LIMIT 1
+	`, wallet, asset).Scan(&result)
+
+	if tx.Error != nil {
+		return fmt.Errorf("failed to check ongoing escrow operation: %w", tx.Error)
+	}
+	if tx.RowsAffected == 0 {
+		return nil
+	}
+
+	switch result.TransitionType {
+	case core.TransitionTypeEscrowLock:
+		return fmt.Errorf("escrow lock is still ongoing")
+
+	case core.TransitionTypeMutualLock:
+		return fmt.Errorf("mutual lock is still ongoing")
+
+	case core.TransitionTypeEscrowDeposit:
+		if result.EscrowChannelVersion != nil && result.StateVersion != *result.EscrowChannelVersion {
+			return fmt.Errorf("escrow deposit finalization is still ongoing")
+		}
+
+	case core.TransitionTypeEscrowWithdraw:
+		if result.EscrowChannelVersion != nil && result.StateVersion != *result.EscrowChannelVersion {
+			return fmt.Errorf("escrow withdrawal finalization is still ongoing")
+		}
+	}
+
+	return nil
+}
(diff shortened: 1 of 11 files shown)


ihsraham (approved):
approved for m-07 closure.

the fix removes the broad `EscrowChannelID != nil` block and uses the escrow-state check on both receiver paths. finalized deposit and withdrawal states can receive again after the escrow channel version has caught up, while real in-flight locks still block. that matches the finding and keeps the safer on-chain sync guard.

nitronode/store/database/db_store.go
```diff
@@ -237,3 +237,65 @@ func (s *DBStore) EnsureNoOngoingStateTransitions(wallet, asset string) error {
...
+
+	case core.TransitionTypeEscrowDeposit:
+		if result.EscrowChannelVersion != nil && result.StateVersion != *result.EscrowChannelVersion {
+			return fmt.Errorf("escrow deposit finalization is still ongoing")
+		}
+
+	case core.TransitionTypeEscrowWithdraw:
+		if result.EscrowChannelVersion != nil && result.StateVersion != *result.EscrowChannelVersion {
+			return fmt.Errorf("escrow withdrawal finalization is still ongoing")
+		}
```
nksazonov, line 297: **Behavioral gap: `EscrowChannelVersion == nil` silently allows instead of blocking.**

The condition `result.EscrowChannelVersion != nil && result.StateVersion != *result.EscrowChannelVersion` treats a nil version (LEFT JOIN found no row in `channels`) as finalized. If `s.escrow_channel_id` is set on the signed state but the corresponding row is absent from `channels` (e.g. node restart before the channel record was synced, or DB inconsistency), the on-chain status is unknown yet the function allows issuance of receiver states.

Suggested fix: invert the nil guard so a missing channel record is treated as pending: change to `if result.EscrowChannelVersion == nil || result.StateVersion != *result.EscrowChannelVersion` and return the finalization-in-progress error.

nitronode/store/database/db_store_test.go
```diff
@@ -646,3 +646,209 @@ func TestDBStore_EnsureNoOngoingStateTransitions(t *testing.T) {
...
+		store := NewDBStore(db)
+		require.NoError(t, store.CreateChannel(homeChannel))
+		require.NoError(t, store.CreateChannel(newEscrowChannel(1)))
+
+		state := newSignedState(2, core.TransitionTypeEscrowLock, true)
+		state.UserSig = nil
+		state.NodeSig = nil
+		storeState(t, store, state)
+
+		err := store.EnsureNoOngoingEscrowOperation(wallet, asset)
```
nksazonov, line 1214: **Missing test: `EscrowDeposit`/`EscrowWithdraw` when the escrow channel row is absent from `channels`.**

All existing sub-tests for deposit and withdrawal rely on a `channels` row being present. There is no coverage for the case where `escrow_channel_id` is set on the signed state but `CreateChannel` was never called for the escrow channel, which is precisely the path affected by the nil-version gap above.

Suggested fix: add sub-tests `"EscrowDeposit - escrow channel missing from DB - block"` and `"EscrowWithdraw - escrow channel missing from DB - block"` that store an escrow-type signed state without creating the escrow channel, then assert that `EnsureNoOngoingEscrowOperation` returns an error.

nksazonov (changes requested):
The PR correctly addresses the MF-M07 audit finding. The `EscrowChannelID != nil` guard was replaced with a purpose-built `EnsureNoOngoingEscrowOperation()` method that correctly distinguishes unresolved escrow locks from finalized deposit/withdrawal states. The approach is more conservative than the audit recommended (it additionally requires on-chain version confirmation for deposit/withdraw states, rather than relying solely on `NextState()` semantics), which is a sound design choice. The test coverage is thorough for the sunny-path variants.

There is one behavioral gap in the new predicate and a corresponding missing test case that should be addressed before merging.

nitronode/store/database/db_store.go
philanton: Yeah, good catch. Fixed in [b91766d](https://github.com/layer-3/nitrolite/pull/735/changes/b91766d5e312c77acc35818a549481cffbd2e5ea) — flipped the nil guard to `== nil ||` so a missing `channels` row is treated as pending. Applied the same fix to `EnsureNoOngoingStateTransitions` since all 6 branches had the same gap.

nitronode/store/database/db_store_test.go
philanton: Added in [b91766d](https://github.com/layer-3/nitrolite/pull/735/changes/b91766d5e312c77acc35818a549481cffbd2e5ea) — sub-tests `EscrowDeposit/EscrowWithdraw - escrow channel missing from DB - block` cover the gap. Also added the equivalent missing-channel-row tests for each branch in `EnsureNoOngoingStateTransitions` since the same nil-guard fix landed there.

nksazonov (approved):
Good work!

PR #737, opened 2026-05-08
MF-L02: docs(protocol): qualify enforcement guarantee for intent-specific execution paths

nksazonov:
Qualifies the enforcement guarantee across three protocol docs.

The blanket statement "Any party MAY submit the latest signed state at any time" is not fully accurate: an OPERATE state with non-zero user net flow delta cannot be used to create or checkpoint a channel from VOID status, because OPERATE requires zero user net flow change relative to the last enforced state.

**Changes:**
- `docs/protocol/security-and-limitations.md`: rewrite Enforcement Guarantees section — adds intent-path qualifier and a new bullet requiring parties to retain intermediate states (e.g. DEPOSIT) before discarding them
- `docs/protocol/enforcement.md`: qualify the Enforcement Model bullet; add caveat in Channel Creation section about OPERATE intent not being usable when user net flow was accumulated from an unconfirmed DEPOSIT state
- `docs/protocol/overview.md`: qualify the State Enforcement paragraph

## Documentation
- Updated protocol enforcement submission rules, including requirements for valid execution paths when submitting states
- Clarified channel creation restrictions and requirements for state operations
- Enhanced enforcement guarantees documentation with improved clarity on state management and enforcement behavior

### DIFF
--- a/docs/protocol/enforcement.md
+++ b/docs/protocol/enforcement.md
@@ -31,7 +31,7 @@ Node-issued pending states (those carrying only the node's signature) are NOT en
 Off-chain states and on-chain enforcement states are related as follows:
 
 - Participants advance state off-chain through signed updates
-- At any time, any party MAY submit the latest mutually signed state to the blockchain layer
+- Any party MAY submit the latest mutually signed state to the blockchain layer, provided a valid execution path exists for that state's intent in the current channel context
 - The blockchain layer validates the submitted state and updates its record
 - On-chain state always reflects the latest successfully checkpointed state
 
@@ -69,7 +69,7 @@ The creation process:
 3. The blockchain layer validates signatures, creates the channel record, and applies fund effects according to the state's intent
 4. The channel is now active on the on-chain layer
 
-The state submitted for channel creation MAY carry a DEPOSIT, WITHDRAW, or OPERATE intent.
+The state submitted for channel creation MAY carry a DEPOSIT, WITHDRAW, or OPERATE intent. OPERATE intent requires the user net flow delta to be zero relative to the previous on-chain state (which is the empty state for a new channel). An OPERATE state that carries accumulated user net flow from an unconfirmed prior DEPOSIT state cannot be used to create a channel — parties MUST enforce the DEPOSIT state on-chain before advancing to subsequent OPERATE states that depend on it.
 
 ## State Submission
 
--- a/docs/protocol/enforcement.md
+++ b/docs/protocol/enforcement.md
@@ -69,7 +69,7 @@ The creation process:
 3. The blockchain layer validates signatures, creates the channel record, and applies fund effects according to the state's intent
 4. The channel is now active on the on-chain layer
 
-The state submitted for channel creation MAY carry a DEPOSIT, WITHDRAW, or OPERATE intent. OPERATE intent requires the user net flow delta to be zero relative to the previous on-chain state (which is the empty state for a new channel). An OPERATE state that carries accumulated user net flow from an unconfirmed prior DEPOSIT state cannot be used to create a channel — parties MUST enforce the DEPOSIT state on-chain before advancing to subsequent OPERATE states that depend on it.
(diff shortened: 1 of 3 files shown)


philanton (approved):

docs/protocol/overview.md
```diff
@@ -60,7 +60,7 @@ A state represents the current agreed asset allocations and metadata shared betw
 User and a node advance states off-chain by exchanging signed state transitions. Each new state MUST have a version exactly one greater than the previous state. Transitions include deposits, withdrawals, transfers, commits, releases, escrow operations, and migrations.
 
 **State Enforcement**
-Any party MAY submit the latest signed state to the blockchain layer for on-chain enforcement. The blockchain layer validates signatures, version ordering, and ledger invariants before accepting a state.
+Any party MAY submit the latest mutually signed state to the blockchain layer for on-chain enforcement, provided that state's intent has a valid execution path in the current channel context. The blockchain layer validates signatures, version ordering, and ledger invariants before accepting a state.
```
ihsraham, line 63: could we also update the opening paragraph and design goal in this file? line 5 still says a user can enforce the latest agreed state “at any time,” and line 26 says users can fall back to enforce the latest state.

those old sentences can be read as the same guarantee this pr is correcting. adding the same valid-execution-path qualifier there would make the docs consistent.

docs/protocol/security-and-limitations.md
```diff
@@ -41,7 +41,8 @@ Each state update MUST satisfy transition-specific rules. Invalid transitions ar
 
 The blockchain layer provides the following guarantees:
 
-- Any party MAY submit the latest signed state at any time
+- Any party MAY submit the latest mutually signed state to the blockchain layer; enforcement succeeds when a valid execution path exists for that state's intent in the current channel context
```
ihsraham, line 44: worth updating the earlier security goal in this same file too. line 20 still says the latest mutually signed state can always be enforced on-chain, which conflicts with this new qualifier.

for l-02 closure, the guarantee should be consistent everywhere: the latest signed state is enforceable only when its intent has a valid execution path, and users must keep any needed intermediate states.

docs/protocol/enforcement.md
```diff
@@ -69,7 +69,7 @@ The creation process:
 3. The blockchain layer validates signatures, creates the channel record, and applies fund effects according to the state's intent
 4. The channel is now active on the on-chain layer
 
-The state submitted for channel creation MAY carry a DEPOSIT, WITHDRAW, or OPERATE intent.
+The state submitted for channel creation MAY carry a DEPOSIT, WITHDRAW, or OPERATE intent. OPERATE intent requires the user net flow delta to be zero relative to the previous on-chain state (which is the empty state for a new channel). An OPERATE state that carries accumulated user net flow from an unconfirmed prior DEPOSIT state cannot be used to create a channel — parties MUST enforce the DEPOSIT state on-chain before advancing to subsequent OPERATE states that depend on it.
```
ihsraham: this should match the pr note and the contract behavior: the `OPERATE` state is not only blocked for creation, it is also blocked for checkpointing or advancement when the prior `DEPOSIT` was not enforced.

could we say “cannot be used to create or checkpoint a channel” and use “unenforced” instead of “unconfirmed”? that makes the user action clear: keep the `DEPOSIT` state until it has been submitted on-chain.

docs/protocol/overview.md
nksazonov: 👍 Fixed in dcef8284 — updated both the opening paragraph (line 5) and the "Blockchain security guarantees" design goal (line 26) with the valid-execution-path qualifier.

docs/protocol/security-and-limitations.md
nksazonov: 👍 Fixed in dcef8284 — "State finality" goal now reads: "the latest mutually signed state can be enforced on-chain when a valid execution path exists for its intent; parties MUST retain any intermediate states required to establish that path".

docs/protocol/enforcement.md
nksazonov: 👍 Fixed in dcef8284 — changed "unconfirmed" to "unenforced" and expanded "cannot be used to create a channel" to "cannot be used to create or checkpoint a channel" in both enforcement.md and the matching bullet in security-and-limitations.md.

ihsraham (approved):
approved, this now closes `L-02` with the enforcement-path caveat documented in the right places.

PR #739, opened 2026-05-08
MF-H02: bind session key registration to a single owner per kind

philanton:
## Summary

- Block cross-wallet session-key shadow registration by adding `UNIQUE (session_key, kind)` to `current_session_key_states_v1` and rejecting collisions with a generic `invalid_session_key_state: session_key not allowed` (no ownership leak).
- Require a `session_key_sig` co-signature from the session-key holder on every submit, so nobody can register a key they do not control. App side reuses the existing packed payload (already binds `user_address`); channel side gets a new `PackChannelSessionKeyOwnershipV1` that adds `user_address` to the ownership payload.
- Reject self-collisions (`user_address == session_key`).

### Why

Without these, an attacker could register the victim's session key under the attacker's wallet, outpace the victim's version, and have `GetAppSessionKeyOwner` (which resolved ownership by `MAX(version)` over `session_key` alone) return the attacker for the victim's signatures. Result: the victim's delegated quorum signatures got attributed to the attacker or rejected as non-participant.

An attacker without the session-key private key never reaches the pointer lookup — they fail at `session_key_sig does not match session_key` and learn nothing about existing registrations. The generic `session_key not allowed` response is reachable only when the submitter can produce a valid possession proof, i.e. when they already control the key in question; surfacing existence to that caller is acceptable.

### Migration

`20260508000000_session_key_ownership_constraints.sql` runs a pre-flight `RAISE EXCEPTION` on existing `(session_key, kind)` duplicates before adding the constraint. Any duplicates surface for ops review since they are exploitation evidence under the old behavior.

### SDK + tooling

- `sdk/go`: `SignAppSessionKeyOwnership` / `SignChannelSessionKeyOwnership` helpers.
- `sdk/ts`: `client.signAppSessionKeyOwnership` / `client.signChannelSessionKeyOwnership` + `packChannelSessionKeyOwnershipV1`.
- `sdk/ts-compat`: facade methods exposed.
- `cerebro` CLI: requires the session-key private key up-front and refuses to register a key it does not control.
- `example-app`: enable + active-disable populate `session_key_sig`; arbitrary-key revoke surfaces a friendly error since the example app does not retain other keys' private keys.

### Known design implication

`session_key_sig` is required on every submit, including revokes. A wallet that loses the session-key private key cannot revoke its own delegation early — it has to wait for `expires_at`. Documented inline in the SDK READMEs and example app.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] `cd sdk/ts && npm run typecheck && npm test && npm run lint`
- [x] `cd sdk/ts-compat && npm test && npm run lint`


pkg/core/session_key.go
```diff
@@ -91,6 +92,27 @@ func PackChannelKeyStateV1(sessionKey string, metadataHash common.Hash) ([]byte,
 	return packed, nil
 }
 
+// PackChannelSessionKeyOwnershipV1 packs the bytes the session-key holder signs to prove
+// possession of the key being registered. The user_address is bound into the payload so the
+// signature cannot be replayed under a different wallet; the session_key itself is recovered
+// from the signature and need not be repeated in the packed bytes.
+func PackChannelSessionKeyOwnershipV1(userAddress string, metadataHash common.Hash) ([]byte, error) {
```
ihsraham: i would block on this payload shape. `user_sig` signs `(session_key, metadataHash)`, while this ownership proof signs `(user_address, metadataHash)`, so both messages are the same `(address, bytes32)` shape.

that lets a prior reciprocal channel `user_sig` be reused as `session_key_sig`. please add domain separation and include the purpose or kind plus both `user_address` and `session_key` in this ownership payload.

nitronode/config/migrations/postgres/20260508000000_session_key_ownership_constraints.sql
```diff
@@ -0,0 +1,45 @@
...
+ALTER TABLE app_session_key_states_v1
+    ADD COLUMN session_key_sig TEXT;
+
+ALTER TABLE channel_session_key_states_v1
+    ADD COLUMN session_key_sig TEXT;
+
+-- Pre-flight: refuse the migration if duplicate (session_key, kind) rows are present in
+-- current_session_key_states_v1. Such rows are evidence of cross-wallet collisions that
+-- the old code path allowed; manual remediation is required before the constraint adds.
+DO $$
```
ihsraham, line 18: this `DO $$` block needs `-- +goose StatementBegin/StatementEnd`. `goose validate -v` parses this file as 8 up statements, so the PL/pgSQL block can be split before the unique constraint is added.

wrapping the block keeps the migration executable, which is required for the `(session_key, kind)` invariant to exist in postgres.

nitronode/store/database/current_session_key_state.go
```diff
@@ -66,55 +72,56 @@ func upsertCurrentSessionKeyState(tx *gorm.DB, userAddress, sessionKey string, k
...
 
-// LockSessionKeyState ensures a pointer row exists for (user, session_key, kind) and locks it
-// for the duration of the surrounding transaction. Returns the current version (0 if newly
-// created). Mirrors LockUserState. On non-postgres dialects, falls back to read-without-lock.
+// LockSessionKeyState seeds the pointer row for (userAddress, session_key, kind) if absent
+// and locks the (session_key, kind) row for the surrounding transaction. Returns the latest
+// stored version for the caller's row, or ErrSessionKeyNotAllowed if the key is bound to a
+// different wallet for this kind.
+//
+// The (session_key, kind) unique constraint guarantees there is at most one pointer row per
```
ihsraham, line 85: this relies on the `(session_key, kind)` unique constraint, but that constraint is only added by the postgres migration. sqlite uses AutoMigrate, so it can still allow multiple pointer rows for the same key and kind, and this query can read whichever row comes first.

if sqlite is still a supported runtime path, please add the same unique index to the model or sqlite migration path.

nitronode/config/migrations/postgres/20260508000000_session_key_ownership_constraints.sql
```diff
@@ -0,0 +1,45 @@
+-- +goose Up
+-- Bind a session_key to a single owner (per kind) and require co-signature at submit time.
+
+-- Co-signature: the session-key holder proves possession at registration and on every update.
+-- Nullable to accommodate rows written before this column existed; new submits enforce non-null
+-- in application code. Columns add first so a constraint failure below does not leave the
+-- session_key_sig schema partially applied.
+ALTER TABLE app_session_key_states_v1
```
ihsraham, line 8: nullable is fine for reading old rows, but future writes should fail closed at the db layer too. otherwise an old binary during rollout can still insert a `NULL` `session_key_sig` row and make an unproven key binding durable.

one way to handle this is a `NOT VALID` check constraint for non-empty `session_key_sig`, plus a cleanup or quarantine path for existing current rows without proof.

ihsraham (commented):
i'd hold approval for now. the new submit path is closer, but the notes below still affect h-02 closure.

pkg/core/session_key.go
philanton: Yep, you're right. Folded `user_address` into `metadataHash` and dropped `PackChannelSessionKeyOwnershipV1` — both sigs now sign the same `PackChannelKeyStateV1(session_key, metadataHash)`, so `session_key` binds the packed bytes and `user_address` binds the hash. The payload is bound to a single `(wallet, session_key)` pair from both sides. Also merged the two validates since they share the payload now. Fixed in [6e01cda](https://github.com/layer-3/nitrolite/pull/739/changes/6e01cdabe5b1a6d4fe98971caa6dfdda734820cf)

nitronode/config/migrations/postgres/20260508000000_session_key_ownership_constraints.sql
philanton: Yep, missed the `StatementBegin/StatementEnd` markers. Fixed in [2e713a8](https://github.com/layer-3/nitrolite/pull/739/changes/2e713a86e1d5547089db2370e1c8b8a928e98600)

nitronode/store/database/current_session_key_state.go
philanton: Yep, AutoMigrate doesn't pick up the postgres constraint. Added a matching `uniqueIndex` gorm tag (same name as the postgres constraint so both paths converge) + a regression test that pins the invariant on the dialect that goes through AutoMigrate. Fixed in [2e713a8](https://github.com/layer-3/nitrolite/pull/739/changes/2e713a86e1d5547089db2370e1c8b8a928e98600)

nitronode/config/migrations/postgres/20260508000000_session_key_ownership_constraints.sql
philanton: Hmmm, I'd skip this one. Sig presence is enforced at the app layer twice already — `ValidateChannelSessionKeyStateV1` rejects empty `session_key_sig` and the handler early-returns before any write. The rolling-deploy angle is the only realistic source of NULL writes, but a pre-MF-H02 binary doesn't know about `current_session_key_states_v1` either, so its history-table writes would land orphaned without a matching pointer row — ownership reads key on the pointer, so an orphan history row isn't load-bearing for auth.

The `NOT VALID` check is cheap, but I don't see a path where it changes correctness — only adds belt to existing suspenders. I'd leave it out unless you've got a concrete scenario in mind)

nitronode/store/database/current_session_key_state.go
```diff
@@ -66,55 +77,56 @@ func upsertCurrentSessionKeyState(tx *gorm.DB, userAddress, sessionKey string, k
...
-			UpdatedAt:   time.Now().UTC(),
-		}
-		if err := s.db.Clauses(clause.OnConflict{DoNothing: true}).Create(&seed).Error; err != nil {
-			return 0, fmt.Errorf("failed to ensure current session key state row exists: %w", err)
-		}
+	seed := CurrentSessionKeyStateV1{
+		UserAddress: userAddress,
+		SessionKey:  sessionKey,
+		Kind:        kind,
+		Version:     0,
```
nksazonov, line 108: `LockSessionKeyState` inserts a seed row at `version=0` for `(user_address, session_key, kind)` before checking whether the key is already claimed. When a submit fails after the seed — version mismatch, cap exceeded, or any later error — the `version=0` row persists permanently. A different wallet can never claim that `(session_key, kind)` even though the original caller never completed a successful registration. `CountSessionKeysForUser` correctly excludes `version=0` rows from the cap so quotas aren't inflated, but the permanent ownership reservation from an aborted registration is non-obvious.

**Suggestion:** Add a comment at the seed site explicitly documenting that `version=0` rows are permanent ownership reservations (not just transient seeds), so future maintainers understand why cleanup on failure is intentionally absent and what the trade-off is.

nitronode/api/app_session_v1/submit_session_key_state.go
```diff
@@ -104,13 +117,38 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
...
+	recoveredKey, err := ethMsgRecoverer.Recover(packedState, keySigBytes)
+	if err != nil {
+		c.Fail(rpc.Errorf("invalid_session_key_state: failed to recover session_key_sig signer: %v", err), "")
+		return
+	}
+
+	if !strings.EqualFold(recoveredKey, coreState.SessionKey) {
+		c.Fail(rpc.Errorf("invalid_session_key_state: session_key_sig does not match session_key"), "")
+		return
+	}
```
nksazonov: The channel handler delegates both signature checks to `core.ValidateChannelSessionKeyAuthSigV1` and `core.ValidateChannelSessionKeySigV1` (two lines); this handler performs the equivalent hex-decode → create-validator → recover → compare logic inline (~40 lines). `pkg/app/session_key_v1.go` exports `PackAppSessionKeyStateV1` but no corresponding validate functions. The transaction body (`LockSessionKeyState` + cap check + version check) is also copied verbatim between both handlers, differing only in the `kind` constant and final store call.

**Suggestion:** Add `ValidateAppSessionKeyAuthSigV1` and `ValidateAppSessionKeyKeySigV1` to `pkg/app/session_key_v1.go` mirroring the channel pattern in `pkg/core/session_key.go`; this reduces the inline block to two delegating calls. The duplicated transaction body is a separate follow-up since it spans package boundaries.

nksazonov (commented):
Clean approach to a hard problem — the pointer-table + `(session_key, kind)` unique constraint elegantly closes the global-max-version attack, and the co-signature requirement is a solid possession proof. Two non-blocking observations below.

nitronode/config/migrations/postgres/20260508000000_session_key_ownership_constraints.sql
```diff
@@ -0,0 +1,47 @@
...
```
ihsraham, line 8: i would still block on this being nullable unless the deploy is guaranteed to drain old pods before this migration runs. the base branch already writes `current_session_key_states_v1`, so an old binary can still insert a current pointer with `session_key_sig = null`, and the new lookup will later trust it.

could we add `NOT VALID` check constraints on both history tables for `session_key_sig IS NOT NULL AND session_key_sig <> ''`? that keeps legacy rows from blocking the migration, but stops new unproven writes.

sdk/ts/src/client.ts
```diff
@@ -1676,9 +1679,39 @@ export class Client {
...
+   * bound into the metadata hash so a signature minted for one key cannot be replayed for
+   * another.
+   *
+   * @param state - The channel session key state to sign (session_key_sig field is excluded)
+   * @param sessionKeySigner - StateSigner whose address equals state.session_key
+   * @returns The hex-encoded signature string
+   */
+  async signChannelSessionKeyOwnership(
+    state: ChannelSessionKeyStateV1,
+    sessionKeySigner: StateSigner
```
ihsraham: this accepts any `StateSigner`, but some public `StateSigner` wrappers return prefixed or compound signatures. the node expects `session_key_sig` to be a raw 65-byte EIP-191 signature, so a normal caller can use this helper and still produce a request the server rejects.

could we narrow this to a raw message signer, or validate/strip the returned signature before returning it? same concern applies to `signAppSessionKeyOwnership` below.

sdk/go/app_session.go
```diff
@@ -393,3 +398,22 @@ func (c *Client) SignSessionKeyState(state app.AppSessionKeyStateV1) (string, er
 	// Strip the channel signer type prefix byte; session key auth uses plain EIP-191 signatures
 	return hexutil.Encode(sig[1:]), nil
 }
+
+// SignAppSessionKeyOwnership produces the session-key holder's ownership signature over the
+// packed app-session key state. The signer must be the holder of the session key being
+// registered; the resulting hex-encoded signature is intended to populate state.SessionKeySig
+// before submitting via SubmitAppSessionKeyState. The packed state already binds user_address,
+// so replay across wallets is not possible.
+func SignAppSessionKeyOwnership(state app.AppSessionKeyStateV1, sessionKeySigner sign.Signer) (string, error) {
```
ihsraham: same shape here: `sign.Signer` is broad enough to include wrappers that add protocol bytes, while the server recovers `SessionKeySig` as a raw Ethereum message signature. that makes the helper able to return a value the node will reject.

please either narrow the accepted signer type or validate the returned signature is the raw 65-byte form before returning it. the channel ownership helper has the same issue.

docs/api.yaml
```diff
@@ -380,6 +380,9 @@ types:
         - name: user_sig
           type: string
           description: User's signature over the session key metadata to authorize the registration/update of the session key
+        - name: session_key_sig
+          type: string
+          description: Session-key holder's signature over (user_address, metadata_hash) proving possession of the key being registered. Required on every submit to prevent registration of keys the submitter does not control.
```
ihsraham: this description does not match the channel implementation. channel `session_key_sig` signs `PackChannelKeyStateV1(session_key, metadataHash)`, with `user_address` inside `metadataHash`.

if external clients follow this wording and sign `(user_address, metadata_hash)` directly, their submits will fail. please align the docs with the helper payload.

sdk/ts/examples/example-app/README.md
```diff
@@ -261,6 +261,7 @@ await client.submitChannelSessionKeyState(state);
 
 // 5. Compute the metadata hash
 const metadataHash = getChannelSessionKeyAuthMetadataHashV1(
+  address,
```
ihsraham, line 269: this example still submits before setting `session_key_sig`. after this pr, `submitChannelSessionKeyState` rejects that request, so copying the snippet will fail.

please add `session_key_sig: ''` to the state and set it with `client.signChannelSessionKeyOwnership(...)` before submit. the revoke snippet below needs the same update.

ihsraham (commented):
i left a few closure comments inline. the main upgraded path looks much closer, but i would resolve these before approval.

sdk/ts/src/client.ts
philanton: Yep, narrowed both `signChannelSessionKeyOwnership` and `signAppSessionKeyOwnership` to accept `EthereumMsgSigner` directly instead of `StateSigner`. That sidesteps the prefix-strip question entirely — `EthereumMsgSigner.signMessage` returns the raw 65-byte EIP-191 sig the server expects, and the wrapping signers (`ChannelDefaultSigner`, `ChannelSessionKeyStateSigner`, etc.) no longer fit the param type. Fixed in [534b319](https://github.com/layer-3/nitrolite/pull/739/changes/534b3192d56d5eb16d271dbb384161b8fba84502)

sdk/go/app_session.go
philanton: Yep, same treatment as the TS side — narrowed `SignChannelSessionKeyOwnership` and `SignAppSessionKeyOwnership` to `*sign.EthereumMsgSigner`. Also widened `NewEthereumMsgSigner` / `NewEthereumMsgSignerFromRaw` return types from `sign.Signer` to `*sign.EthereumMsgSigner` so existing callers keep compiling without type-asserting (concrete still satisfies the interface). Fixed in [534b319](https://github.com/layer-3/nitrolite/pull/739/changes/534b3192d56d5eb16d271dbb384161b8fba84502)

docs/api.yaml
philanton: Yep, you're right — the description was lying. Updated it to say the signature is over `PackChannelKeyStateV1(session_key, metadata_hash)` (with `user_address` baked into `metadata_hash`), which matches the helper. Fixed in [c5810f3](https://github.com/layer-3/nitrolite/pull/739/changes/c5810f32c3fea1b81cc0816c4935a3e7fdd2724e)

sdk/ts/examples/example-app/README.md
philanton: Yep, the snippet would have failed under the new validator. Added `session_key_sig: ''` to both the register and revoke states, plus a `client.signChannelSessionKeyOwnership(state, sessionKeySigner)` call before submit in each (using `EthereumMsgSigner` directly, per the narrowed signature). Fixed in [c5810f3](https://github.com/layer-3/nitrolite/pull/739/changes/c5810f32c3fea1b81cc0816c4935a3e7fdd2724e)

nitronode/config/migrations/postgres/20260508000000_session_key_ownership_constraints.sql
philanton: Yeah, you're right — I had my deploy ordering wrong in the earlier reply. The pre-MF-H02 binary *is* MF-H01, which already creates `current_session_key_states_v1` and writes pointers, so an old pod during the MF-H02 rollout can absolutely land a history row with `NULL` `session_key_sig`, and the new lookups would trust it. Added `NOT VALID` check constraints on both `app_session_key_states_v1` and `channel_session_key_states_v1` requiring non-empty `session_key_sig`. `NOT VALID` skips the legacy backfill scan so pre-existing rows pass; only future inserts are checked, which is exactly the rollout window we care about. Fixed in [455e4ae](https://github.com/layer-3/nitrolite/pull/739/changes/455e4ae6b3ef08a791cd530502f41e2965d39bc9)

nitronode/store/database/current_session_key_state.go
philanton: Yep, good call — the version=0 row being a permanent ownership reservation rather than a transient seed was easy to miss. Extended the `LockSessionKeyState` docstring to spell this out: the seed *is* the (session_key, kind) ownership bind, cleanup-on-failure is intentionally absent, and `CountSessionKeysForUser` already filters version=0 so the cap math stays unaffected. Fixed in [b7287a7](https://github.com/layer-3/nitrolite/pull/739/changes/b7287a795c5c3b66ca9cd967fa43dc65b1c174d8)

nitronode/api/app_session_v1/submit_session_key_state.go
philanton: Yep, exactly the right move — the channel handler delegates in two lines while the app handler open-coded ~50. Added `app.ValidateAppSessionKeyStateV1` mirroring `core.ValidateChannelSessionKeyStateV1`, reduced the handler to a single delegating call, and ported the same set of unit tests (good sigs, empty session_key_sig, wrong user/key, tampered version, cross-wallet/cross-key replay). Left the tx-body duplication (`LockSessionKeyState` + cap + version check) for a follow-up as you suggested. Fixed in [a409407](https://github.com/layer-3/nitrolite/pull/739/changes/a409407ec6267e9caf747f95b4b16c99c62a9a32)

nksazonov (approved):
Good job!

sdk/ts-compat/src/client.ts
```diff
@@ -859,6 +859,13 @@ export class NitroliteClient {
         return this.innerClient.signChannelSessionKeyState(state);
     }
 
+    async signChannelSessionKeyOwnership(
+        state: ChannelSessionKeyStateV1,
+        sessionKeySigner: StateSigner,
```
ihsraham: i would fix this before approval. the inner sdk now requires `EthereumMsgSigner`, but compat still accepts `StateSigner`, so `npm run typecheck` fails here and callers can be pointed at a signer shape the node rejects. can we mirror the main sdk type here and in the app helper below?

sdk/ts/README.md
```diff
@@ -492,27 +494,26 @@ const sessionKeySigner = new AppSessionKeySignerV1(sessionKeyMsgSigner);
...
 ### App Session Keys
 
+Registration requires two signatures: the wallet's `user_sig` (authorizing the
+delegation) and the session-key holder's `session_key_sig` (proving possession of the key
+being registered). The node rejects submits that lack a valid `session_key_sig`.
+
 ```typescript
-// Sign and submit an app session key state
-const sig = await client.signSessionKeyState({
+// sessionKeyHolder is a StateSigner whose address equals state.session_key.
```
ihsraham: worth updating this example, and the channel one below, to create and pass an `EthereumMsgSigner`, not a generic `StateSigner`. the helper is intentionally narrowed to the raw message signer, so copying this snippet can lead to a type error or a rejected `session_key_sig`.

sdk/go/README.md
```diff
@@ -418,8 +420,12 @@ sig, _ := appSessionSigner.Sign(packedRequest)
 
 ### Session Keys — App Sessions
 
+Registration requires two signatures: the wallet's `UserSig` (authorizing the delegation)
+and the session-key holder's `SessionKeySig` (proving possession of the key being
+registered). The node rejects submits that lack a valid `SessionKeySig`.
+
 ```go
-// Sign and submit an app session key state
+// sessionKeySigner is a sign.Signer whose address equals state.SessionKey.
```
ihsraham: small docs mismatch: this should say `*sign.EthereumMsgSigner`, not any `sign.Signer`. the helper now requires the concrete raw message signer, so the example should not suggest a wrapped signer can be used.

ihsraham:
looks close on h-02. the server path now requires the session-key ownership proof and locks `(session_key, kind)` ownership, so the main finding looks addressed from the node side.

i left one approval-blocking compat note because `sdk/ts-compat` does not typecheck yet, plus two small docs notes so clients use the same raw signer shape.

sdk/ts-compat/src/client.ts
philanton: Yep, you're right — the inner SDK was narrowed but the compat wrapper kept the old `StateSigner` shape, so `npm run typecheck` was indeed failing in sdk-compat. Narrowed both `signChannelSessionKeyOwnership` and `signAppSessionKeyOwnership` to `EthereumMsgSigner`. Imported the type as `type EthereumMsgSigner` (type-only, so it's erased at compile time — safe wrt the SSR / no-barrel-re-export constraint documented in `src/index.ts`). Fixed in [5d3c59e](https://github.com/layer-3/nitrolite/pull/739/changes/5d3c59ec411f16d274db3ec71cefa5308e3549cf)

sdk/ts/README.md
philanton: Yep, fair — copying the snippet as written would trip the narrowed type. Updated both the app and channel examples to construct an `EthereumMsgSigner` directly (via `new EthereumMsgSigner(sessionKeyPrivateKey)`) and added a short why-comment so readers don't reach for a wrapped `StateSigner`. Fixed in [5d3c59e](https://github.com/layer-3/nitrolite/pull/739/changes/5d3c59ec411f16d274db3ec71cefa5308e3549cf)

sdk/go/README.md
philanton: Yep, same fix on the Go side — both the app and channel snippets now construct `*sign.EthereumMsgSigner` via `sign.NewEthereumMsgSigner(sessionKeyPrivHex)` instead of suggesting a generic `sign.Signer`, with the same why-comment about the node recovering `SessionKeySig` as a raw 65-byte EIP-191 sig. Fixed in [5d3c59e](https://github.com/layer-3/nitrolite/pull/739/changes/5d3c59ec411f16d274db3ec71cefa5308e3549cf)

ihsraham (approved):
approving for h-02. the node-side fix now binds session-key registration to key possession, the compat signer type and snapshot are aligned, and all review threads are resolved.

i would keep legacy rows without `session_key_sig` as an ops cleanup note, not a blocker for the new-write fix.

PR #744, opened 2026-05-11
MF-M04: feat: provide tooling for and enhance docs on ValidatorRegistered event

nksazonov:
## Issue

Security audit finding H-01: the `VALIDATOR_ACTIVATION_DELAY` mitigation is only effective if users actively monitor on-chain validator registrations. Previously there was no tooling to do so, and the documentation understated the attack surface (escrowed funds, not just undeposited approvals) and omitted the monitoring obligation entirely.

## Fix

- **Documentation** (`contracts/SECURITY.md`, `protocol-description.md`, `docs/protocol/security-and-limitations.md`): document the monitoring requirement, permanent nature of registered validators, ERC20 approval risk, and escrow exit guidance; add validator registry as a named trust assumption.
- **Go SDK** (`pkg/blockchain/evm/validator_watcher.go`, `sdk/go/validator_watcher.go`): add `WatchValidatorRegistered` — gap-free event delivery via subscribe-first + historical `FilterLogs` replay with duplicate dedup on the live phase.
- **TypeScript SDK** (`sdk/ts/src/blockchain/evm/validator_watcher.ts`, `sdk/ts/src/client.ts`): add `watchValidatorRegistered` — async generator with historical replay anchored at `headBlock+1` to close the getLogs-to-first-poll transition gap; 15 unit tests.


pkg/blockchain/evm/validator_watcher.go
```diff
@@ -0,0 +1,138 @@
...
+				return
+			case err := <-sub.Err():
+				// Suppress the error log on clean ctx cancellation: go-ethereum delivers
+				// a cancellation error on sub.Err() before ctx.Done() is scheduled,
+				// which would otherwise log a spurious "subscription error" on shutdown.
+				if err != nil && ctx.Err() == nil {
+					logger.Error("ValidatorRegistered subscription error", "error", err, "contract", contractAddress.Hex(), "blockchainID", blockchainID)
+				}
+				return
+			case l, ok := <-logCh:
```
philanton, line 107: Hmmm, I think there's a duplicate-event window here. `SubscribeFilterLogs` starts buffering live logs at T1, then `HeaderByNumber` returns block N at T2 > T1. The historical query then covers `[fromBlock, N]`, and the live channel already holds everything from T1 onward — so any block mined in the `[T1, T2]` interval lands in both streams and gets yielded twice.

The TS side avoids this by anchoring the live phase at `headBlock + 1n` (see the `liveFromBlock` calc in `validator_watcher.ts`). I'd flip the order here: fetch the header first, then call `SubscribeFilterLogs` with `FromBlock: header.Number + 1`, or alternatively dedupe by `(BlockNumber, LogIndex)` before delivery.

sdk/go/examples/validator_watcher/main.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	defer client.Close()
+
+	fmt.Printf("Monitoring ValidatorRegistered events on chain %d...\n", chainID)
+	fmt.Println("Press Ctrl+C to stop.")
+
+	// fromBlock tracks where to resume on reconnect. Zero means "start from latest"
+	// on the first call; subsequent reconnects pass lastBlock+1 to replay any events
+	// emitted while the subscription was down.
+	var fromBlock uint64
```
philanton, line 86: Initial-subscription gap here: `fromBlock` starts at 0, so if the subscription dies before any event has arrived, the reconnect call also passes 0 → historical fetch is skipped → events emitted during the outage are lost.

I'd fetch the current head once on first connect and use that as the anchor (`fromBlock = head + 1`). Same issue applies to the TS example pattern in the `client.ts` JSDoc.

contracts/SECURITY.md
```diff
@@ -315,6 +315,20 @@ The registration is an observable on-chain event, and with monitoring in place,
...
+
+2. **The 1-day window is the entire response budget.** After `VALIDATOR_ACTIVATION_DELAY` elapses, the validator is active and any outstanding ERC20 approvals to ChannelHub are immediately exploitable via a forged `createChannel`.
+
+3. **Users must subscribe to `ValidatorRegistered` events.** Monitor the `ValidatorRegistered(uint8 indexed validatorId, address indexed validator)` event on the ChannelHub contract. Any unexpected registration should be treated as a potential compromise. Revoke all ERC20 approvals granted to ChannelHub immediately if an unrecognised validator appears.
+
+4. **Avoid large standing ERC20 approvals.** Do not grant unlimited (`type(uint256).max`) or long-lived ERC20 allowances to ChannelHub. Prefer exact-amount approvals per operation. A standing approval only becomes exploitable once a malicious validator activates, so minimising the approved amount caps the worst-case loss.
+
+The TypeScript and Go SDKs expose validator-monitoring helpers that subscribe to `ValidatorRegistered` events and surface alerts when new validators are registered, making this surveillance practical for all integrators.
```
philanton: This sentence reads like the two SDKs are at parity, but the implementations diverge:
1) the Go watcher requires a WebSocket/IPC endpoint — `SubscribeFilterLogs` doesn't poll and the new code explicitly errors out otherwise;
2) the TS watcher works over HTTP via viem polling.

I'd either call the difference out here (e.g. "Go SDK requires a WS endpoint; TS SDK falls back to HTTP polling"), or add a polling fallback to Go for true parity.

contracts/SECURITY.md
```diff
@@ -315,6 +315,20 @@ The registration is an observable on-chain event, and with monitoring in place,
...
 
+**User responsibilities and monitoring requirement:**
+
+The `VALIDATOR_ACTIVATION_DELAY` is only effective if users actively monitor on-chain validator registrations within the 1-day window. The delay provides no protection for users who are not watching. Key points every user must understand:
+
+1. **Validators are permanent.** Once a validator is registered at a given ID, it cannot be deactivated or overwritten. If a malicious validator becomes active, it remains active for the lifetime of the ChannelHub deployment. There is no on-chain recovery mechanism after the delay expires.
+
+2. **The 1-day window is the entire response budget.** After `VALIDATOR_ACTIVATION_DELAY` elapses, the validator is active and any outstanding ERC20 approvals to ChannelHub are immediately exploitable via a forged `createChannel`.
+
+3. **Users must subscribe to `ValidatorRegistered` events.** Monitor the `ValidatorRegistered(uint8 indexed validatorId, address indexed validator)` event on the ChannelHub contract. Any unexpected registration should be treated as a potential compromise. Revoke all ERC20 approvals granted to ChannelHub immediately if an unrecognised validator appears.
```
philanton: Small thing — "revoke all ERC20 approvals" only protects undeposited funds. Funds already in escrow stay exposed via forged user-signatures (a malicious validator can sign any user-sig op, not just `createChannel`), so this advice understates the surface a bit.

I'd add a line about exiting via `initiateEscrowWithdrawal` within the window for already-escrowed positions, assuming that exit path doesn't itself depend on the rogue validator.

protocol-description.md
```diff
@@ -694,7 +694,7 @@ The current implementation binds each ChannelHub to a single node address at dep
...
 
-A contract-enforced `VALIDATOR_ACTIVATION_DELAY` (1 day) provides a partial, targeted defence within this trust boundary: a newly registered validator cannot be used until the delay has elapsed, creating an observable window during which a key compromise can be detected and users can revoke ERC20 approvals before the attack on undeposited funds can execute.
+A contract-enforced `VALIDATOR_ACTIVATION_DELAY` (1 day) provides a partial, targeted defence within this trust boundary: a newly registered validator cannot be used until the delay has elapsed, creating an observable window during which a key compromise can be detected and users can revoke ERC20 approvals before the attack on undeposited funds can execute. **This window is only effective if users actively monitor the `ValidatorRegistered` event on the ChannelHub contract.** Validators are permanent once registered — there is no deactivation mechanism. Users should subscribe to validator-registration events and avoid granting large standing ERC20 approvals to ChannelHub; see `contracts/SECURITY.md` for concrete guidance.
```
philanton: Small tone clash in this paragraph: the original sentence still calls the delay "a partial, targeted defence", then the added text says the window is "only effective if users actively monitor". Those two readings pull in different directions.

I'd rephrase the first half so they line up — something like "a partial, targeted defence, effective only when users actively monitor on-chain registrations within the window".

sdk/go/examples/validator_watcher/main.go
```diff
@@ -0,0 +1,121 @@
...
+// handleValidatorRegistered is called for every ValidatorRegistered event.
+// Replace with your own alerting, logging, or user-notification logic.
+func handleValidatorRegistered(ev *core.ValidatorRegisteredEvent) {
+	fmt.Printf(
+		"[ALERT] New validator registered on chain %d at block %d: ID=%d address=%s\n",
+		ev.BlockchainID, ev.BlockNumber, ev.ValidatorID, ev.Validator,
+	)
+	fmt.Println("Action required: verify this validator is expected.")
+	fmt.Println("If unexpected, revoke all ERC20 approvals to the ChannelHub contract immediately.")
+	fmt.Println("You have 1 day (VALIDATOR_ACTIVATION_DELAY) before the validator becomes active.")
```
philanton, line 132: This handler alerts on every registration — including the legitimate first-time bootstrap one. App builders who paste this in verbatim will spam users during normal operations.

I'd show an allowlist pattern here (e.g. a set of expected validator IDs/addresses loaded from config) and only fire the alert when an unknown one shows up. That matches the spirit of the SECURITY.md guidance more closely.

philanton (commented):
A few things I noticed while reading through — left them inline.

ihsraham:
my read is that this closes M-04 as an accepted trust assumption with monitoring, not as the onchain per-user validator approval mapping from the finding.

could we state that in the pr notes, and say the watcher path is the intended closure route? the idea is to keep the audit record clear, so this is not later read as a protocol-level fix.

sdk/go/examples/validator_watcher/main.go
nksazonov: This is just an example usage of the watcher. However, you are right about the confusion its direct usage can bring. Thus, I will clarify in the comments on the lines referenced that the watcher also fires on legitimate signature validator registrations, and that to differentiate the malicious ones, app builders and users should follow the official Nitrolite communication channels.

nksazonov: a513b832f322cd00c2dbad95d65e8c6afae541de

protocol-description.md
nksazonov: a513b832f322cd00c2dbad95d65e8c6afae541de

contracts/SECURITY.md
nksazonov: a513b832f322cd00c2dbad95d65e8c6afae541de

contracts/SECURITY.md
nksazonov: a513b832f322cd00c2dbad95d65e8c6afae541de

sdk/go/examples/validator_watcher/main.go
nksazonov: a513b832f322cd00c2dbad95d65e8c6afae541de

pkg/blockchain/evm/validator_watcher.go
nksazonov: a513b832f322cd00c2dbad95d65e8c6afae541de

philanton (approved):

ihsraham (approved):

PR #745, opened 2026-05-12
MF-M08: fix(rpc): replace Origin label with application_id on connection gauge

philanton:
## Summary
- Replace the client-controlled `Origin` header label on `rpc_connections_active` with `application_id`, sourced from the `app_id` query parameter validated at WebSocket upgrade. Unauthenticated remotes can no longer drive unbounded Prometheus label cardinality by sending unique `Origin` headers.
- Track per-application connection counts inside `ConnectionHub` and call `DeleteLabelValues` once a bucket reaches zero, so series for disconnected applications are shed instead of lingering as zero-valued gauges.
- Expose `ApplicationID()` on the `rpc.Connection` interface (alongside the existing `Origin()`) so the hub can label without re-parsing the request. The raw `Origin` field is retained for logging only.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/rpc/... ./nitronode/metrics/...`
- [ ] Verify `rpc_connections_active{application_id="..."}` on a running node and that the series disappears after the last connection for that app closes.

## Release Notes

* **New Features**
  * Session key pagination with 10-item page limit for improved query performance.
  * Per-user session key limits (default 100) to prevent abuse.
  * WebSocket DoS hardening with inbound frame size caps and per-connection byte-rate limiting.
  * Protocol drift validation in CI ensures RPC, ABI, and signing consistency across code layers.

* **Documentation**
  * Expanded token compatibility requirements; tokens without ERC-20 `decimals()` are rejected.
  * Clarified on-chain signature domain behavior (EIP-191 vs raw ECDSA).
  * Protocol drift guardrails and trust assumptions documented.

[Review Change Stack](https://app.coderabbit.ai/change-stack/layer-3/nitrolite/pull/745)

### DIFF
--- a/nitronode/metrics/exporter.go
+++ b/nitronode/metrics/exporter.go
@@ -289,11 +289,11 @@ func NewRuntimeMetricExporter(reg prometheus.Registerer) (RuntimeMetricExporter,
 		rpcConnectionsTotal: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Namespace: MetricNamespace,
 			Name:      "rpc_connections_active",
-			Help: "Active RPC (WebSocket) connections. Labels: region, origin " +
-				"(both sourced from request headers / client metadata at connect " +
-				"time). Operator-decided values — bounded only by the set of " +
-				"connecting clients.",
-		}, []string{"region", "origin"}),
+			Help: "Active RPC (WebSocket) connections, labeled by application_id " +
+				"sourced from the app_id query parameter at connect time. " +
+				"Connections without an app_id are bucketed under _DEFAULT. " +
+				"Series for an application_id are deleted once its count drops to 0.",
+		}, []string{"application_id"}),
 		rpcInflight: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Namespace: MetricNamespace,
 			Name:      "rpc_inflight",
@@ -412,8 +412,15 @@ func (m *runtimeMetricExporter) ObserveRPCDuration(method, path string, success 
 	m.rpcRequestDurationSeconds.WithLabelValues(method, path, result.String()).Observe(duration.Seconds())
 }
 
-func (m *runtimeMetricExporter) SetRPCConnections(region, origin string, count uint32) {
-	m.rpcConnectionsTotal.WithLabelValues(region, origin).Set(float64(count))
+func (m *runtimeMetricExporter) SetRPCConnections(applicationID string, count uint32) {
+	label := getApplicationIDLabelValue(applicationID)
+	if count == 0 {
+		// Shed the series when the bucket empties so unique application_id values
+		// from clients cannot accumulate unbounded gauge labels over time.
+		m.rpcConnectionsTotal.DeleteLabelValues(label)
+		return
+	}
+	m.rpcConnectionsTotal.WithLabelValues(label).Set(float64(count))
 }
 
 func (m *runtimeMetricExporter) IncRPCInflight(method string) {
--- a/nitronode/metrics/exporter.go
+++ b/nitronode/metrics/exporter.go
@@ -292,7 +292,11 @@ func NewRuntimeMetricExporter(reg prometheus.Registerer) (RuntimeMetricExporter,
 			Help: "Active RPC (WebSocket) connections, labeled by application_id " +
 				"sourced from the app_id query parameter at connect time. " +
 				"Connections without an app_id are bucketed under _DEFAULT. " +
-				"Series for an application_id are deleted once its count drops to 0.",
+				"Series for an application_id are deleted once its count drops to 0. " +
+				"NOTE: cardinality is bounded by the app_id format check " +
+				"(^[a-z0-9_-]{1,66}$) and by series shedding on disconnect — long-lived " +
+				"connections from many distinct but format-valid app_ids are not gated " +
+				"by a registry or per-app connection cap.",
 		}, []string{"application_id"}),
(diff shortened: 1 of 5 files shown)


pkg/rpc/connection_hub.go
```diff
@@ -69,9 +70,9 @@ func (hub *ConnectionHub) Add(conn Connection) error {
 
 	hub.connections[connID] = conn
 
-	sourceID := getSourceID(conn.Origin())
-	hub.sourceMap[sourceID]++
-	hub.observeConnections(defaultConnectionRegion, conn.Origin(), uint32(hub.sourceMap[sourceID]))
+	appID := conn.ApplicationID()
+	hub.appConnCount[appID]++
+	hub.observeConnections(appID, hub.appConnCount[appID])
```
nksazonov: The `observeConnections` callback is called while `hub.mu.Lock()` is still held (via `defer`). It resolves to `SetRPCConnections` which acquires prometheus-internal mutexes (`DeleteLabelValues` / `WithLabelValues`). Under connection churn, all hub readers — including `Publish` for message routing — are serialized behind Prometheus writes.

Suggest capturing `appID` and `count` before the lock, dropping the `defer`, and calling `hub.observeConnections(appID, count)` after an explicit `hub.mu.Unlock()`. Same pattern applies to line 122 in `Remove`.

pkg/rpc/connection_hub.go
```diff
@@ -111,14 +112,14 @@ func (hub *ConnectionHub) Remove(connID string) {
...
-			delete(hub.sourceMap, sourceID)
+	appID := conn.ApplicationID()
+	if count, exists := hub.appConnCount[appID]; exists && count > 0 {
+		hub.appConnCount[appID]--
+		if hub.appConnCount[appID] == 0 {
+			delete(hub.appConnCount, appID)
 		}
 	}
-	hub.observeConnections(defaultConnectionRegion, conn.Origin(), uint32(hub.sourceMap[sourceID]))
+	hub.observeConnections(appID, hub.appConnCount[appID])
```
nksazonov: `observeConnections` fires unconditionally after the guard, even when the `exists && count > 0` branch did not execute (key absent or already-zero). In that case `hub.appConnCount[appID]` returns the Go zero value, triggering `DeleteLabelValues` for an `app_id` the gauge never tracked.

Suggest moving `hub.observeConnections(...)` inside the `if` block so the callback only fires when state actually changed.

nitronode/metrics/exporter.go
```diff
@@ -412,8 +412,15 @@ func (m *runtimeMetricExporter) ObserveRPCDuration(method, path string, success
...
-	m.rpcConnectionsTotal.WithLabelValues(region, origin).Set(float64(count))
+func (m *runtimeMetricExporter) SetRPCConnections(applicationID string, count uint32) {
+	label := getApplicationIDLabelValue(applicationID)
+	if count == 0 {
+		// Shed the series when the bucket empties so unique application_id values
+		// from clients cannot accumulate unbounded gauge labels over time.
+		m.rpcConnectionsTotal.DeleteLabelValues(label)
+		return
+	}
+	m.rpcConnectionsTotal.WithLabelValues(label).Set(float64(count))
```
nksazonov, line 427: `IsValidApplicationID` validates format only (`^[a-z0-9_-]{1,66}$`), not registration. A client that opens one persistent WebSocket per distinct valid `app_id` and keeps all connections alive will never trigger `DeleteLabelValues` for those series — the cardinality fix is complete for disconnected clients but does not bound live-series count under a persistent-connection flood.

Whether a per-`app_id` connection cap or registry check is needed is a product decision. Worth at minimum a note in the gauge `Help` string acknowledging where the remaining bound lies.

nksazonov (commented):
Clean approach to bounding metric cardinality. Two things need attention before this lands.

**Field naming:** `rpcConnectionsTotal *prometheus.GaugeVec` (line 180 of `exporter.go`) names a gauge as if it were a counter — minor but confusing. Consider renaming to `rpcConnectionsActive` alongside this change.

pkg/rpc/connection_hub.go
philanton: Yep, good catch. Fixed in [885bbfc](https://github.com/layer-3/nitrolite/pull/745/changes/885bbfcbe360d0636f2fd7d8f7260af511b2666f) — snapshot the count, drop `hub.mu`, then call `observeConnections` outside the lock. Same pattern applied to `Remove`.

pkg/rpc/connection_hub.go
philanton: Right, that was emitting a `DeleteLabelValues` for buckets the gauge never tracked. Fixed in [885bbfc](https://github.com/layer-3/nitrolite/pull/745/changes/885bbfcbe360d0636f2fd7d8f7260af511b2666f) — observer only fires now when the bucket actually changed.

nitronode/metrics/exporter.go
philanton: Yeah, you're right — format check alone doesn't bound live-series under a persistent-connection flood. Added a caveat to the gauge `Help` in [885bbfc](https://github.com/layer-3/nitrolite/pull/745/changes/885bbfcbe360d0636f2fd7d8f7260af511b2666f) calling out that the bound rests on the format regex plus disconnect-shedding — no registry / per-app cap. Agree the cap itself is a separate product call.

nksazonov (approved):
Well done!

PR #746, opened 2026-05-12
MF-C02: fix(core): add ChannelStatusClosing to gate post-finalize state transitions

nksazonov:
## Problem                                                                                                                                
                                                                                                                                          
MF-C02 identified an epoch-rebinding vulnerability. A user could checkpoint channel X at version 1, submit an off-chain `Finalize`, and then — before the on-chain close event arrived — submit a fake `HomeDeposit` referencing the old channel ID. Because the DB still showed channel X as `Open`, `SubmitState` accepted it. `EnsureNoOngoingStateTransitions` then saw `fake_deposit.version == channel_X.state_version` (both `1`) and treated the deposit as settled on-chain. The attacker could spend the fake balance, draining the node vault.                                   

The necessary pre-condition for the attack is that an active channel exists according to the DB **after** the off-chain Finalize has been co-signed. Closing that window is sufficient to block the attack.
                                                                                                                                            
  ---                                                            
                                                
  ## Fix

  ### `ChannelStatusClosing` (new status)

A new status `Closing = 3` is inserted between `Challenged (2)` and `Closed (4)`. When `SubmitState` stores a co-signed `Finalize` state, it atomically updates the channel status to `Closing` in the same DB transaction. Both writes succeed or both roll back — there is no window between them.                                                                                                                      
                                                                 
`CheckActiveChannel` (the renamed and extended `CheckOpenChannel`) queries `status <= Open (1)`. Since `Closing = 3 > 1`, a channel in `Closing` state is no longer considered active. `SubmitState` receives a nil status pointer and returns "user has no active channel" before `ValidateAdvancement` is ever reached. The fake `HomeDeposit` never gets signed.                                                   
                                                                 
  ### `RequestCreation` gate                    

`RequestCreation` now checks `HasNonClosedHomeChannel` before accepting a new channel creation request. If any home channel for the user/asset pair has status other than `Closed`, the request is rejected. This seals the epoch-rebinding vector: the attacker cannot open a new channel in the new epoch while the old one is still pending on-chain settlement, so there is no channel lifecycle to rebind to.      
                                                                 
  ### Why R2 and R3 are sealed as a consequence                                                                                             
   
  Two contributing factors identified in the analysis become unreachable once the gate is in place:                                         
                                                                 
  - **R2** (`EnsureNoOngoingStateTransitions` epoch-blind version check) — only matters if a fake signed `HomeDeposit` exists in the DB. Because `SubmitState` is blocked before signing, no such state can be stored.
  - **R3** (node-internal `issueTransferReceiverState` / `issueReleaseReceiverState` bypassing `ValidateAdvancement`) — these issue states onto nil-channel states. The node does not sign nil-channel states, so they cannot be used to claim funds or satisfy the signed-state check in `EnsureNoOngoingStateTransitions`. Both are structurally latent but not exploitable via this attack path.
                                                                                                                                            
  ### `CheckActiveChannel` (renamed from `CheckOpenChannel`)     
                                                
  The function now returns `(approvedSigValidators string, status *core.ChannelStatus, error)`. The status pointer is `nil` when no active channel exists. This allows `SubmitState` to additionally enforce that `MutualLock` and `EscrowLock` transitions require the channel to be materialized on-chain (`status == Open`), not merely registered in the DB (`status == Void`).


philanton (commented):
After tracing the attack step-by-step, I'd say MF-C02's root cause is narrower than the R1/R2/R3 framing in the PR body. The fix can be much smaller, and I think it's worth considering before we land this surgery on `ValidateAdvancement`.

The attack only works because `CheckOpenChannel` accepts a channel that's been finalized off-chain but not yet closed on-chain. That window is the actual root cause — once a `Finalize` is co-signed, the channel has no business accepting new user-initiated state transitions, but the DB row still says `Open`.

### Proposed remediation

Add `ChannelStatusClosing` to the enum:

```go
ChannelStatusClosing ChannelStatus = 4
```

State transitions:
- `Open → Closing` on doubly-signed `Finalize` — must be **atomic** with the state-store write, otherwise we leave a window open;
- `Closing → Closed` on observed on-chain `Close` event;
- `Closing → Challenged` if a challenge can race a co-signed Finalize (need to confirm whether the protocol allows that).

Then two gates:

1) `CheckOpenChannel` returns true only for `status == Open`. Apply at every user-input entry that feeds `ValidateAdvancement` — at minimum `channel_v1.SubmitState` and `app_session_v1.SubmitDepositState`;
2) `RequestCreation` rejects if any row for `(wallet, asset)` has `status != Closed`. Enforces \"one in-flight channel per asset\" — simpler invariant to audit than allowing `Closing` and a new `Open` to coexist.

Trace against the attack from the PR body:
- step 2: off-chain `Finalize` → status flips to `Closing`;
- step 3: passive `TransferReceive` issuance — still fine, doesn't depend on channel status;
- step 4: `SubmitState` with fake `HomeDeposit` → `CheckOpenChannel` rejects → dies before reaching `ValidateAdvancement`.

The variant where step 3 is skipped (fake `HomeDeposit` directly after `Finalize`) dies at the same gate.

### About R2 and R3

I'd say R2 and R3 aren't actually root causes of MF-C02 — they're contributing factors. Once the `CheckOpenChannel` gate is in place:

- **R2** (`EnsureNoOngoingStateTransitions` epoch-blind): the SQL is still buggy, but R2 only gets exercised adversarially once a doubly-signed fake state exists. The gate prevents that state from being minted. **R2 becomes unreachable from user input.**
- **R3** (passive paths bypass `ValidateAdvancement`): the passive paths still skip the validator, but every user-input consumer that could exploit a malformed passive state is now gated upstream. The adoption block in `ValidateAdvancement` becomes **unreachable** from outside.

So my recommendation is: document this in a security note alongside the gate (\"R2 and R3 are sealed by the `Closing` gate — latent but unreachable from user input\"), track them as follow-up hardening items, and **don't bundle them into the MF-C02 fix**. Smaller PR, cleaner audit story, easier to bisect/revert. If we ever need to refactor that area later and it makes sense to also fix R2/R3 properly, we can do it then.

### What I'd drop from this PR

If we go this route:
1) the `ValidateAdvancement` signature change (`State` → `*State`);
2) `validateIsolatedPassiveState`;
3) the `validationBase`/`stateBase` split in `issueTransferReceiverState` / `issueReleaseReceiverState`;
4) `effectiveBase` synthesis in `RequestCreation`.

The adoption block in `ValidateAdvancement` can stay as-is — it becomes dead code defensively, but it's still load-bearing for the legitimate `RequestCreation` flow (where the new channel ID is supposed to be adopted onto a finalized/nil-HCID base).

### Caveats to watch

- **Atomicity** of the `Open → Closing` flip with the `Finalize` store — must be the same DB tx. Any window where `Finalize` is signed but status still `Open` re-opens the exploit.
- **`Void` lifecycle** — make sure `Void` rows get reaped/transitioned on tx failure, otherwise users get locked out of new channel creation.
- **`CheckOpenChannel` callers** — not all of them want \"is the channel still alive in any form\" semantics (e.g. read-only / balance queries may legitimately want to see `Closing` too). Only tighten the predicates used by paths that feed `ValidateAdvancement`.

Happy to walk through this in a sync if it helps:)

philanton (commented):
Took another pass on `c7c0e4f3`. Direction looks good — the new approach matches what we discussed, and a few of the implementation details are nice:

- Inserting `Closing = 3` between `Challenged` and `Closed` keeps the `status <= ChannelStatusOpen` predicate intact, so every existing `CheckOpenChannel`/`CheckActiveChannel` caller automatically rejects `Closing` without changes. Clean.
- Finalize co-sign and the `Open → Closing` flip are inside the same `useStoreInTx` — atomicity holds.
- `RequestCreation` `HasNonClosedHomeChannel` gate with a dedicated test (`TestRequestCreation_NonClosedChannelRejection`) is exactly the regression we wanted.
- `Closing` strings/enums propagated to TS SDK, ts-compat, and `docs/api.yaml`. No drift.

A few things I'd add/check before merging:

### Test gaps

1) The attack-replay test is missing. Would be good to have a `TestSubmitState_*` case that reproduces MF-C02 step 4 directly: channel in `Closing`, user calls `SubmitState` with a fake `HomeDeposit` referencing the old channel ID → `CheckActiveChannel` returns `nil` status → rejected with `\"user has no active channel\"`. Same for `SubmitDepositState`. This is the regression test the audit reviewer will look for.
2) `TestDBStore_CheckActiveChannel` only covers Open / Closed / no-channel / wrong-asset. I'd add explicit cases for `Closing` and `Challenged` returning nil status — pins down the `status <= ChannelStatusOpen` invariant against future enum reordering.

### Finalize on Void

`SubmitState` already requires `*channelStatus == ChannelStatusOpen` for `MutualLock` / `EscrowLock` (line 76). `Finalize` doesn't have the same check — so technically a user could `Finalize` a `Void` channel (DB row exists but channel hasn't been materialized on-chain yet). Probably nonsensical at the protocol level, but it does flip the row to `Closing` and lock further submissions. I'd add the same `if *channelStatus != ChannelStatusOpen { return ... }` guard for `Finalize` to keep the state machine tight.

### Migration

1) No down migration. If we ever need to revert, `UPDATE channels SET status = 3 WHERE status = 4` would do it — but only if there are no `Closing(3)` rows by then, otherwise they'd collide with the old `Closed(3)` semantics. Worth either including a down migration that handles both cases or documenting the revert procedure in `MIGRATION-NITRONODE.md`.
2) Rolling-deploy race: if any node instance runs new code before the migration completes, it'll read legacy `status=3` (historical Closed) as `ChannelStatusClosing` and treat truly-closed channels as in-flight. Need to make sure the deploy pipeline runs migrations strictly before the new binaries come up. Probably already the case — just worth confirming.
3) Migration filename `20260520000000` is dated 8 days in the future relative to today. Minor, but the rest of the migrations use chronological dates — might cause confusion when scanning history.

### Challenge during Closing

`HandleChannelChallenged` sets `Status = Challenged` unconditionally. If a challenge lands during `Closing`, the row goes `Closing → Challenged` and then eventually `Challenged → Closed` via `HandleHomeChannelClosed`. Logically OK (the co-signed Finalize becomes irrelevant once the chain takes over), but I'd add either a test or a short comment in `HandleChannelChallenged` documenting that this transition is expected, so the intent is captured.

### Nice to have (non-blocking)

The PR body's framing of \"R2 and R3 sealed as a consequence\" is well-stated — would be worth dropping a one-liner in `pkg/core/state_advancer.go` near the adoption block too, something like `// Reachable only via RequestCreation; user-input paths are gated by ChannelStatusClosing in SubmitState/SubmitDepositState.` so a future reader doesn't reintroduce the adoption-bypass concern.

Other than these, looks good to me. Most of it is test coverage and migration ops hygiene, not correctness issues with the core fix:)

nksazonov:
@philanton addressed most of the points from your second review:

**Fixed:**
- **Finalize on Void** — added `if *channelStatus != ChannelStatusOpen` guard to the `Finalize` case in `SubmitState`, matching the existing `MutualLock`/`EscrowLock` pattern
- **Attack-replay test** — added `TestSubmitState_ClosingChannel_Rejected`: channel in Closing state, user submits `HomeDeposit` → `CheckActiveChannel` returns nil → rejected with "user has no active channel"
- **`TestDBStore_CheckActiveChannel` gaps** — added `Closing` and `Challenged` sub-cases pinning the `status <= ChannelStatusOpen` invariant
- **Migration filename** — renamed `20260520000000` → `20260513000000`
- **Challenge during Closing** — added comment in `HandleHomeChannelChallenged` documenting that `Closing → Challenged` is expected, plus `TestHandleHomeChannelChallenged_FromClosingState`
- **Code comment references** — removed internal ticket IDs from test comments, replaced with short problem descriptions

**Not addressed:**
- **Down migration / rolling-deploy race** (points 4–5) — deferred, no active deployment yet

nitronode/api/channel_v1/submit_state.go
```diff
@@ -235,10 +235,26 @@ func (h *Handler) SubmitState(c *rpc.Context) {
...
+				// Atomically mark the channel as Closing so no further user-initiated state
+				// transitions are accepted until the on-chain close event is confirmed.
+				channel, err := tx.GetChannelByID(*incomingState.HomeChannelID)
+				if err != nil {
+					return rpc.Errorf("failed to get channel for finalize: %v", err)
+				}
+				if channel == nil {
+					return rpc.Errorf("channel not found for finalize: %s", *incomingState.HomeChannelID)
+				}
+				channel.Status = core.ChannelStatusClosing
```
ihsraham, line 254: I would block on this path because setting `Closing` here also hides the channel from `GetHomeChannel`: that read path calls `GetActiveHomeChannel`, which filters to `status <= Open`. The SDK `checkpoint()` flow fetches the home channel before it calls the on-chain `close`, so `CloseHomeChannel -> Checkpoint` can fail with `channel_not_found` after this state is signed.

Could we keep the signing gate strict but allow the close/checkpoint read path to fetch a `Closing` home channel, and add a regression for the normal close flow?

nitronode/config/migrations/postgres/20260513000000_add_channel_status_closing.sql
```diff
@@ -0,0 +1,12 @@
+-- Introduce ChannelStatusClosing (3) between Challenged (2) and Closed (4).
```
ihsraham, line 3: This file still needs the Goose direction annotation. The node applies embedded Postgres migrations with `goose.Up`, and Goose SQL files must start with `-- +goose Up`; without it, this migration can fail before the `status = 3 -> 4` rewrite runs.

Please add the annotation, and ideally a down or revert note if rollback is supported, before merge.

pkg/core/types.go
```diff
@@ -30,7 +30,8 @@ var (
 	ChannelStatusVoid       ChannelStatus = 0
 	ChannelStatusOpen       ChannelStatus = 1
 	ChannelStatusChallenged ChannelStatus = 2
-	ChannelStatusClosed     ChannelStatus = 3
+	ChannelStatusClosing    ChannelStatus = 3 // co-signed Finalize stored off-chain; on-chain close pending
```
ihsraham, line 33: Reusing the old `Closed` numeric value for `Closing` is risky during rolling deploy or rollback. An old process can still write `Closed` as `3`, while new code reads `3` as `Closing`, and then `HasNonClosedHomeChannel` blocks the next channel.

Can we either append `Closing` at a new value, or spell out and enforce a stop-the-world migration/deploy sequence that prevents old binaries from writing after the remap?

sdk/ts-compat/src/types.ts
```diff
@@ -37,6 +37,7 @@ export enum RPCChannelStatus {
     Closed = 'closed',
     Resizing = 'resizing',
     Challenged = 'challenged',
+    Closing = 'closing',
```
ihsraham, line 40: Adding `Closing` here also needs the compat runtime mapper updated. `sdk/ts-compat/src/client.ts` still maps numeric `3` to `closed` and has no `4`, so compat clients will see `closing` as `closed` and real closed channels as `"4"`.

Please update `STATUS_MAP` and add a small compat test for both `3 -> closing` and `4 -> closed`.

ihsraham (changes requested):
I would keep this in fix review for now. The `Closing` gate is the right direction, and the new tests help, but I still see a few blockers before I would call MF-C02 closed.

Main points are inline: the close/checkpoint path needs to keep working after a finalize is co-signed, the Goose migration needs to be valid, the `Closed` value reuse needs a rollout-safe plan, and ts-compat needs the runtime status mapping updated. I also still want one explicit `SubmitState` invariant or regression that proves post-final or nil-channel advancement cannot be signed if the status gate is bypassed by legacy data or migration drift.

pkg/core/types.go
nksazonov: This update is safe as the only environment we have right now is stress test, which we will wipe out and re-release. Sandbox and Production are offline.

philanton (approved):

nitronode/config/migrations/postgres/20260513000000_add_channel_status_closing.sql
nksazonov: Fixed in `dda977e7` — added `-- +goose Up` at the top and a `-- +goose Down` section that shifts `status=4` back to `status=3`, with a note that any `Closing=3` rows at rollback time have no valid representation in the old schema and must be resolved manually.

sdk/ts-compat/src/types.ts
nksazonov: Fixed in `dda977e7` — `STATUS_MAP` updated to `3: 'closing'` and `4: 'closed'`. Added a 5-case parametrized test in `client-mapping.test.ts` covering all values (0–4), and updated the public-API snapshot.

nitronode/api/channel_v1/submit_state.go
nksazonov: Addressed differently in `dda977e7`. Rather than widening `GetActiveHomeChannel` (which would break callers that rely on its strict `<= Open` semantic), we added `GetNotClosedHomeChannel` (`status != Closed`) and switched the `GetHomeChannel` handler to use it. `GetActiveHomeChannel` and `CheckActiveChannel` are unchanged. `TestGetHomeChannel_ClosingChannel` added as the regression: channel in `Closing` → `GetHomeChannel` returns it with status `"closing"` rather than `channel_not_found`.

ihsraham (approved):
Approved for MF-C02 closure on the current head. The close/checkpoint read path, Goose migration annotations, and ts-compat status mapping are fixed, and the regression coverage is in place.

The enum remap is still deployment-sensitive, but based on the stated rollout context (stress data will be wiped, sandbox/prod offline), I am treating that as a non-blocking release-procedure risk rather than a code blocker.

PR #747, opened 2026-05-12
YNU-917: reset stale store gauges and seed bounded counters

philanton:
## Summary
Remediates three issues with nitronode's Prometheus instrumentation that surfaced during the metrics audit:

- **Stale store gauges** — `app_sessions_total`, `channels_total`, `users_active`, `app_sessions_active`, `total_value_locked`, `node_balance`, `user_balance_*` are populated from `GROUP BY` queries that omit zero-count rows, so a label tuple whose last entity disappears (e.g., last open channel for an asset closes) keeps its prior non-zero value indefinitely. Added `Reset*` methods on `StoreMetricExporter` and call them inside each successful-query branch of the store-metrics worker before re-publishing the snapshot. `ResetActiveUsers` / `ResetActiveAppSessions` use `DeletePartialMatch(timespan=...)` so a failure in one time window does not blank the others. On query error the gauges are left untouched, so transient DB outages do not blank healthy series.
- **Cold-start gauge silence** — the store-metrics worker waited a full `fetchInterval` before its first refresh, so balance and channel gauges were absent during the warmup window. Tick body extracted into a `refresh` closure and invoked once before the `select` loop.
- **Cold-start counter silence** — counters whose label space is fully known at boot were absent until first traffic. Pre-publish 0-valued series for every (label tuple): `channel_state_sig_validations_total` (seeded in `NewRuntimeMetricExporter`), `rpc_messages_emitted_total` / `rpc_requests_total` / `rpc_request_duration_seconds` (seeded via `SeedRPCMethodMetrics` after route registration), and `blockchain_events_total` (seeded via `SeedBlockchainEventMetrics` after chain config load).

Counters carrying an unbounded `application_id` or `asset` axis are intentionally not seeded (out of scope for this story; tracked under the long-term cardinality concern).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/rpc/... ./nitronode/...`
- [ ] On a freshly started nitronode, `curl /metrics` and verify the seeded series (`channel_state_sig_validations_total`, `rpc_requests_total{...,path="default"}`, `blockchain_events_total`) appear with value 0 before any traffic or on-chain event arrives.
- [ ] Close the last `open` channel for an asset; verify `channels_total{asset="…",status="open"}` disappears after the next store-metrics tick (instead of sticking at its prior value).

## Summary by CodeRabbit

* **Chores**
  * Metrics now publish zero-valued series at startup for RPC methods, per-path RPC counters, channel signer validations, and blockchain events so dashboards/alerts show baseline series.
  * Store metric polling centralized; gauges are reset and repopulated only after successful queries to avoid stale label tuples.

* **New Features**
  * Nodes expose an API to list registered RPC methods.
  * Added a published mapping of RPC methods to allowed request paths.

* **Tests**
  * Expanded tests covering metric seeding/reset behavior and RPC method listing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/layer-3/nitrolite/pull/747?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/nitronode/main.go
+++ b/nitronode/main.go
@@ -355,85 +355,107 @@ func runStoreMetricsExporter(
 	ticker := time.NewTicker(fetchInterval)
 	defer ticker.Stop()
 
+	timeSpans := []struct {
+		label    string
+		duration time.Duration
+	}{
+		{"day", 24 * time.Hour},
+		{"week", 7 * 24 * time.Hour},
+		{"month", 30 * 24 * time.Hour},
+	}
+
+	// refresh runs one full pass over every store-backed gauge. Each block resets
+	// its gauge before re-publishing the latest snapshot so label tuples that have
+	// dropped out of the store (e.g., the last open channel for an asset closed)
+	// disappear from the metric instead of sticking at their previous value. Reset
+	// happens only inside the successful-query branch so a transient DB error does
+	// not blank a healthy gauge.
+	refresh := func() {
+		channelCounts, err := store.GetChannelsCountByLabels()
+		if err != nil {
+			logger.Error("failed to get channel counts by labels", "error", err)
+		} else {
+			metricExported.ResetChannels()
+			for _, c := range channelCounts {
+				metricExported.SetChannels(c.Asset, c.Status, c.Count)
+			}
+		}
+
+		appSessionCounts, err := store.GetAppSessionsCountByLabels()
+		if err != nil {
+			logger.Error("failed to get app sessions counts by labels", "error", err)
+		} else {
+			metricExported.ResetAppSessions()
+			for _, c := range appSessionCounts {
+				metricExported.SetAppSessions(c.Application, c.Status, c.Count)
+			}
+		}
+
+		tvlCounts, err := store.GetTotalValueLocked()
+		if err != nil {
+			logger.Error("failed to get total value locked", "error", err)
+		} else {
+			metricExported.ResetTotalValueLocked()
+			for _, c := range tvlCounts {
+				metricExported.SetTotalValueLocked(c.Domain, c.Asset, c.Value.InexactFloat64())
+			}
+		}
+
+		nodeBalances, err := store.GetNodeBalance()
+		if err != nil {
+			logger.Error("failed to get node balances", "error", err)
+		} else {
+			metricExported.ResetNodeBalance()
+			for _, nb := range nodeBalances {
+				metricExported.SetNodeBalance(nb.BlockchainID, nb.Asset, nb.Value.InexactFloat64())
+			}
+		}
+
+		offChain, err := store.GetUserBalanceSummary()
+		if err != nil {
+			logger.Error("failed to get off-chain liquidity", "error", err)
+		} else {
+			metricExported.ResetUserBalances()
+			for _, l := range offChain {
+				metricExported.SetUserBalanceTotal(l.BlockchainID, l.Asset, l.Total.InexactFloat64())
+				metricExported.SetUserBalanceUnderfunded(l.BlockchainID, l.Asset, l.Underfunded.InexactFloat64())
+				metricExported.SetUserBalanceReleasable(l.BlockchainID, l.Asset, l.Releasable.InexactFloat64())
+			}
+		}
+
+		for _, tw := range timeSpans {
+			counts, err := store.CountActiveUsers(tw.duration)
+			if err != nil {
+				logger.Error("failed to count active users", "timeframe", tw.label, "error", err)
+				continue
+			}
+			metricExported.ResetActiveUsers(tw.label)
+			for _, c := range counts {
+				metricExported.SetActiveUsers(c.Label, tw.label, c.Count)
+			}
+		}
+
+		for _, tw := range timeSpans {
+			counts, err := store.CountActiveAppSessions(tw.duration)
+			if err != nil {
+				logger.Error("failed to count active app sessions", "timeframe", tw.label, "error", err)
+				continue
+			}
+			metricExported.ResetActiveAppSessions(tw.label)
+			for _, c := range counts {
+				metricExported.SetActiveAppSessions(c.Label, tw.label, c.Count)
+			}
+		}
+	}
+
+	// Populate gauges immediately so they are non-empty before the first ticker fire
+	// (otherwise alerts using absent() trip during the cold-start window).
+	refresh()
+
 	for {
 		select {
 		case <-ticker.C:
-			timeSpans := []struct {
-				label    string
-				duration time.Duration
-			}{
-				{"day", 24 * time.Hour},
-				{"week", 7 * 24 * time.Hour},
-				{"month", 30 * 24 * time.Hour},
-			}
-
-			channelCounts, err := store.GetChannelsCountByLabels()
-			if err != nil {
-				logger.Error("failed to get channel counts by labels", "error", err)
-			} else {
-				for _, c := range channelCounts {
-					metricExported.SetChannels(c.Asset, c.Status, c.Count)
-				}
-			}
-
-			appSessionCounts, err := store.GetAppSessionsCountByLabels()
-			if err != nil {
-				logger.Error("failed to get app sessions counts by labels", "error", err)
-			} else {
-				for _, c := range appSessionCounts {
-					metricExported.SetAppSessions(c.Application, c.Status, c.Count)
-				}
-			}
-
-			tvlCounts, err := store.GetTotalValueLocked()
-			if err != nil {
-				logger.Error("failed to get total value locked", "error", err)
-			} else {
-				for _, c := range tvlCounts {
-					metricExported.SetTotalValueLocked(c.Domain, c.Asset, c.Value.InexactFloat64())
-				}
-			}
-
-			nodeBalances, err := store.GetNodeBalance()
-			if err != nil {
-				logger.Error("failed to get node balances", "error", err)
-			} else {
-				for _, nb := range nodeBalances {
-					metricExported.SetNodeBalance(nb.BlockchainID, nb.Asset, nb.Value.InexactFloat64())
-				}
-			}
-
-			offChain, err := store.GetUserBalanceSummary()
-			if err != nil {
-				logger.Error("failed to get off-chain liquidity", "error", err)
-			} else {
-				for _, l := range offChain {
-					metricExported.SetUserBalanceTotal(l.BlockchainID, l.Asset, l.Total.InexactFloat64())
-					metricExported.SetUserBalanceUnderfunded(l.BlockchainID, l.Asset, l.Underfunded.InexactFloat64())
-					metricExported.SetUserBalanceReleasable(l.BlockchainID, l.Asset, l.Releasable.InexactFloat64())
-				}
-			}
-
-			for _, tw := range timeSpans {
-				if counts, err := store.CountActiveUsers(tw.duration); err != nil {
-					logger.Error("failed to count active users", "timeframe", tw.label, "error", err)
-				} else {
-					for _, c := range counts {
-						metricExported.SetActiveUsers(c.Label, tw.label, c.Count)
-					}
-				}
-			}
-
-			for _, tw := range timeSpans {
-				if counts, err := store.CountActiveAppSessions(tw.duration); err != nil {
-					logger.Error("failed to count active app sessions", "timeframe", tw.label, "error", err)
-				} else {
-					for _, c := range counts {
-						metricExported.SetActiveAppSessions(c.Label, tw.label, c.Count)
-					}
-				}
-			}
-
+			refresh()
 		case <-ctx.Done():
 			logger.Info("stopping store metrics exporter")
 			return
--- a/nitronode/metrics/exporter.go
+++ b/nitronode/metrics/exporter.go
@@ -165,6 +165,27 @@ func (m *storeMetricExporter) SetUserBalanceReleasable(blockchainID, asset strin
 	m.userBalanceReleasable.WithLabelValues(blockchainID, asset).Set(value)
 }
 
+func (m *storeMetricExporter) ResetAppSessions()      { m.appSessionsTotal.Reset() }
+func (m *storeMetricExporter) ResetChannels()         { m.channelsTotal.Reset() }
+func (m *storeMetricExporter) ResetTotalValueLocked() { m.totalValueLocked.Reset() }
+func (m *storeMetricExporter) ResetNodeBalance()      { m.nodeBalance.Reset() }
+func (m *storeMetricExporter) ResetUserBalances() {
+	m.userBalanceTotal.Reset()
+	m.userBalanceUnderfunded.Reset()
+	m.userBalanceReleasable.Reset()
+}
+
+// ResetActiveUsers clears only the rows for the given timespan so a failure in one
+// window does not blank the gauges for windows that may still be succeeding.
+func (m *storeMetricExporter) ResetActiveUsers(timeSpanLabel string) {
+	m.usersActive.DeletePartialMatch(prometheus.Labels{"timespan": timeSpanLabel})
+}
+
+// ResetActiveAppSessions clears only the rows for the given timespan; see ResetActiveUsers.
+func (m *storeMetricExporter) ResetActiveAppSessions(timeSpanLabel string) {
+	m.appSessionsActive.DeletePartialMatch(prometheus.Labels{"timespan": timeSpanLabel})
+}
+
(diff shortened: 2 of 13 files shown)


nitronode/metrics/exporter.go
```diff
@@ -165,6 +165,27 @@ func (m *storeMetricExporter) SetUserBalanceReleasable(blockchainID, asset strin
...
+}
+
+// ResetActiveUsers clears only the rows for the given timespan so a failure in one
+// window does not blank the gauges for windows that may still be succeeding.
+func (m *storeMetricExporter) ResetActiveUsers(timeSpanLabel string) {
+	m.usersActive.DeletePartialMatch(prometheus.Labels{"timespan": timeSpanLabel})
+}
+
+// ResetActiveAppSessions clears only the rows for the given timespan; see ResetActiveUsers.
+func (m *storeMetricExporter) ResetActiveAppSessions(timeSpanLabel string) {
```
nksazonov, line 189: **[P1] No test for the reset-only-on-success invariant.**

The behavioral contract these methods implement — drop stale label tuples on a successful query, preserve gauge values on DB error — is the core correctness guarantee of this PR, and it goes entirely untested. In particular, `ResetActiveUsers`/`ResetActiveAppSessions` use `DeletePartialMatch` with a timespan label (not a full `Reset()`), so their partial-scope semantics are also unverified.

Suggestion: add `nitronode/metrics/exporter_test.go`. For each method: (1) set a gauge with label set A, (2) call Reset, (3) set with label set B ⊂ A, (4) gather and assert that A∖B series are absent. For the timespan variants, add a case that verifies only the target timespan rows are cleared while other timespan rows survive.

nitronode/metrics/exporter.go
```diff
@@ -376,9 +397,54 @@ func NewRuntimeMetricExporter(reg prometheus.Registerer) (RuntimeMetricExporter,
...
+// allActionResults is the closed enum of result label values used by counters
+// whose outcome dimension is success/failed.
+var allActionResults = []ActionResult{ActionResultSuccess, ActionResultFailed}
+
+// SeedRPCMethodMetrics publishes 0-valued series for every (method, path="default",
+// result) and (msg_type, method) tuple, so per-method PromQL queries return defined
+// values immediately after boot. Methods whose request payload determines path
+// (intent/transition) are not enumerated here — those variants appear on first call
+// and are bounded by their respective enums.
+func (m *runtimeMetricExporter) SeedRPCMethodMetrics(methods []string) {
```
nksazonov: **[P1] No test for `SeedRPCMethodMetrics` / `SeedBlockchainEventMetrics`.**

These are the primary fix for cold-start counter silence, yet neither function has a test. A regression (e.g. accidentally dropping one `WithLabelValues` call) would be invisible.

Suggestion: use `prometheus.NewRegistry()` to isolate tests: call `NewRuntimeMetricExporter(reg)`, invoke each seeder with known inputs, then use `prometheus/testutil.GatherAndCompare` or `CollectAndCount` to assert the expected series exist at value 0 before any `Inc*` call.

pkg/rpc/node.go
```diff
@@ -379,6 +384,17 @@ func (wn *WebsocketNode) Handle(method string, handler Handler) {
 	wn.routes[method] = []string{wn.groupId, method}
 }
 
+// RegisteredMethods returns every method registered on the node. Order is not
+// guaranteed. The underlying map is not mutex-protected, so the call must happen
+// after registration has finished (registration is expected to be single-threaded).
+func (wn *WebsocketNode) RegisteredMethods() []string {
```
nksazonov, line 390: **[P1] No test for `RegisteredMethods()` completeness.**

Both root-level `Handle` (line 384) and group-level `Handle` (line 561) write to `wn.routes`. If the group-level write were ever accidentally dropped, `RegisteredMethods()` would return a partial list and the seeded counter set would be incomplete — with no test to catch it.

Suggestion: add a test that creates a `WebsocketNode`, registers methods via both `Handle` and `group.Handle`, calls `RegisteredMethods()`, sorts the result, and asserts all registered methods appear. Cover the empty-node edge case too.

nitronode/metrics/exporter.go
```diff
@@ -376,9 +397,54 @@ func NewRuntimeMetricExporter(reg prometheus.Registerer) (RuntimeMetricExporter,
...
+	const defaultPath = "default"
+	for _, method := range methods {
+		for _, mt := range msgTypes {
+			m.rpcMessagesTotal.WithLabelValues(mt.String(), method)
+		}
+		for _, res := range allActionResults {
+			m.rpcRequestsTotal.WithLabelValues(method, defaultPath, res.String())
+			m.rpcRequestDurationSeconds.WithLabelValues(method, defaultPath, res.String())
+		}
+	}
```
nksazonov, line 441: **[P2] `rpcInflight` gauge is not seeded here despite having a bounded label space.**

The per-method `rpcInflight` GaugeVec has a single `method` label — the same bounded, fully-known label space as the counters seeded above. Any `absent()`-style alert watching inflight metrics will still fire during the cold-start window for each method until its first request, inconsistent with the goal of this PR.

Suggestion: add `m.rpcInflight.WithLabelValues(method)` inside the `for _, method := range methods` loop alongside the existing seeding calls.

nitronode/main.go
```diff
@@ -355,85 +355,107 @@ func runStoreMetricsExporter(
...
+	// dropped out of the store (e.g., the last open channel for an asset closed)
+	// disappear from the metric instead of sticking at their previous value. Reset
+	// happens only inside the successful-query branch so a transient DB error does
+	// not blank a healthy gauge.
+	refresh := func() {
+		channelCounts, err := store.GetChannelsCountByLabels()
+		if err != nil {
+			logger.Error("failed to get channel counts by labels", "error", err)
+		} else {
+			metricExported.ResetChannels()
```
nksazonov, line 378: **[P2] `Reset()` → `Set()` creates a brief absent-gauge window on every tick.**

Between `ResetChannels()` and the first `SetChannels(...)` call the gauge has no series. A Prometheus scrape landing in that window (even a microsecond one) will see `channels_total` as absent — exactly the condition `absent()`-based alerts catch. With a 30 s scrape interval matching the refresh interval, every tick is a potential false-positive window.

Suggestion: either build the new snapshot into a local map first and then call `Delete()` only for label tuples absent from the new snapshot (instead of `Reset()` + full re-publish), or explicitly document that the blink is accepted and recommend that `absent()` alerts use `absent_over_time([2m])` to tolerate it.

nitronode/metrics/exporter.go
```diff
@@ -376,9 +397,54 @@ func NewRuntimeMetricExporter(reg prometheus.Registerer) (RuntimeMetricExporter,
...
+			m.channelStateSigValidationsTotal.WithLabelValues(st.String(), res.String())
+		}
+	}
+
 	return m, nil
 }
 
+// allActionResults is the closed enum of result label values used by counters
+// whose outcome dimension is success/failed.
+var allActionResults = []ActionResult{ActionResultSuccess, ActionResultFailed}
```
nksazonov: **[P2] `allActionResults` is declared after the function that first consumes it.**

Go initialises package-level vars before any function call so this is correct, but the conventional style is to declare vars before the functions that use them. Declaring the dependency eight lines below its first consumer violates reading order and can look like a forward reference.

Suggestion: move `allActionResults` to before `NewRuntimeMetricExporter` (e.g. after the `const` block), so readers encounter the dependency before the function body.

pkg/rpc/node.go
```diff
@@ -56,6 +56,11 @@ type Node interface {
 	// Groups can have their own middleware and can be nested to create
 	// hierarchical handler structures.
 	NewGroup(name string) HandlerGroup
+
+	// RegisteredMethods returns every RPC method name registered on the node
+	// (across the root group and all subgroups). Intended for instrumentation
+	// at startup, not for request routing.
+	RegisteredMethods() []string
```
nksazonov, line 63: **[P2] Instrumentation introspection added to the core routing interface.**

`Node`'s contract is request routing and connection management. Adding `RegisteredMethods()` means any future alternative `Node` implementation (e.g. an in-process or HTTP/2 transport) must implement startup-metric seeding even if it has no Prometheus integration.

Suggestion: define a separate optional interface (e.g. `InstrumentableNode`) with only `RegisteredMethods() []string`, and use a type assertion at the call-site in `rpc_router.go`. If the node doesn't satisfy `InstrumentableNode`, skip seeding (or log a warning). This keeps `Node` minimal and the instrumentation coupling explicit and opt-in.

nksazonov (changes requested):
Good work overall — the three-part fix is correct and the inline documentation is thorough. A few things need attention before this can land, primarily around test coverage for the new behavioral contracts and one instrumentation completeness gap.

**Note — no test files in `nitronode/metrics/`:** Comments 1 and 2 both require new files (`nitronode/metrics/exporter_test.go`, `pkg/rpc/node_test.go`). The metrics package has zero test coverage today.

nitronode/metrics/exporter.go
philanton: Yep, fair. Added unit tests for the Reset* methods themselves in `nitronode/metrics/exporter_test.go` in [3b185cb](https://github.com/layer-3/nitrolite/pull/747/changes/3b185cbe1749aeef2e1ae073f00e2826a86252f9) — covers the full-Reset cases plus the timespan-scoped `DeletePartialMatch` behavior for `ResetActiveUsers` / `ResetActiveAppSessions` (other timespans survive).

The reset-only-on-success contract itself lives in the orchestration in `nitronode/main.go` (the `refresh` closure), which is harder to reach without extracting it onto a small struct. I'd hold off on that refactor in this PR — the branching is `if err … else ResetX(); SetX(...)` at every site and visually trivial. Happy to do it as a follow-up if you want it covered.

nitronode/metrics/exporter.go
philanton: Done in [3b185cb](https://github.com/layer-3/nitrolite/pull/747/changes/3b185cbe1749aeef2e1ae073f00e2826a86252f9):

- `TestSeedRPCMethodMetrics_PublishesZeroValuedSeries` — verifies `rpc_messages_emitted_total`, `rpc_requests_total`, `rpc_request_duration_seconds` and `rpc_inflight` series exist at 0 across the expected label tuples;
- `TestSeedRPCMethodMetrics_EmptyMethodsIsNoop` — guards the empty-input branch;
- `TestSeedBlockchainEventMetrics_PublishesZeroValuedSeries` — same idea per `(blockchain_id, result)`;
- `TestSeedBlockchainEventMetrics_PreservesPriorIncrements` — catches the regression where someone swaps `WithLabelValues(...)` for something that resets an already-incremented counter.

The channel-state-sig seeding in `NewRuntimeMetricExporter` itself is covered by `TestNewRuntimeMetricExporter_SeedsChannelStateSigValidations`.

pkg/rpc/node.go
philanton: Done in [3b185cb](https://github.com/layer-3/nitrolite/pull/747/changes/3b185cbe1749aeef2e1ae073f00e2826a86252f9). `pkg/rpc/node_test.go` covers:

- empty node returns empty slice;
- root-level `Handle` registers (`TestRegisteredMethods_RootHandle`);
- group + nested-group `Handle` are all surfaced (`TestRegisteredMethods_IncludesGroupRegistrations`) — that exercises both writes to `wn.routes`.

nitronode/metrics/exporter.go
philanton: Good catch. Added `m.rpcInflight.WithLabelValues(method)` inside the seeding loop and updated the doc comment + test in [3b185cb](https://github.com/layer-3/nitrolite/pull/747/changes/3b185cbe1749aeef2e1ae073f00e2826a86252f9).

nitronode/main.go
philanton: Hmmm, the race is real but I think it's mostly theoretical at our scrape cadence — the `refresh` closure runs in microseconds and a scrape landing inside that window with no preceding sample of the same series is pretty unlikely. The Delete-only-stale-tuples rewrite is non-trivial too (need to diff against the previous snapshot for each gauge family, and keep that snapshot somewhere), and I'd rather not bundle it here.

I'd say the right mitigation on the alerting side is `absent_over_time(...[2m])` for all `absent()`-style rules — that tolerates both this blink and ordinary scrape gaps. Happy to follow up with a doc note on the alert side. Let me know if you'd prefer a separate PR for the Delete-diff variant though, I can do that.

nitronode/metrics/exporter.go
philanton: Moved in [3b185cb](https://github.com/layer-3/nitrolite/pull/747/changes/3b185cbe1749aeef2e1ae073f00e2826a86252f9).

pkg/rpc/node.go
philanton: Yeah, I went back and forth on this. I'd lean towards keeping it on `Node` for now — there's only one implementation today and the optional-interface + type-assertion pattern adds a runtime branch at the call site for a hypothetical second transport. If/when an HTTP/2 or in-process `Node` impl actually lands, splitting `InstrumentableNode` out at that point is mechanical and the call site is one line. Happy to revisit then.

nitronode/main.go
nksazonov: Yes, the doc-note would be fine.

nksazonov (approved):

nitronode/metrics/exporter.go
```diff
@@ -376,9 +401,51 @@ func NewRuntimeMetricExporter(reg prometheus.Registerer) (RuntimeMetricExporter,
...
+// appear on first call and are bounded by their respective enums.
+func (m *runtimeMetricExporter) SeedRPCMethodMetrics(methods []string) {
+	msgTypes := []rpc.MsgType{rpc.MsgTypeReq, rpc.MsgTypeResp, rpc.MsgTypeRespErr}
+	const defaultPath = "default"
+	for _, method := range methods {
+		for _, mt := range msgTypes {
+			m.rpcMessagesTotal.WithLabelValues(mt.String(), method)
+		}
+		for _, res := range allActionResults {
+			m.rpcRequestsTotal.WithLabelValues(method, defaultPath, res.String())
```
ihsraham: Before closing YNU-917, I would seed the non-default RPC path values here too. `rpc_requests_total` and `rpc_request_duration_seconds` are currently pre-created only for `path="default"`, but runtime labeling returns bounded enum paths for `app_sessions.v1.submit_app_state`, `channels.v1.request_creation`, and `channels.v1.submit_state` via `getMethodPath`. That leaves valid bounded combinations absent on startup, for example `channels.v1.submit_state` with `path="escrow_lock"`, until first traffic.

One way to close AC3 is to enumerate those method-specific path domains from the existing `AppStateUpdateIntent` and `TransitionType` values and seed `{method,path,result}` for those bounded paths as well. If the intended scope is only default paths, please spell that out in the YNU-917 acceptance notes because it is narrower than "every counter with a fully-known label space".

philanton: Yeah, fair — landed in [b1300a3](https://github.com/layer-3/nitrolite/pull/747/changes/b1300a3d3e09f67aa9211896e77cb1929802fcad). `SeedRPCMethodMetrics` now takes a `method->paths` map, and `nitronode/api.MethodPathDomains()` enumerates the bounded enums for the three payload-derived methods (`AppStateUpdateIntent` for `app_sessions.v1.submit_app_state`, `TransitionType` for `channels.v1.{request_creation,submit_state}`). Methods absent from the map keep the default-only behavior.

To keep this from silently regressing if a new enum value lands without the seed update, I exported `app.AllAppStateUpdateIntents` and `core.AllTransitionTypes` next to their const blocks, and added a drift-guard test (`TestMethodPathDomains_MatchesEnums` + `TestMethodPathDomains_CoversGetMethodPathSwitch`) that pins the map against both the enums and `getMethodPath`'s switch.

ihsraham: Confirmed, this closes my concern. `MethodPathDomains()` now enumerates the bounded path domains for `app_sessions.v1.submit_app_state`, `channels.v1.request_creation`, and `channels.v1.submit_state`, and `SeedRPCMethodMetrics` seeds those method-specific paths instead of only `path="default"`.

The drift tests also cover both the enum lists and the `getMethodPath` switch, so a new bounded path should fail tests if the seed map is not updated.

ihsraham (approved):
Approving from my side. The bounded RPC path seeding concern is addressed, the thread is resolved, and the current checks are green.

PR #748, opened 2026-05-12
YNU-915: return success on absent-resource GET endpoints

philanton:
## Summary

Four single-resource lookups (`channels.v1.get_latest_state`, `channels.v1.get_home_channel`, `channels.v1.get_escrow_channel`, `app_sessions.v1.get_app_definition`) used to respond with `RespErr` when the record was absent. That counted "expected absence" as a failure in `rpc_requests_total{result="failed"}`, skewing success-rate dashboards, and forced clients to parse error strings to distinguish "not found" from a real failure.

This PR makes absence a successful response: the relevant payload field is now optional/nullable. Handlers drop the `nil → Fail` branch. Go and TS SDK methods return `nil`/`null` on absence; callers explicitly check for nil instead of treating the error path as the "no channel yet" branch.

Notion: https://www.notion.so/yellow-net/Return-empty-response-instead-of-error-for-absent-resource-GET-endpoints-35d75b8feb0480bcba31d203a64bc208

## Changes

- `pkg/rpc/api.go` — pointerize `Channel` / `State` / `Definition` response fields, `omitempty`
- `nitronode/api/channel_v1/{get_latest_state,get_home_channel,get_escrow_channel}.go`, `app_session_v1/get_app_definition.go` — drop `nil → Fail`, return success with empty payload
- `sdk/go/{channel,app_session}.go`, `cerebro/commands.go`, `sdk/go/examples/challenge/main.go` — return `(nil, nil)` on absence; callers (`Deposit`, `Withdraw`, `Transfer`, `Acknowledge`, `CloseHomeChannel`, `Checkpoint`, `SubmitAppSessionDeposit`) handle nil explicitly
- `sdk/ts/src/{rpc/api.ts,client.ts}` — return `Promise<T | null>`, drop redundant `try/catch` around `getLatestState` in high-level ops; explicit nil checks in `closeHomeChannel` / `checkpoint` / `submitAppSessionDeposit`
- `sdk/ts-compat/src/client.ts` — nil-aware (`getChannelData`, asset scans, `getAppDefinition`, `getEscrowChannel`)
- `docs/api.yaml` — mark 4 response fields `optional: true`, remove `*_not_found` errors
- Tests: `*_NotFound` paths now assert success + nil payload; success paths add `require.NotNil`; new `NotFound` cases for `get_latest_state` and `get_escrow_channel`; TS public-API drift snapshot refreshed

## Test plan

- [x] `go test ./...` passes
- [x] `go vet ./...` clean
- [x] `cd sdk/ts && npm test` passes (154/154)
- [x] `cd sdk/ts-compat && npm test` passes (34/34)
- [x] `cd sdk/ts && npm run typecheck` clean
- [x] `cd sdk/ts-compat && npm run typecheck` clean
- [ ] Verify on a deployed nitronode: `get_latest_state` for a wallet with no state returns `Resp` with `state` omitted, `rpc_requests_total{result="success"}` increments

## Compatibility

Wire-shape change: response payload fields become optional. SDK ≥ this PR works against new nitronode. Older SDKs see the response unchanged in the success case; they would only fail on absence if they assumed the field was always populated. Roll out nitronode first, then bump SDK consumers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Query APIs now treat missing channels, states, and app definitions as successful "not found" (null) results instead of errors; SDKs surface nullable return values and only error on real failures.

* **Documentation**
  * API docs updated to document omitted/optional response fields for not-found cases.

* **Tests**
  * Added and updated tests across services and SDKs to verify null/absent responses and adjusted success/not-found assertions.

* **Examples**
  * Examples updated to guard against absent state results.

[Review Change Stack](https://app.coderabbit.ai/change-stack/layer-3/nitrolite/pull/748?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


sdk/ts-compat/src/client.ts
```diff
@@ -528,14 +528,14 @@ export class NitroliteClient {
...
-                if (ch.channelId === _channelId) {
+                if (ch && ch.channelId === _channelId) {
                     return {
                         channel: ch,
                         state: await this.innerClient.getLatestState(this.userAddress, info.symbol, false),
                     };
                 }
             } catch {
-                // no channel for this asset
+                // transient RPC failure for this asset — skip
```
ihsraham: With the new nullable semantics, `getHomeChannel` returns `null` for normal absence, so exceptions from this block are genuine failures. The catch also covers the matched channel’s `getLatestState` call; if that request fails, `getChannelData` reports `Channel ... not found` instead of surfacing the RPC/auth/transport error.

Please rethrow the lookup error, or only throw not-found after all lookups completed successfully and returned no match.

ihsraham (changes requested):
I would keep this open for two closure items.

The backend/API changes and the main Go/TS SDK paths look aligned with the new absence semantics, and the current checks are green. Before this is ready, I would address:

1. `sdk/ts-compat` still swallows real `getHomeChannel` / `getLatestState` failures in `getChannelData` and reports them as `Channel ... not found` (inline).
2. YNU-915 AC8 asks for `test/integration` absent-then-present coverage per endpoint, but this diff only adds handler/unit coverage and does not add or update integration tests.

The unresolved `assert` / `require` note in `get_latest_state_test.go` looks non-blocking to me.

sdk/ts-compat/src/client.ts
```diff
@@ -528,14 +528,14 @@ export class NitroliteClient {
         for (const [, info] of this.assetsBySymbol) {
             try {
                 const ch = await this.innerClient.getHomeChannel(this.userAddress, info.symbol);
-                if (ch.channelId === _channelId) {
+                if (ch && ch.channelId === _channelId) {
                     return {
                         channel: ch,
```
nksazonov: `getLatestState` now returns `null` when no state exists. This line passes that null directly into the return object as `state`, so callers receive `{ channel: ch, state: null }` instead of a state value. Because the return type is `Promise<any>`, TypeScript won't catch downstream `data.state.homeLedger.userBalance` accesses.

The previous behavior was for `getLatestState` to throw (caught by the surrounding `try` block), so callers never saw a null state. The nil guard was added everywhere else but missed here.

**Suggestion:** `state: (await this.innerClient.getLatestState(...)) ?? undefined`, or at minimum document that `state` can be null so callers guard it themselves.

sdk/go/client_test.go
```diff
@@ -23,7 +23,7 @@ func TestClient_GetHomeChannel(t *testing.T) {
 	mockDialer.Dial(context.Background(), "", nil)
 
 	mockResp := rpc.ChannelsV1GetHomeChannelResponse{
-		Channel: rpc.ChannelV1{
+		Channel: &rpc.ChannelV1{
```
nksazonov, line 26: `TestClient_GetHomeChannel` (and the three companion tests for `GetEscrowChannel`, `GetLatestState`, `GetAppDefinition`) test only the happy path with a populated response. None of the four exercise the nil-return path: a response where `Channel`/`State`/`Definition` is absent.

The nil-guard code in `sdk/go/channel.go` and `sdk/go/app_session.go` that returns `(nil, nil)` on absence is therefore untested at the client layer.

**Suggestion:** Add a `_NilResponse` variant for each test that registers an empty response (`rpc.ChannelsV1GetHomeChannelResponse{}`), calls the method, and asserts `(nil, nil)`.

sdk/ts/test/unit/client.test.ts
```diff
@@ -11,12 +11,13 @@ const HOME_CHANNEL_ID = '0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
...
 
-function createHighLevelClient(latestState?: core.State, latestStateError?: Error) {
+function createHighLevelClient(latestState?: core.State | null, latestStateError?: Error) {
     const getLatestState = jest.fn();
     if (latestStateError) {
         getLatestState.mockRejectedValue(latestStateError);
     } else {
-        getLatestState.mockResolvedValue(latestState);
+        // Default to null (absence) when no state provided.
+        getLatestState.mockResolvedValue(latestState ?? null);
```
nksazonov, line 20: `createHighLevelClient` now handles `null` for `getLatestState`, but there are no tests for `getHomeChannel`, `getEscrowChannel`, or `getAppDefinition` returning `null`. The nil-guard code at `client.ts:1071–1169` that propagates `null` for each of these methods is untested at the `Client` level.

**Suggestion:** Add tests that mock `rpcClient` to return responses with `channel`/`state`/`definition` absent (undefined/null), and assert each `Client` method returns `null` without throwing.

nksazonov (commented):
Good work — clean approach, and the core change (nil instead of error for absent resources) is implemented consistently across all four endpoints. A few things need attention before landing.

---

**Out-of-diff notes:**

**`sdk/ts/examples/example-app/src/components/WalletDashboard.tsx:209`** — Inside the allowance-error recovery path for `handleCheckpoint`, `state.homeLedger.blockchainId` is accessed without a null check. `getLatestState` now returns `null` for absent state; if this path is hit before a channel exists, it throws `TypeError: Cannot read properties of null`. Fix: guard with `if (!state) throw new Error('no state available for approval');` before the `blockchainId` access.

**`sdk/ts/examples/example-app/src/components/ActionModal.tsx:123`** — Same issue: `latestState.homeLedger.blockchainId` immediately after `getLatestState(...)` with no null check. Same fix pattern.

**`sdk/ts-compat/src/rpc.ts:434`** — `parseGetAppDefinitionResponse` is a public exported compat function. When the server omits `definition` (new nil-response case), it returns `{ params: {} }`. Callers who access `params.protocol`, `params.quorum`, or `params.participants` get `undefined` silently — no error, no null signal. Fix: return `{ params: null }` and document that callers must check before accessing fields.

**`sdk/ts-compat/test/unit/`** — The nil-to-throw bridges added in `NitroliteClient.getAppDefinition` and `NitroliteClient.getEscrowChannel` (both throw when inner SDK returns null) have no tests. Add a test for each that mocks `innerClient` to return null and verifies the expected throw.

sdk/ts/test/unit/client.test.ts
```diff
@@ -138,8 +139,8 @@ describe('Client.getOnChainBalance', () => {
 });
 
 describe('Client.acknowledge', () => {
-    it('creates a channel with acknowledgement when latest state lookup fails', async () => {
-        const client = createHighLevelClient(undefined, new Error('state not found'));
+    it('creates a channel with acknowledgement when no latest state exists', async () => {
+        const client = createHighLevelClient(null);
```
nksazonov, line 143: `Client.checkpoint`, `Client.closeHomeChannel`, and `Client.submitAppSessionDeposit` have zero unit tests in this file. The new nil guards at `client.ts:1025` (`if (!state)` in `checkpoint`), `client.ts:1035` (`if (!channel)` in `checkpoint`), `client.ts:731` (`closeHomeChannel`), and `client.ts:1493` (`submitAppSessionDeposit`) are uncovered.

**Suggestion:** Add minimal tests: mock `getLatestState` → `null` and assert `checkpoint`/`closeHomeChannel` throw; mock `getLatestState` → `null` and assert `submitAppSessionDeposit` throws with `"no channel state to advance"`.

nitronode/api/channel_v1/get_latest_state_test.go
```diff
@@ -236,3 +238,38 @@ func TestGetLatestState_NormalizesWallet(t *testing.T) {
...
+	ctx := &rpc.Context{
+		Context: context.Background(),
+		Request: rpc.Message{Method: "channels.v1.get_latest_state", Payload: payload},
+	}
+
+	handler.GetLatestState(ctx)
+
+	// Absence is a successful response with state == nil.
+	assert.NotNil(t, ctx.Response.Payload)
+	assert.Nil(t, ctx.Response.Error())
```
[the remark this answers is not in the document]
philanton: Done in [50f812e](https://github.com/layer-3/nitrolite/pull/748/files/50f812eae05378ee62957cd26c1017df8efc337a).

sdk/ts-compat/src/client.ts
philanton: Yep, good catch. Rewrote `getChannelData` in [50f812e](https://github.com/layer-3/nitrolite/pull/748/files/50f812eae05378ee62957cd26c1017df8efc337a) to track lookup errors per asset and rethrow if every `getHomeChannel` call failed — only per-asset misses are suppressed now. `getLatestState` on the matched channel also runs outside the catch so its failures surface as well.

sdk/ts-compat/src/client.ts
philanton: Right, this would have leaked `state: null` straight into v0.5.3 callers and crashed them on `data.state.homeLedger.userBalance`. Fixed in [50f812e](https://github.com/layer-3/nitrolite/pull/748/files/50f812eae05378ee62957cd26c1017df8efc337a) — matched channel with null latest state now throws an explicit error instead of returning the null.

sdk/go/client_test.go
philanton: Added `_NilResponse` variants for all four endpoints in [50f812e](https://github.com/layer-3/nitrolite/pull/748/files/50f812eae05378ee62957cd26c1017df8efc337a).

sdk/ts/test/unit/client.test.ts
philanton: Added in [50f812e](https://github.com/layer-3/nitrolite/pull/748/files/50f812eae05378ee62957cd26c1017df8efc337a) — three null-return tests covering `getHomeChannel`, `getEscrowChannel`, and `getAppDefinition` at the `Client` level.

sdk/ts/test/unit/client.test.ts
philanton: Added in [50f812e](https://github.com/layer-3/nitrolite/pull/748/files/50f812eae05378ee62957cd26c1017df8efc337a) — three nil-state guard tests for `checkpoint`, `closeHomeChannel`, and `submitAppSessionDeposit`, each stubbing `getLatestState` to `null` and asserting the expected error message.

philanton:
@ihsraham both closure items addressed in [50f812e](https://github.com/layer-3/nitrolite/pull/748/files/50f812eae05378ee62957cd26c1017df8efc337a):

1) `sdk/ts-compat/getChannelData` rewritten — per-asset `getHomeChannel` errors are tracked but no longer all swallowed; if every lookup blew up, the last error is rethrown instead of the bogus "not found". `getLatestState` on the matched channel runs outside the catch so its failures surface too. Also covers @nksazonov's catch on the `state: null` leak — matched channel with absent state now throws explicitly instead of returning `{channel, state: null}` to v0.5.3 callers.
2) AC8 — the ticket was edited to drop the `test/integration` coverage, since we don't have that setup wired up yet(. Handler/unit coverage stays.

Also picked up @nksazonov's test-coverage notes — added `Client`-level null-return tests for `getHomeChannel`/`getEscrowChannel`/`getAppDefinition` and nil-state guard tests for `checkpoint`/`closeHomeChannel`/`submitAppSessionDeposit`, plus matching `_NilResponse` / `_NilChannel` cases on the Go side.

sdk/ts-compat/src/client.ts
```diff
@@ -525,18 +525,30 @@ export class NitroliteClient {
...
+            if (!ch || ch.channelId !== _channelId) {
+                continue;
+            }
+            const state = await this.innerClient.getLatestState(this.userAddress, info.symbol, false);
+            if (!state) {
+                throw new Error(`Channel ${_channelId} has no latest state`);
+            }
+            return { channel: ch, state };
+        }
+        if (!sawSuccessfulLookup && lookupError) {
```
ihsraham: This still masks partial lookup failures. If one asset lookup throws and any other asset lookup succeeds without matching, `sawSuccessfulLookup` is true and this falls through to `Channel ... not found` even though the requested channel may be behind the failed lookup.

With the nullable contract, misses come back as `null`; a caught exception means that asset was not inspected. Please rethrow when `lookupError` is set before emitting not-found, unless a matching channel was already found.

ihsraham (changes requested):
Re-reviewed the latest head `50f812eae05378ee62957cd26c1017df8efc337a` and the author replies.

The added Go/RPC/TS unit coverage and the fail-fast assertion change look good. I am not keeping the `test/integration` item as a blocker if YNU-915 was updated as described in the author reply.

One compat issue remains: the `getChannelData` fix still converts partial lookup failures into `Channel ... not found`.

sdk/ts-compat/src/client.ts
philanton: Yep, you're right — the `sawSuccessfulLookup` gate let partial failures slip through to "not found". Dropped it in [ea3dcb5](https://github.com/layer-3/nitrolite/pull/748/files/ea3dcb53af2dbb98910a29079cb0fb17353816b7). Now any unresolved `lookupError` rethrows before the not-found path, since a caught exception means that asset wasn't inspected and we can't conclude absence.

ihsraham (approved):
Approving from my side. The latest `getChannelData` update now rethrows any captured lookup error before the not-found path, so the partial-failure case is covered.

I also rechecked the author replies and the added nil-response coverage from the prior commit. No remaining blocker from my previous review.

nksazonov (approved):
Great job!

PR #749, opened 2026-05-13
MF-L08: fix(nitronode/api): default get_last_key_states to active-only with include_inactive opt-in

philanton:
## Summary

- `get_last_key_states` on both `channels.v1` and `app_sessions.v1` returned the latest version per session key with no `expires_at` filter, so expired latest states surfaced as "active" — contradicting endpoint name and docs (auth itself was unaffected because `GetAppSessionKeyOwner` / `ValidateChannelSessionKeyForAsset` apply their own expiry filter).
- Add `include_inactive` request flag, defaulting `false`. With the default the store filters `expires_at > now` on both the list and the count using a single `now` binding so pagination stays consistent. Setting it to `true` returns all latest states (expired or revoked) for ops/debug.
- Cerebro version-detection caller opts in to `include_inactive=true` so post-expiry rotation still observes the prior version and the server-side monotonic pointer is not violated.

## Out of scope (follow-up PR)

True revocation semantics (allowing `expires_at` in the past at submit time, freeing the per-user cap slot for revoked keys, log on revoke, doc pass for re-activation) ship on a separate branch. This PR is purely the API correctness fix; the flag is forward-compatible with that work.

## Test plan

- [x] `go vet ./...`
- [x] `go test ./...` (full suite green, incl. real-DB tests in `nitronode/store/database`)
- [ ] CI: `test-go.yml`
- [ ] Manual: hit `channels.v1.get_last_key_states` / `app_sessions.v1.get_last_key_states` with and without `include_inactive` against a node carrying a mix of expired and active keys

## Release Notes

* **New Features**
  * Added validator registration event monitoring for enhanced security visibility
  * Added pagination and filtering support for session key state queries
  * Enhanced WebSocket DoS hardening with frame-size limits and rate limiting

* **Bug Fixes**
  * Fixed ERC20 transfer handling for oversized return data
  * Improved concurrent session key submission reliability
  * Fixed signature validation and state enforcement path validation

* **Documentation**
  * Updated security model with trust assumptions and validator responsibilities
  * Clarified on-chain protocol enforcement conditions and constraints
  * Added comprehensive validator monitoring guidance

[Review Change Stack](https://app.coderabbit.ai/change-stack/layer-3/nitrolite/pull/749)


cerebro/commands.go
```diff
@@ -1135,10 +1135,14 @@ func (o *Operator) createAppSessionKey(ctx context.Context, sessionKeyAddr, expi
...
-	// Determine version by fetching existing keys
+	// Determine version by fetching existing keys. include_inactive=true so an expired
+	// prior version still surfaces — otherwise rotation would restart from version 1 and
+	// collide with the monotonic pointer enforced server-side.
+	includeInactive := true
 	var version uint64 = 1
 	existingStates, err := o.client.GetLastAppKeyStates(ctx, wallet, &sdk.GetLastKeyStatesOptions{
-		SessionKey: &sessionKeyAddr,
+		SessionKey:      &sessionKeyAddr,
+		IncludeInactive: &includeInactive,
```
ihsraham, line 1170: The same opt-in is missing in `createChannelSessionKey` (`cerebro/commands.go:1015-1020`). That caller still passes only `SessionKey` with no `IncludeInactive`, so once the prior channel key for `(user, sessionKey)` has expired, the new active-only default filters it out, `version` stays at 1, and the next `submit_session_key_state` is rejected at `nitronode/api/channel_v1/submit_session_key_state.go:101` with `expected version N+1, got 1`.

Mirroring this block (the `IncludeInactive: &includeInactive` opt-in plus the `if s.Version >= version { version = s.Version + 1 }` walk below) keeps channel rotation working. The comment above applies verbatim there.

ihsraham (commented):
Looks close on the L-08 closure. The store and the API surface line up with what the finding asked for, and the count and list both share the same `now` so pagination stays consistent at page boundaries.

One blocker on the way out: `createChannelSessionKey` in `cerebro/commands.go` did not get the symmetric `include_inactive=true` opt-in that `createAppSessionKey` got. After the previous channel key expires, channel rotation will fail on submit. Inline note at the call site.

Worth tracking as follow-ups, not blocking on this PR: `sdk/ts` and `sdk/ts-compat` have no `include_inactive` knob yet, so TS callers cannot opt back into expired-included results from their side. The pagination test in `nitronode/store/database/...` also does not exercise mixed active and expired entries with `offset > 0`, which is the most likely place for count and list to drift if anything regresses later.

cerebro/commands.go
philanton: Yeah, you're right — channel rotation would've hit the same wall. Mirrored the block (`include_inactive=true` + the defensive `s.Version >= version` walk) in [2429842](https://github.com/layer-3/nitrolite/pull/749/changes/2429842ecf871d2719aaba28bd5dbf9b2940354d). The walk is overkill with the `SessionKey` filter today since it returns at most one row, but it keeps the two call sites symmetric if the filter ever loosens.

philanton:
Decided to roll both follow-ups into this PR rather than spinning another branch:

1) `sdk/ts` and `sdk/ts-compat` got an optional `{ includeInactive }` third arg on `getLastChannelKeyStates` / `getLastKeyStates`. Default omits `include_inactive` on the wire so the server's active-only default still kicks in unchanged. Two TS unit tests cover the forwarding and the default;
2) added the mixed active/expired pagination test (with `offset > 0`) on both channel and app session-key store tables, asserting count and list stay consistent under the same `now` for both `includeInactive` values.

Both in [2429842](https://github.com/layer-3/nitrolite/pull/749/changes/2429842ecf871d2719aaba28bd5dbf9b2940354d).

nitronode/api/app_session_v1/interface.go
```diff
@@ -56,7 +56,7 @@ type Store interface {
 	CountSessionKeysForUser(userAddress string) (uint32, error)
 	StoreAppSessionKeyState(state app.AppSessionKeyStateV1) error
 	GetLastAppSessionKeyVersion(wallet, sessionKey string) (uint64, error)
-	GetLastAppSessionKeyStates(wallet string, sessionKey *string, limit, offset uint32) ([]app.AppSessionKeyStateV1, uint32, error)
+	GetLastAppSessionKeyStates(wallet string, sessionKey *string, includeInactive bool, limit, offset uint32) ([]app.AppSessionKeyStateV1, uint32, error)
```
nksazonov, line 63: `GetLastAppSessionKeyStates` has no doc comment here, while the channel variant (`channel_v1/interface.go:108–111`) has a three-line comment explaining the `includeInactive` semantics. Since this is an interface method, the doc comment is the primary place for consumers to learn the behavior. Add a doc comment mirroring the channel side.

sdk/ts-compat/src/client.ts
```diff
@@ -894,8 +895,12 @@ export class NitroliteClient {
...
     }
 
-    async getLastKeyStates(userAddress: string, sessionKey?: string): Promise<AppSessionKeyStateV1[]> {
-        return this.innerClient.getLastKeyStates(userAddress, sessionKey);
+    async getLastAppKeyStates(
+        userAddress: string,
+        sessionKey?: string,
+        options?: { includeInactive?: boolean },
+    ): Promise<AppSessionKeyStateV1[]> {
+        return this.innerClient.getLastAppKeyStates(userAddress, sessionKey, options);
```
nksazonov, line 903: The compat layer's stated purpose is to bridge 0.5.x consumers to v1 without breaking changes. Removing `getLastKeyStates` with no deprecation stub silently breaks any existing caller. Consider keeping `async getLastKeyStates(...)` as a thin `@deprecated` wrapper that delegates to `getLastAppKeyStates`, or document this as a breaking change in the compat package's CHANGELOG with a version bump.

nksazonov (changes requested):
Clean implementation — the pagination consistency design (single `now` binding for both count and list queries) is well thought out, and test coverage is solid. One issue must be fixed before landing.

**⚠️ CI is broken:** `scripts/drift/runtime-smoke.mjs` calls `client.getLastKeyStates(wallet)`, which no longer exists after the rename. The drift check runs in `.github/workflows/main-pr.yml` and will throw `TypeError: client.getLastKeyStates is not a function`.

---

**Note — `scripts/drift/runtime-smoke.mjs:331–334`** (not in diff):
`client.getLastKeyStates(wallet)` is called on line 334, but this method was renamed to `getLastAppKeyStates` in this PR. The smoke test runs in CI (`check-protocol-drift.sh`) and will fail with a TypeError. Fix: rename the call (and the `logStep` label on line 331) to `getLastAppKeyStates`.

**Note — `sdk/ts/src/rpc/api.ts:147` and `:264`** (not in diff):
Both response type JSDoc comments still read "List of **active** ... session key states for the user". After adding `include_inactive`, the list can now contain expired states. Update to something like "Latest session key states for the user, filtered by `include_inactive`."

nitronode/api/app_session_v1/interface.go
philanton: Yep, good catch — interfaces are the canonical doc surface. Added the mirrored three-line comment in [d00376f](https://github.com/layer-3/nitrolite/pull/749/changes/d00376fe8210e0146b0d2ecee656d7379f63d4ba).

sdk/ts-compat/src/client.ts
philanton: Yeah, you're right — silently dropping the 0.5.x method from a compat layer kind of defeats the point. Restored `getLastKeyStates` as a thin `@deprecated` stub forwarding to `getLastAppKeyStates` in [d00376f](https://github.com/layer-3/nitrolite/pull/749/changes/d00376fe8210e0146b0d2ecee656d7379f63d4ba) — no `includeInactive` on it since the old signature didn't have one, callers wanting the new behavior should migrate. Also refreshed the README row, the public-api-drift snapshot, and added a forwarding unit test. Major-bump removal can ride a later release.

nksazonov (approved):

PR #751, opened 2026-05-13
MF-L09: fix(nitronode): validate parsed app session nonce

philanton:
## Summary
- `CreateAppSession` raw-string nonce check (`Nonce == "" || Nonce == "0"`) had a dead branch and a bypass: `unmapAppDefinitionV1` errors out on empty input before the check, and `strconv.ParseUint` accepts zero-padded inputs (`"00"`, `"000"`, ...) that parse to `0` and skip the `== "0"` comparison. Net result: an app session could be stored with `Nonce = 0`.
- Replaced the raw-string check with `appDef.Nonce == 0` so the validation runs against the parsed numeric value.
- Extended `TestCreateAppSession_ZeroNonce` into a table-driven test covering `"0"`, `"00"`, `"000"`.

## Test plan
- [x] `go test ./nitronode/api/app_session_v1/...`
- [x] `go vet ./nitronode/api/app_session_v1/...`

### DIFF
--- a/nitronode/api/app_session_v1/create_app_session.go
+++ b/nitronode/api/app_session_v1/create_app_session.go
@@ -60,7 +60,7 @@ func (h *Handler) CreateAppSession(c *rpc.Context) {
 
 	// Validate nonce
 	if appDef.Nonce == 0 {
-		c.Fail(nil, "nonce is zero or not provided")
+		c.Fail(nil, "nonce must be non-zero")
 		return
 	}
 
--- a/nitronode/api/app_session_v1/create_app_session.go
+++ b/nitronode/api/app_session_v1/create_app_session.go
@@ -59,7 +59,7 @@ func (h *Handler) CreateAppSession(c *rpc.Context) {
 		"nonce", reqPayload.Definition.Nonce)
 
 	// Validate nonce
-	if reqPayload.Definition.Nonce == "" || reqPayload.Definition.Nonce == "0" {
+	if appDef.Nonce == 0 {
 		c.Fail(nil, "nonce is zero or not provided")
 		return
 	}
--- a/nitronode/api/app_session_v1/create_app_session_test.go
+++ b/nitronode/api/app_session_v1/create_app_session_test.go
@@ -223,68 +223,70 @@ func TestCreateAppSession_QuorumWithMultipleSignatures(t *testing.T) {
 }
 
 func TestCreateAppSession_ZeroNonce(t *testing.T) {
-	// Setup
-	mockStore := new(MockStore)
+	// Zero-padded values must also be rejected: strconv.ParseUint accepts "00",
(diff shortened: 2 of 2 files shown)


ihsraham (approved):
Approving for the L-09 closure. The check now reads `appDef.Nonce == 0` on the parsed `uint64`, so every base-10 zero-spelling that `strconv.ParseUint` accepts is rejected, and the new table-driven `TestCreateAppSession_ZeroNonce` covers `"0"`, `"00"`, and `"000"` against the PoC.

Side note: this brings `app_session_v1` in line with the parsed-value pattern that `channel_v1/request_creation.go:45` already uses, so the two surfaces no longer disagree on how a non-zero nonce is enforced. I grepped every `strconv.ParseUint`/`Atoi` site across `nitronode`, `pkg`, and `sdk` for the same anti-pattern and came up clean.

One small thing, non-blocking: with the check moved after `unmapAppDefinitionV1`, an empty `""` nonce now surfaces `invalid app definition: invalid nonce: strconv.ParseUint: parsing "": invalid syntax` instead of the cleaner `nonce is zero or not provided`. Still a rejection, the error string is noisier. A short pre-check on `reqPayload.Definition.Nonce == ""` before the unmap would restore the friendlier message. Fine to skip if you'd rather not churn this PR.

nitronode/api/app_session_v1/create_app_session.go
```diff
@@ -59,7 +59,7 @@ func (h *Handler) CreateAppSession(c *rpc.Context) {
 		"nonce", reqPayload.Definition.Nonce)
 
 	// Validate nonce
-	if reqPayload.Definition.Nonce == "" || reqPayload.Definition.Nonce == "0" {
+	if appDef.Nonce == 0 {
 		c.Fail(nil, "nonce is zero or not provided")
```
nksazonov: The phrase "not provided" is misleading here. An empty or non-numeric nonce string is rejected earlier, in `unmapAppDefinitionV1` (lines 49–52), with a parse error — it never reaches this check. By the time execution reaches this line, the only reachable case is an integer `0`. The sibling handler `channel_v1/request_creation.go:46` uses the accurate message `"nonce must be non-zero"`.

**Suggestion:** Change to `"nonce must be non-zero"` to match the sibling handler and accurately describe what this check catches.

nksazonov (commented):
Good job! The fix is correct — the bypass via zero-padded decimal strings is properly closed by moving from a raw-string comparison to an integer check on the parsed value.

nitronode/api/app_session_v1/create_app_session.go
philanton: Yep, good catch. Fixed in [37219cc](https://github.com/layer-3/nitrolite/pull/751/commits/37219cc69347c83a0e6284f3c18fc2b61fcb8539).

nksazonov (approved):

PR #753, opened 2026-05-13
MF-M09: fix(nitronode): auto-challenge home channel on withheld escrow finalize

philanton:
## Summary
- When `EscrowDepositChallenged` fires without a newer fully-signed `FINALIZE_ESCROW_DEPOSIT` available locally, the node now schedules a `challengeChannel(...)` on the home blockchain using the `INITIATE_ESCROW_DEPOSIT` state and a node-produced challenger signature.
- Closes the cross-chain attack where a user lets the non-home escrow challenge expire, recovers escrow-chain funds, and then enforces FINALIZE on home against the node's locked allocation.
- Introduces generic `ActionTypeChallenge = 2` + `ScheduleChallenge(stateID, chainID)`. `BlockchainWorker` packs via `core.PackChallengeState`, signs with the node `ChannelSigner`, and submits via `BlockchainClient.Challenge(state, sig, ChannelParticipantNode)`.

## Behavior change in `HandleEscrowDepositChallenged`
- Branch A (newer signed FINALIZE exists) — unchanged, still schedules `FinalizeEscrowDeposit` on the escrow chain.
- Branch B (no newer FINALIZE, or last signed state is the INITIATE itself) — fetches the INITIATE state via `GetStateByChannelIDAndVersion`, resolves its home channel, and schedules `Challenge` on the home blockchain. Skips with a warning if the INITIATE state is missing locally, has no `HomeChannelID`, or the home channel is not `Open`.
- After the home challenge timer expires, operator runs `closeChannel(...)` manually to recover the node allocation (existing manual recovery path).

## Test plan
- [x] `go test ./nitronode/event_handlers/... -run TestHandleEscrowDepositChallenged -count=1`
- [x] `go test ./nitronode/... ./pkg/... -count=1` (full suite green)
- [x] `go build ./...`
- [x] `go vet ./...`
- [ ] Manual integration: user withholds FINALIZE state, observe home channel transitions to DISPUTED on-chain after escrow challenge.

## Summary by CodeRabbit

* **New Features**
  * Added validator registration event monitoring with recovery support on reconnection.
  * Implemented pagination (default/max page size 10) for key state query endpoints.
  * Added WebSocket per-connection rate limiting and maximum message size enforcement.

* **Documentation**
  * Clarified signature domain compatibility between on-chain and off-chain validators.
  * Updated protocol enforcement rules for channel state submission requirements.
  * Added validator registration monitoring and ERC-20 approval security guidance.

* **Configuration**
  * Added configurable WebSocket DoS hardening environment variables.
  * Made database SSL mode configurable (default: `require`).

* **Bug Fixes & Improvements**
  * Enhanced escrow operation locking and user signature persistence.
  * Enforced per-user session key caps for improved resource management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/layer-3/nitrolite/pull/753)

### DIFF
--- a/nitronode/event_handlers/service.go
+++ b/nitronode/event_handlers/service.go
@@ -359,7 +359,19 @@ func (s *EventHandlerService) scheduleHomeChannelChallengeForEscrowDeposit(ctx c
 		return nil
 	}
 	if homeChannel.Status != core.ChannelStatusOpen {
-		logger.Warn("home channel not Open, skipping auto-challenge", "homeChannelId", *initiateState.HomeChannelID, "homeStatus", homeChannel.Status, "escrowChannelId", escrowChanID)
+		switch homeChannel.Status {
+		case core.ChannelStatusChallenged:
+			logger.Warn("home channel already Challenged, skipping auto-challenge", "homeChannelId", *initiateState.HomeChannelID, "escrowChannelId", escrowChanID)
+		case core.ChannelStatusClosed:
+			logger.Error("home channel Closed, defense window passed", "homeChannelId", *initiateState.HomeChannelID, "escrowChannelId", escrowChanID)
+		default:
+			logger.Warn("home channel not Open, skipping auto-challenge", "homeChannelId", *initiateState.HomeChannelID, "homeStatus", homeChannel.Status, "escrowChannelId", escrowChanID)
+		}
+		return nil
+	}
+
+	if initiateState.HomeLedger.BlockchainID == 0 {
+		logger.Error("INITIATE_ESCROW_DEPOSIT state has zero home BlockchainID, cannot defend home channel automatically", "homeChannelId", *initiateState.HomeChannelID, "escrowChannelId", escrowChanID)
 		return nil
 	}
 
--- a/nitronode/event_handlers/service.go
+++ b/nitronode/event_handlers/service.go
@@ -266,9 +266,13 @@ func (s *EventHandlerService) HandleEscrowDepositInitiated(ctx context.Context, 
 }
 
 // HandleEscrowDepositChallenged processes the EscrowDepositChallenged event emitted when an escrow
-// deposit is challenged on-chain. Similar to home channel challenges, it marks the channel as Challenged,
-// sets the expiration time, and automatically schedules a checkpoint with the latest signed state
-// to resolve the challenge.
+// deposit is challenged on-chain. It marks the channel as Challenged and sets the expiration time.
+// Resolution policy depends on whether the node holds a newer fully-signed state for this channel:
+//   - If a newer signed FINALIZE_ESCROW_DEPOSIT exists, finalize the escrow on the non-home chain.
+//   - Otherwise the user is withholding finalize: defend the node allocation on the home chain by
+//     scheduling challengeChannel(...) with the INITIATE_ESCROW_DEPOSIT state. Without this, the user
+//     can let the non-home challenge expire, recover escrow-chain funds, and still threaten the
+//     home-chain finalize path against the node's locked allocation.
 func (s *EventHandlerService) HandleEscrowDepositChallenged(ctx context.Context, tx core.ChannelHubEventHandlerStore, event *core.EscrowDepositChallengedEvent) error {
 	logger := log.FromContext(ctx)
 	chanID := event.ChannelID
@@ -304,17 +308,17 @@ func (s *EventHandlerService) HandleEscrowDepositChallenged(ctx context.Context,
 	if err != nil {
 		return err
 	}
-	if lastSignedState == nil {
-		logger.Warn("no state found for channel during EscrowDepositChallenged event", "channelId", chanID)
-	} else if lastSignedState.Version <= event.StateVersion {
-		logger.Warn("last signed state version is not greater than challenged state version", "channelId", chanID, "lastSignedStateVersion", lastSignedState.Version, "challengedStateVersion", event.StateVersion)
-	} else {
+	if lastSignedState != nil && lastSignedState.Version > event.StateVersion {
 		if lastSignedState.EscrowLedger == nil {
 			logger.Warn("last signed state has no escrow ledger during EscrowDepositChallenged event", "channelId", chanID)
 		} else {
 			if err := tx.ScheduleFinalizeEscrowDeposit(lastSignedState.ID, lastSignedState.EscrowLedger.BlockchainID); err != nil {
 				return err
 			}
+		}
+	} else {
+		if err := s.scheduleHomeChannelChallengeForEscrowDeposit(ctx, tx, chanID, event.StateVersion); err != nil {
+			return err
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


nitronode/event_handlers/service.go
```diff
@@ -326,6 +330,51 @@ func (s *EventHandlerService) HandleEscrowDepositChallenged(ctx context.Context,
...
+// starts the home-chain challenge timer the operator uses to recover the node allocation.
+func (s *EventHandlerService) scheduleHomeChannelChallengeForEscrowDeposit(ctx context.Context, tx core.ChannelHubEventHandlerStore, escrowChanID string, stateVersion uint64) error {
+	logger := log.FromContext(ctx)
+
+	initiateState, err := tx.GetStateByChannelIDAndVersion(escrowChanID, stateVersion)
+	if err != nil {
+		return err
+	}
+	if initiateState == nil {
+		logger.Error("INITIATE_ESCROW_DEPOSIT state missing locally, cannot defend home channel automatically", "escrowChannelId", escrowChanID, "stateVersion", stateVersion)
```
nksazonov, line 345: I think we should have an alert on all such `logger.Error`s in `scheduleHomeChannelChallengeForEscrowDeposit` as this may mean some attack is in progress.

philanton: Yes, but that's out of scope for this PR

nitronode/event_handlers/service.go
```diff
@@ -326,6 +330,51 @@ func (s *EventHandlerService) HandleEscrowDepositChallenged(ctx context.Context,
...
+	if err != nil {
+		return err
+	}
+	if homeChannel == nil {
+		logger.Error("home channel not found, cannot defend home channel automatically", "homeChannelId", *initiateState.HomeChannelID, "escrowChannelId", escrowChanID)
+		return nil
+	}
+	if homeChannel.Status != core.ChannelStatusOpen {
+		logger.Warn("home channel not Open, skipping auto-challenge", "homeChannelId", *initiateState.HomeChannelID, "homeStatus", homeChannel.Status, "escrowChannelId", escrowChanID)
+		return nil
```
nksazonov, line 375: **Silent `return nil` on critical defense-skip paths.** When `initiateState` is missing, `HomeChannelID` is nil, or the home channel isn't found in the DB, the function logs an error and returns `nil` — which the caller interprets as success. The channel is now marked `Challenged` on-chain with no defensive action taken, but `HandleEscrowDepositChallenged` returns cleanly. The operator's only signal is a log line that can be missed, yet the on-chain challenge timer is already running.

**Suggestion:** Return a sentinel error (e.g. `ErrCannotDefendHomeChannel`) or increment an alerting metric before these `return nil` exits, so monitoring infrastructure surfaces the missed defense automatically. The silent swallow is the highest-risk part of this path.

nitronode/event_handlers/service.go
```diff
@@ -326,6 +330,51 @@ func (s *EventHandlerService) HandleEscrowDepositChallenged(ctx context.Context,
...
+	}
+
+	if err := tx.ScheduleChallenge(initiateState.ID, initiateState.HomeLedger.BlockchainID); err != nil {
```
nksazonov, line 378: **Missing `HomeLedger.BlockchainID == 0` guard.** `HomeLedger` is a value type (not a pointer), so it is never nil and no guard prevents using a zero `BlockchainID`. If the stored `INITIATE_ESCROW_DEPOSIT` state has an empty home ledger (e.g. from a DB migration gap), the action is scheduled for chain 0 where no `BlockchainWorker` exists — the challenge never fires and the node allocation remains exposed with no visible error.

**Suggestion:** Add `if initiateState.HomeLedger.BlockchainID == 0 { logger.Error(...); return <error or nil> }` before this call, matching the style of the existing nil guards above.

nitronode/event_handlers/service.go
```diff
@@ -326,6 +330,51 @@ func (s *EventHandlerService) HandleEscrowDepositChallenged(ctx context.Context,
...
+	homeChannel, err := tx.GetChannelByID(*initiateState.HomeChannelID)
+	if err != nil {
+		return err
+	}
+	if homeChannel == nil {
+		logger.Error("home channel not found, cannot defend home channel automatically", "homeChannelId", *initiateState.HomeChannelID, "escrowChannelId", escrowChanID)
+		return nil
+	}
+	if homeChannel.Status != core.ChannelStatusOpen {
+		logger.Warn("home channel not Open, skipping auto-challenge", "homeChannelId", *initiateState.HomeChannelID, "homeStatus", homeChannel.Status, "escrowChannelId", escrowChanID)
```
nksazonov: **`logger.Warn` doesn't distinguish recoverable from critical skip.** `ChannelStatusChallenged` means the home channel is already in dispute — the node is likely already defended, which is benign. `ChannelStatusClosed` means the finalization window has passed and recovery may no longer be possible. Both states trigger the same `"home channel not Open, skipping auto-challenge"` log at `Warn` level, making post-incident triage harder.

**Suggestion:** Log a distinct message per status — at minimum, distinguish `Challenged` (likely safe to skip) from `Closed` (potentially too late, worth an `Error`-level log).

nitronode/event_handlers/service_test.go
```diff
@@ -389,6 +389,163 @@ func TestHandleEscrowDepositChallenged_Success(t *testing.T) {
...
+
+	require.NoError(t, err)
+	mockStore.AssertExpectations(t)
+	mockStore.AssertNotCalled(t, "ScheduleChallenge", mock.Anything, mock.Anything)
+	mockStore.AssertNotCalled(t, "ScheduleFinalizeEscrowDeposit", mock.Anything, mock.Anything)
+}
+
 func TestHandleEscrowDepositFinalized_Success(t *testing.T) {
 	// Setup
 	mockStore := new(MockStore)
```
nksazonov, line 783: **Missing test coverage for the error paths that matter most.** The three new cases cover the happy path, the "home channel not Open" skip, and the "no local state" skip. The paths where the automated defense fails silently are not yet covered: `HomeChannelID == nil` on the fetched state, `homeChannel == nil` (DB returns nil for a valid ID), errors returned from `GetStateByChannelIDAndVersion`, and errors returned from `GetChannelByID(homeChannelID)`. These are the exact conditions under which the node is left exposed without a trace in the error return.

**Suggestion:** Add a test case for each — they map directly to the existing early-exit branches and can reuse the `MockStore` pattern already in place.

nksazonov (changes requested):
Clean approach and good test coverage of the happy-path scenarios. Several comments need to be addressed before this can land.

nitronode/event_handlers/service.go
philanton: Yep, valid concern. I considered returning a sentinel error here, but the outer tx would then roll back — including the `Status=Challenged` update on the escrow channel, which is the on-chain truth we need persisted. So metric/alert is the right fix here, not error propagation. Same scope as the alerts comment on :345 — I'll fold both into a follow-up.

nitronode/event_handlers/service.go
philanton: Yep, good catch. Fixed in [6dfe9ef](https://github.com/layer-3/nitrolite/pull/753/changes/6dfe9ef7c53ecb8e25af7411af020421d789bbc3) — added an explicit `HomeLedger.BlockchainID == 0` guard before `ScheduleChallenge`, logs at `Error` and returns nil to keep the outer tx intact.

nitronode/event_handlers/service.go
philanton: Yep, fair point. Split in [6dfe9ef](https://github.com/layer-3/nitrolite/pull/753/changes/6dfe9ef7c53ecb8e25af7411af020421d789bbc3) — `Warn` for `Challenged` (likely already defended), `Error` for `Closed` (defense window passed), and `Warn` default for any other non-Open status.

nitronode/event_handlers/service_test.go
philanton: Yep, covered in [6dfe9ef](https://github.com/layer-3/nitrolite/pull/753/changes/6dfe9ef7c53ecb8e25af7411af020421d789bbc3). Added five cases: `HomeChannelIDNil_SkipsChallenge`, `HomeChannelNotFound_SkipsChallenge`, `GetStateByVersionError_Propagates`, `GetHomeChannelError_Propagates`, and `HomeBlockchainIDZero_SkipsChallenge` (covering the new guard from :366 too).

nksazonov (approved):

PR #759, opened 2026-05-14
MF2-H01: harden receiver-state issuance and dispute resolution

philanton:
## Summary

Fixes two security findings:

- **MF2-H01** — receiver-credit issuance node-signed newer states for a `ChannelStatusChallenged` home channel, enabling dispute reset via checkpoint.
- **MF2-H02** — receiver's latest stored state could pin a closed home channel as the active head: `request_creation` rejected fresh channels with `channel is already initialized`, follow-up RPCs failed with `user has no active channel`, and incoming receives could still node-sign credits onto the closed channel. User had to migrate wallets to recover use of the asset.

Five remediations, each adding tests:

1. **Restrict node co-signing of receiver states to `Open` home channels.** `issueTransferReceiverState` (`channel_v1`) and `issueReleaseReceiverState` (`app_session_v1`) now allowlist via `CheckActiveChannel`: node-signs only when the receiver's active home channel is `Open`. Receiver states for `Challenged`, `Closing`, and `Closed` channels are persisted unsigned (still tracked for reconciliation, never returned for onchain submission). Closes the MF2-H01 dust-checkpoint vector and also closes the MF2-H02 vector of incoming sends getting node-signed onto an already-closed home channel.

2. **Backfill node sig on the off-chain head when a challenge clears.** `HandleHomeChannelCheckpointed` now recomputes the node signature for the channel's highest stored version (`GetLastStateByChannelID`) rather than the on-chain enforced version, so any during-challenge receiver state that became the new latest row is fully co-signed. `UpdateStateUserSigIfMissing` is generalized to `UpdateStateSigsIfMissing(channelID, version, userSig, nodeSig)`. Open→Open checkpoints stay on the prior user-sig-only path. Defensive guard skips backfill when the head transition is not a receiver-credit type.

3. **Squash challenge-window net channel change into `ChallengeRescue` on `Challenged → Closed`; emit unconditionally to clear the closed-channel head (MF2-H02).** New DB query `SumNetTransitionAmountAfterVersion` returns `receives − sends` above the closure version, where receives = (`TransferReceive`, `Release`) and sends = (`TransferSend`, `Commit`), across both signed and unsigned rows. Caller clamps net at zero (negative net is reachable only when the user closes at a version where her own channel balance exceeded the off-chain head — adversarial rollback of her own sends/commits; on-chain has already paid above head, rescue must not dock further). `HomeChannelClosed` emits a single `ChallengeRescue` (transition + transaction type `201`) on the user's ledger — fresh epoch, version 1, `HomeChannelID` nil, `AccountID` = closed channel ID — built via `core.NewChallengeRescueState`. Emission is unconditional on `Challenged → Closed` (zero-amount allowed): this is the MF2-H02 fix. Without it, the receiver's latest stored state retains the closed `HomeChannelID`, `IsFinal()` returns false on the non-finalize head, `NextState()` inherits the closed channel ID, `channels.v1.request_creation` fails with `channel is already initialized`, and `submit_state` calls fail with `user has no active channel`. The rescue advances the user's head to `HomeChannelID = nil` so `request_creation` proceeds and subsequent receiver issuance produces the "credit a user with no open channel" shape. `SumNetTransitionAmountAfterVersion` is pinned to the previous head's epoch; the caller asserts `prev.Version > closureVersion` whenever the sum is non-zero. (Cooperative `Open → Closing → Closed` closes via Finalize, so `IsFinal()` is true and no rescue is needed; that path is unaffected.)

4. **Serialize event-handler status mutations with receiver-issuance RPCs via `LockUserState`.** `HandleHomeChannelClosed`, `HandleHomeChannelChallenged`, and `HandleHomeChannelCheckpointed` now acquire `LockUserState(user, asset)` after type guards and before status mutation / sum / rescue / backfill. Closes a race window where an in-flight RPC reading `Status = Open` (or `Challenged`) via `CheckActiveChannel` could commit a node-signed or unsigned receiver row after the handler had already flipped status, producing orphaned rows (`Open` channel with unsigned latest head not reached by backfill; node-signed receiver on a disputed channel; receiver row above closure version that the rescue sum already finalized). `LockUserState` added to `core.ChannelHubEventHandlerStore` and `evm.ChannelHubReactorStore` (DBStore already implemented it).

5. **Document the accepted session-key off-chain scope-enforcement gap.** `SessionKeyValidator` on-chain validates cryptographic signatures only; off-chain `ValidateChannelSessionKeyForAsset` (expiration, asset scope) runs only on the node's acknowledgement path (`channels.v1.submit_state` with `TransitionTypeAcknowledgement`). A holder of a session key ever authorized on the channel — including expired, revoked, or retired keys — can fetch the node-signed receive state, sign manually, and submit directly on-chain. Outcome is bounded by states the user already signed and the receive state's strictly-additive allocation effect on the user, so the gap is griefing-only on the user's own sends, not theft; and the direct-submission path is the only recovery when the node is unavailable. Documented in `contracts/SECURITY.md` and `docs/protocol/security-and-limitations.md`.

Protocol docs (`docs/protocol/enforcement.md`, `protocol-description.md`) updated to describe the broader scope. `challenge_rescue` added to `transition_type` and `transaction_type` enums in `docs/api.yaml`. `contracts/SECURITY.md` updated for the net-change rescue scope and the accepted limitation.

## Test plan

- [x] `go test ./...` — full suite green
- [x] Unit: receiver-state issuance against a non-`Open` home channel (Challenged / Closing / Closed) produces unsigned rows in both `channel_v1` (transfer) and `app_session_v1` (release) flows; `Open` produces node-signed
- [x] Unit: `HandleHomeChannelCheckpointed` re-signs the off-chain head (highest stored version) when challenge clears; no-op when head is already signed; skips when head transition is not a receiver-credit type; no head-sig call on Open→Open
- [x] Unit + DB: `HomeChannelClosed` while `Challenged` emits one `ChallengeRescue` with `max(0, receives − sends)` above closure; zero-amount rescue still emitted to advance epoch (MF2-H02 regression); deterministic txID = `GetReceiverTransactionID(closedChannelID, newStateID)`; no rescue on Open→Closed (cooperative finalize)
- [x] DB: `SumNetTransitionAmountAfterVersion` — dust receives during challenge; signed pre-challenge receives stranded above closure included; HomeDeposit / finalize / acknowledgement / escrow / migrate excluded; adversarial rollback (negative net) returned; Commit deduction; strict `>` closure boundary; cross-epoch exclusion
- [x] Handler: negative-net clamp at caller
- [x] Concurrency: `HandleHomeChannelClosed`, `HandleHomeChannelChallenged`, `HandleHomeChannelCheckpointed` acquire `LockUserState` before status mutation / backfill — captured-flag ordering tests on `UpdateChannel`
- [x] Core: `NewChallengeRescueState` accepts amount ≥ 0 (rejects negative), rejects `prev.HomeChannelID == nil`; produces version=1, empty token/blockchain, balance credited

Postgres `FOR UPDATE` integration coverage for both lock orders (RPC wins vs event wins) deferred to a follow-up; the lock primitive is proven by `DBStore.LockUserState`'s own tests, and the call sites are pinned by the ordering tests above.


nitronode/api/channel_v1/submit_state_test.go
```diff
@@ -228,6 +229,131 @@ func TestSubmitState_TransferSend_Success(t *testing.T) {
 	mockTxStore.AssertExpectations(t)
 }
 
+func TestSubmitState_TransferSend_ReceiverHomeChannelChallenged_NoNodeSig(t *testing.T) {
+	// When the receiver's home channel is in Challenged status the node must persist the
+	// receiver state row but must NOT node-sign it. This prevents the dust-credit checkpoint
+	// reset described in security finding MF2-H01.
```
nksazonov: Tt is better to describe the vector of attack, rather than mentioning the finding by id, as we don't have a centralized database of security findings in the repo.

nitronode/event_handlers/service.go
```diff
@@ -214,14 +272,89 @@ func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx co
...
+	// the row at the closure version itself is the closing state (or the prior head) and
+	// must be excluded — only receiver credits issued strictly after the dispute version
+	// belong in the squash. When the sum is non-zero, those contributing rows imply the
+	// off-chain head fetched below sits at prev.Version > closureVersion.
+	total, err := tx.SumUnsignedReceiverStateAmountsAfterVersion(channel.ChannelID, closureVersion)
+	if err != nil {
+		return err
+	}
+	if total.IsNegative() {
+		return fmt.Errorf("unexpected negative challenge rescue sum: %s", total.String())
```
nksazonov: It took me a moment to realize why this is unexpected. The reason is obvious when the reader is in the context of the problem: "challenge rescue" squashes only the "receive" states, sum of which can not be negative.
I suggest mentioning explicitly that receive states were summed to ease the understanding when this log is seen out of the code context:

```suggestion
		return fmt.Errorf("unexpected negative challenge_rescue sum of receive states: %s", total.String())
```

nitronode/store/database/channel.go
```diff
@@ -167,6 +167,36 @@ func (s *DBStore) CheckActiveChannel(wallet, asset string) (string, *core.Channe
...
 
+// GetHomeChannelStatus returns the current status of a home channel by its ID.
+// Returns nil if no home channel with the given ID exists. The lookup is type-restricted
+// to home channels; callers asking about escrow or other channel kinds must use a
+// different accessor.
+func (s *DBStore) GetHomeChannelStatus(channelID string) (*core.ChannelStatus, error) {
+	channelID = strings.ToLower(channelID)
+	if !strings.HasPrefix(channelID, "0x") {
+		channelID = "0x" + channelID
+	}
```
nksazonov: There is already such code present in `GetChannelByID(...)`. It is worth extracting this code to `normalizeChannelId` helper somewhere, maybe to be also used outside of the DB component, so that channelId normalization is centralized.

pkg/core/types.go
```diff
@@ -711,7 +755,7 @@ func NewTransactionFromTransition(senderState *State, receiverState *State, tran
 	var toAccount, fromAccount string
 	// Transition validator is expected to make sure that all the fields are present and valid.
 
-	if transition.Type != TransitionTypeRelease && senderState == nil {
+	if transition.Type != TransitionTypeRelease && transition.Type != TransitionTypeChallengeRescue && senderState == nil {
 		return nil, fmt.Errorf("sender state must not be nil for non-release transitions")
```
nksazonov: I would update this error string to mention "challenge rescue" transition as well.

nitronode/event_handlers/service.go
```diff
@@ -114,14 +123,56 @@ func (s *EventHandlerService) HandleHomeChannelCheckpointed(ctx context.Context,
...
+// and the user must countersign and acknowledge it via the regular RPC flow.
+func (s *EventHandlerService) backfillOffChainHeadNodeSig(ctx context.Context, tx core.ChannelHubEventHandlerStore, channelID string) error {
+	head, err := tx.GetLastStateByChannelID(channelID, false)
+	if err != nil {
+		return err
+	}
+	if head == nil || head.NodeSig != nil {
+		return nil
+	}
+	packed, err := s.statePacker.PackState(*head)
```
nksazonov, line 187: The spec says "sign only the off-chain head — the highest-version queued *receive* state", but this function will sign any unsigned head regardless of transition type. In practice these are equivalent: user ops are blocked during a challenge (`CheckActiveChannel` returns nil for `Challenged` channels), so only receiver-type states can accumulate. But there's no defensive guard — if that upstream invariant breaks, a non-receiver state gets signed.

Consider adding a type check after the `head == nil` guard:
```go
if head.Transition.Type != core.TransitionTypeTransferReceive &&
   head.Transition.Type != core.TransitionTypeRelease {
    return nil
}
```

nitronode/api/channel_v1/handler.go
```diff
@@ -115,17 +115,31 @@ func (h *Handler) issueTransferReceiverState(ctx context.Context, tx Store, send
...
 		}
 
-		_nodeSig, err := h.nodeSigner.Sign(packedState)
-		if err != nil {
-			return nil, rpc.Errorf("failed to sign receiver state")
+		// While the receiver's home channel is challenged the node must not advance the
+		// channel head by node-signing further receiver states; the state row is still
+		// persisted (unsigned) so amounts accrued during the challenge can be reconciled
+		// when the dispute is resolved.
+		if channelStatus == nil || *channelStatus != core.ChannelStatusChallenged {
```
nksazonov: The denylist `!= Challenged` also signs states for `Closing` (status 3) channels. A `Closing` channel has a co-signed Finalize in-flight; signing new receiver states on it produces node-signed rows that can never reach on-chain settlement.

An allowlist is more explicit and safer:
```go
if channelStatus == nil ||
   *channelStatus == core.ChannelStatusVoid ||
   *channelStatus == core.ChannelStatusOpen {
    // sign
}
```

Not a regression from this PR — `Closing` was always signed — but worth tightening now that the check exists.

nitronode/api/app_session_v1/handler.go
```diff
@@ -164,19 +164,32 @@ func (h *Handler) issueReleaseReceiverState(ctx context.Context, tx Store, recei
...
 		}
 
-		nodeSig, err := h.signer.Sign(packedState)
-		if err != nil {
-			return rpc.Errorf("failed to sign receiver state: %v", err)
+		// While the receiver's home channel is challenged the node must not advance the
+		// channel head by node-signing further receiver states; the state row is still
+		// persisted (unsigned) so amounts accrued during the challenge can be reconciled
+		// when the dispute is resolved.
+		if channelStatus == nil || *channelStatus != core.ChannelStatusChallenged {
```
nksazonov: Same denylist issue as the `channel_v1` counterpart — `!= Challenged` also signs for `Closing` channels. Consider switching to the `Void || Open` allowlist for consistency with the spec and with any future fix to the `channel_v1` path.

nitronode/store/database/state.go
```diff
@@ -221,24 +221,59 @@ func (s *DBStore) GetLastStateByChannelID(channelID string, signed bool) (*core.
...
+		Total decimal.Decimal `gorm:"column:total"`
+	}
+	err := s.db.Raw(`
+		SELECT COALESCE(SUM(transition_amount), 0) AS total
+		FROM channel_states
+		WHERE home_channel_id = ?
+			AND version > ?
+			AND node_sig IS NULL
+			AND transition_type IN (?, ?)
+	`, cid, minVersion, uint8(core.TransitionTypeTransferReceive), uint8(core.TransitionTypeRelease)).Scan(&row).Error
```
nksazonov: No `epoch` filter. Channel IDs are unique per nonce so cross-epoch contamination can't occur in practice, but a defensive `AND epoch = ?` (using `prev.Epoch` from the caller's scope, already available in `issueChallengeRescue`) would make that invariant explicit and guard against any DB inconsistency that reused a channel ID.

nitronode/event_handlers/service_test.go
```diff
@@ -92,7 +95,9 @@ func TestHandleHomeChannelCheckpointed_Success(t *testing.T) {
 			ch.StateVersion == 5
 	})).Return(nil)
 	mockStore.On("RefreshUserEnforcedBalance", userWallet, "usdc").Return(nil)
-	mockStore.On("UpdateStateUserSigIfMissing", channelID, uint64(5), "").Return(nil)
+	mockStore.On("UpdateStateSigsIfMissing", channelID, uint64(5), "", "").Return(nil)
+	// Off-chain head missing → no head-sig backfill.
+	mockStore.On("GetLastStateByChannelID", channelID, false).Return(nil, nil)
```
nksazonov, line 101: The test creates the service as `&EventHandlerService{}` (nil `nodeSigner`, nil `statePacker`), but the channel fixture has `Status: ChannelStatusChallenged`, so `backfillOffChainHeadNodeSig` is reached. It avoids a nil panic only because the mock returns `nil` here — if this mock were updated to return a real unsigned state, `s.statePacker.PackState(*head)` would panic.

`newTestEventHandlerService(t)` was added in this PR exactly for this scenario. Using it here would make the test robust against future mock changes:
```go
service, _ := newTestEventHandlerService(t)
```

nksazonov (commented):
Nice work! The three security remediations described in the added docs are all correctly implemented, well-tested, and match the spec. All findings below are informational.

nitronode/api/channel_v1/submit_state_test.go
```diff
@@ -228,6 +229,129 @@ func TestSubmitState_TransferSend_Success(t *testing.T) {
...
+		nodeSigner:       nodeSigner,
+		nodeAddress:      nodeAddress,
+		minChallenge:     minChallenge,
+		metrics:          metrics.NewNoopRuntimeMetricExporter(),
+		maxSessionKeyIDs: 256,
+		actionGateway:    &MockActionGateway{},
+	}
+
+	userSigner := NewMockSigner()
+	userWalletSigner, _ := core.NewChannelDefaultSigner(userSigner)
```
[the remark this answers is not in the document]
philanton: Yep, fixed in [37341075](https://github.com/layer-3/nitrolite/pull/759/changes/373410754c5d3c557ea0859b5faa4d9598d87958).

nitronode/event_handlers/service_test.go
```diff
@@ -1114,6 +1120,360 @@ func TestHandleHomeChannelCheckpointed_BackfillError(t *testing.T) {
...
+	mockStore.On("UpdateStateSigsIfMissing", channelID, closureVersion, "", "").Return(nil)
+
+	err := service.HandleHomeChannelClosed(ctx, mockStore, event)
+	require.NoError(t, err)
+
+	mockStore.AssertExpectations(t)
+	mockStore.AssertNotCalled(t, "SumUnsignedReceiverStateAmountsAfterVersion", mock.Anything, mock.Anything)
+	mockStore.AssertNotCalled(t, "StoreUserState", mock.Anything, mock.Anything)
+	mockStore.AssertNotCalled(t, "RecordTransaction", mock.Anything, mock.Anything)
+}
```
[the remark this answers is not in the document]
philanton: Comment rewritten to match the `_OpenChannel_NoRescue` scenario in [c830e9d](https://github.com/layer-3/nitrolite/pull/759/changes/c830e9d74135df271f6c24759ff773ff15555aff).

nitronode/event_handlers/service.go
```diff
@@ -214,14 +272,83 @@ func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx co
...
+	prev, err := tx.GetLastStateByChannelID(channel.ChannelID, false)
+	if err != nil {
+		return err
+	}
+	if prev == nil {
+		// Should not happen: receiver-state rows contributed to the sum above, so at
+		// least one state must exist. Surface the inconsistency rather than silently
+		// dropping the rescue.
+		return fmt.Errorf("no state found for closed challenged channel %s while pending receiver credits exist", channel.ChannelID)
+	}
```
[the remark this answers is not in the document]
philanton: Yep, good points on both.

The strict `>` is intentional — the row at `closureVersion` is the closing state itself (backfilled signed earlier in `HandleHomeChannelClosed` via `UpdateStateSigsIfMissing`), so it can never be both onchain-enforced AND rescue-contributing. Added a comment to that effect plus the `prev.Version > closureVersion` assertion (when net is non-zero) in [19b45715](https://github.com/layer-3/nitrolite/pull/759/changes/19b45715a795e347f5081e155a7df220ffbe3654).

Later [09f6d356](https://github.com/layer-3/nitrolite/pull/759/changes/09f6d35695f64c02284e9a478b81d2d6c0fc9ec0) broadened the rescue scope to net (receives + releases − sends − commits) above closure, so the assertion now reads `prev.Version > closureVersion` against the clamped total. Same semantics.

nitronode/api/channel_v1/submit_state_test.go
philanton: Yep, done in [6f38d219](https://github.com/layer-3/nitrolite/pull/759/changes/6f38d219a18c7e17c3bbfb7bd10e7e7deb6b7708).

nitronode/event_handlers/service.go
philanton: Initially updated the error string per your suggestion in [a6cd0148](https://github.com/layer-3/nitrolite/pull/759/changes/a6cd01480a52f42d6f9dae4ae596ef8f1d9dbe74).

Then [09f6d356](https://github.com/layer-3/nitrolite/pull/759/changes/09f6d35695f64c02284e9a478b81d2d6c0fc9ec0) broadened the rescue to include sends + commits as debits, which makes negative net reachable (and expected) when the user adversarially rolls back her own sends/commits. The hard error became a clamp-to-zero + warn log there — onchain has already paid the user above the off-chain head value in that case, so the rescue must not dock further.

nitronode/store/database/channel.go
philanton: I would leave this for a follow-up. The helper itself is fine, but the rabbit hole goes deeper than the DB layer — `core.NormalizeHexAddress` already exists for hex normalization with validation, and there are inline `strings.ToLower` callsites in different contexts (channel IDs, wallet addresses, asset symbols) that probably want different rules. Worth doing as its own pass.

pkg/core/types.go
philanton: Yep, done in [a6cd0148](https://github.com/layer-3/nitrolite/pull/759/changes/a6cd01480a52f42d6f9dae4ae596ef8f1d9dbe74).

nitronode/event_handlers/service.go
philanton: Good point. Added the receiver-type guard in [16db3434](https://github.com/layer-3/nitrolite/pull/759/changes/16db3434fd4cd550a9898c4bcb636a7b622943ec) — `Debug` log rather than `Warn` since the upstream invariant (no user-initiated ops while `Challenged`) is what makes this defensive; if it ever fires we want to see it, but it is not noisy-by-default territory.

nitronode/api/channel_v1/handler.go
philanton: Yep, switched to an Open-only allowlist via `CheckActiveChannel` (which already returns nil status for anything `> Open`) in [16db3434](https://github.com/layer-3/nitrolite/pull/759/changes/16db3434fd4cd550a9898c4bcb636a7b622943ec). Dropped `GetHomeChannelStatus` end-to-end as a side effect — the new check needs no extra DB method.

nitronode/api/app_session_v1/handler.go
philanton: Same fix as the `channel_v1` side, applied in [16db3434](https://github.com/layer-3/nitrolite/pull/759/changes/16db3434fd4cd550a9898c4bcb636a7b622943ec).

nitronode/store/database/state.go
philanton: Yep, added `AND epoch = ?` with `prev.Epoch` from the caller in [19b45715](https://github.com/layer-3/nitrolite/pull/759/changes/19b45715a795e347f5081e155a7df220ffbe3654). Added a DB-level test case for cross-epoch exclusion too.

Heads up — [09f6d356](https://github.com/layer-3/nitrolite/pull/759/changes/09f6d35695f64c02284e9a478b81d2d6c0fc9ec0) broadened the predicate further (dropped `node_sig IS NULL`, added sends + commits as debits) and renamed the method to `SumNetTransitionAmountAfterVersion`. Epoch filter is still there.

nitronode/event_handlers/service_test.go
philanton: Yep, swapped to `newTestEventHandlerService(t)` in [a6cd0148](https://github.com/layer-3/nitrolite/pull/759/changes/a6cd01480a52f42d6f9dae4ae596ef8f1d9dbe74).

nksazonov (approved):

nitronode/event_handlers/service.go
```diff
@@ -169,7 +236,7 @@ func (s *EventHandlerService) HandleHomeChannelChallenged(ctx context.Context, t
 		return err
 	}
 
-	if err := tx.UpdateStateUserSigIfMissing(event.ChannelID, event.StateVersion, event.UserSig); err != nil {
+	if err := tx.UpdateStateSigsIfMissing(event.ChannelID, event.StateVersion, event.UserSig, ""); err != nil {
```
ihsraham, line 261: [P1/blocking] This handler needs to take the same `LockUserState(channel.UserWallet, channel.Asset)` lock before mutating the channel to `Challenged`.

The receiver issuance paths already lock that row, then call `CheckActiveChannel` and may attach `NodeSig` when they still see `Open`. Without the same lock here, an in-flight transfer receive or app-session release can read `Open`, sign the receiver state, and commit after this challenge event, leaving a node-signed higher-version receiver state for a disputed channel.

I would make challenge entry use the same serialization pattern as `HandleHomeChannelClosed`: load the channel, lock the user/asset row, then update status/backfill signatures while still inside the transaction.

nitronode/event_handlers/service.go
```diff
@@ -114,14 +125,70 @@ func (s *EventHandlerService) HandleHomeChannelCheckpointed(ctx context.Context,
...
 		return err
 	}
 
+	// When a challenge is cleared, the off-chain head may sit above event.StateVersion:
+	// any receiver state issued during the challenge was stored unsigned and is now the
+	// channel's actual latest state. Backfill the node signature on that head so future
+	// flows treat it as fully co-signed. On normal Open→Open checkpoints the head row
+	// is already node-signed via the RPC path and this is a no-op.
+	if wasChallenged {
```
ihsraham, line 150: [P2] Challenge clearance should also take the user-state lock before flipping back to `Open` and calling `backfillOffChainHeadNodeSig`.

Otherwise a receiver RPC can read `Challenged`, choose to store an unsigned receiver state, then this handler can backfill the old head and commit before that RPC stores a newer unsigned head. The channel is now open, but the latest local head remains unsigned.

One way to close this is to lock `(channel.UserWallet, channel.Asset)` before `UpdateChannel` and the backfill. Then the RPC either commits before the backfill and is included, or waits and sees `Open` and signs normally.

ihsraham (changes requested):
I would hold approval for now. The receiver-state signing rule looks right in the steady state, but challenge enter/clear is not serialized with the receiver issuance path. That leaves a race where a receive/release RPC can make a node-signed state after the challenge is observed onchain but before the local DB status update is visible, which is enough to recreate the H-01 reset precondition.

nitronode/event_handlers/service.go
philanton: Yep, good catch — same race class as the one [24cdaff1](https://github.com/layer-3/nitrolite/pull/759/changes/24cdaff14ad0a52f0e88f5b1d2bfd16cd6b1f12d) closed for `HandleHomeChannelClosed`. Fixed in [8d914687](https://github.com/layer-3/nitrolite/pull/759/changes/8d914687f634304fa6ae63457ae21da427c26857): `LockUserState(channel.UserWallet, channel.Asset)` acquired right after the nil/type guards, before any mutation. RPC paths block until the challenge event commits and then re-check Status via `CheckActiveChannel`.

Added `TestHandleHomeChannelChallenged_AcquiresUserLockBeforeMutation` to pin the ordering via a captured-flag guard on `UpdateChannel`)

nitronode/event_handlers/service.go
philanton: Yep, same fix applied to `HandleHomeChannelCheckpointed` in [8d914687](https://github.com/layer-3/nitrolite/pull/759/changes/8d914687f634304fa6ae63457ae21da427c26857). `LockUserState` goes in after the nil/type guards and before `UpdateChannel` + `backfillOffChainHeadNodeSig`, so an RPC that read Status=Challenged either commits its unsigned row before us (and the backfill picks it as the head) or blocks until we set Status=Open and then sees that via its own re-check.

Took the lock unconditionally rather than only on `wasChallenged` — the Open→Open path still mutates `StateVersion` and calls `UpdateStateSigsIfMissing`, and uniform pattern across the three handlers is easier to reason about. Added `TestHandleHomeChannelCheckpointed_AcquiresUserLockBeforeMutation` to pin the ordering)

ihsraham (approved):
Re-reviewed the latest head, the blocking race is addressed and focused checks pass.

PR #761, opened 2026-05-14
MF2-M01: fix(sdk): use keccak256(utf8) for session key application ID hashing

nksazonov:
## Problem

`PackAppSessionKeyStateV1` encoded application and app-session IDs by treating each string as hex (`common.HexToHash` in Go, a matching `hexToBytes32` in TypeScript). Any string containing a non-hex character was silently collapsed to the same `0x000...000` bytes32, so `"app-1"` and `"app-2"` — and any other human-readable identifiers — produced **identical signing payloads**. This opened a phishing vector: a signature collected for Integration A's `"app-1"` could be submitted unchanged to authorize Integration B's `"app-2"`.

## Fixes

- **Go** (`pkg/app/session_key_v1.go`): replace `common.HexToHash(id)` → `crypto.Keccak256Hash([]byte(id))` for both `applicationIDHashes` and `appSessionIDHashes`.
- **TypeScript** (`sdk/ts/src/app/packing.ts`): remove the broken `hexToBytes32` helper; replace with `(id) => keccak256(toHex(id))`. Remove now-unused `pad` import.
- **Drift vectors** updated in Go test and TS drift test (new hash: `0x6d404fa6...`). Go and TypeScript produce identical output, confirming cross-SDK consistency.
- **Collision tests** added in both languages covering `ApplicationIDs` and `AppSessionIDs` with human-readable IDs (`"app-1"`, `"app-2"`, `"trading"`, `"gaming"`, `"defi"`) and boundary cases (empty string vs whitespace).

## Notes

This is a **breaking change**: existing signed session key states are invalidated. Clients must re-sign with the updated SDK after the server deploys this change.

### DIFF
--- a/nitronode/api/app_session_v1/README.md
+++ b/nitronode/api/app_session_v1/README.md
@@ -706,6 +706,8 @@ ORDER BY created_at DESC;
 - `version` must be greater than 0
 - `expires_at` must be in the future
 - `user_sig` is required
+- `application_ids` entries must be lowercase strings (non-lowercase values are rejected before signature verification)
+- `app_session_ids` entries must be lowercase strings (non-lowercase values are rejected before signature verification)
 - Version must be sequential (latest_version + 1)
 - Signature must recover to `user_address`
 - Newly registered keys count against the per-user cap (`NITRONODE_MAX_SESSION_KEYS_PER_USER`, default 100). Updates to keys that already exist for the user are not blocked by the cap.
@@ -715,6 +717,7 @@ ORDER BY created_at DESC;
 **Signature Verification**:
 - Uses ABI encoding via `PackAppSessionKeyStateV1` to create a deterministic hash
 - Encodes: user_address (address), session_key (address), version (uint64), application_ids (bytes32[]), app_session_ids (bytes32[]), expires_at (uint64)
+- Each application_id and app_session_id string is converted to bytes32 via `keccak256(utf8(id))` before ABI encoding
 - The `user_sig` field is excluded from packing (it is the signature itself)
 - Recovers signer address from ECDSA signature
 - Validates that recovered address matches `user_address`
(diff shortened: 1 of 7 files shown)


nksazonov:
@coderabbitai review

philanton (approved):

ihsraham:
P2, fix before public/audit-closure hardening.

The `keccak256(utf8(id))` change looks like the right fix for the original `HexToHash` collision. I was too strong calling this a blocker for this PR if this is a fresh, pre-public deployment with no legacy app-session-key state.

One thing is still worth tightening before this is treated as final/public: the signed bytes and persisted scope should use the same canonical form. Today `SubmitSessionKeyState` validates signatures over the exact submitted `application_ids` / `app_session_ids`, then `StoreAppSessionKeyState` lowercases those relation rows (`pkg/app/session_key_v1.go:125`, `pkg/app/session_key_v1.go:130`, `nitronode/store/database/app_session_key_state.go:80`, `nitronode/store/database/app_session_key_state.go:93`).

So a signature over `App-1` can be accepted and later enforced as `app-1`. If app IDs are intended to be lowercase-only, the clean fix is to reject non-canonical `application_ids` and `app_session_ids` before signature validation. Alternatively, define one canonicalization path and use it consistently in Go packing, TS packing, submit validation, storage, and docs.

Non-blocking follow-ups: add a Go-generated human-readable drift vector for `app-1` / `app-2`, and update docs to say the strings are hashed as `keccak256(utf8(id))` before ABI encoding as `bytes32[]`.

ihsraham (approved):
Approving for M-01 

The PR makes the needed `keccak256(utf8(id))` change in both Go and TS, and the targeted Go/TS tests pass. I left the canonicalization mismatch as a P2 hardening item before public/audit-final closure, not as a blocker for this PR.

PR #762, opened 2026-05-14
MF2-C01: preserve post-Finalize closing marker across challenge cycle

philanton:
### C-01: post-Finalize submission gate erased across challenge cycle

When `SubmitState()` signs a `TransitionTypeFinalize` state, it updates the channel status to `ChannelStatusClosing`. `CheckActiveChannel()` then excludes `Closing` channels because it only returns home channels with `status <= ChannelStatusOpen`, preventing further user-initiated state submissions.

However, the event handlers can later regress the local lifecycle status. `HandleHomeChannelChallenged()` unconditionally sets `channel.Status = core.ChannelStatusChallenged`, even if the channel was already marked `ChannelStatusClosing` because the node signed a newer finalized state. If a later `ChannelCheckpointed` event is processed, `HandleHomeChannelCheckpointed()` changes `ChannelStatusChallenged` back to `ChannelStatusOpen`.

This allows the `Closing` marker to be erased by delayed onchain events. For example, a user can checkpoint state `N`, obtain signed states `N + 1` and `N + 2`, challenge onchain with `N + 1`, and immediately request a finalized offchain state `N + 3`. The node marks the channel `Closing`. When the `ChannelChallenged` event for `N + 1` is processed, the DB status becomes `Challenged`. When the user checkpoints `N + 2` onchain, the `ChannelCheckpointed` event moves the DB status back to `Open`.

At that point, the database again contains an active/open channel even though the latest offchain state signed by the node is finalized. The original post-finalize path can become reachable again: `SubmitState()` can accept a new transition after the finalized state because `CheckActiveChannel()` sees an `Open` channel, and `StateAdvancerV1.ValidateAdvancement()` can advance from the finalized state into a new epoch. This undermines the remediation for the critical post-finalize funds-theft issue.

**Fix.** On `ChannelCheckpointed`, when clearing `Challenged`, derive the post-Finalize state from `channel_states` via a new `HasSignedFinalize` lookup. Restore `Closing` instead of `Open` when a node-signed Finalize row exists for the channel. The status field carries onchain reality while the dispute is live; the binding offchain Finalize fact is shadowed in `channel_states` and reapplied on resolution. No new `ChannelStatus` enum value, no DB migration, no API/SDK fan-out.

A second vector through the same regression — `HandleHomeChannelCreated` replaying over an initialized channel (reorg / indexer restart re-fires `Created` over a current `Closing` and erases the Finalize marker) — is guarded by a `Status >= ChannelStatusOpen` short-circuit with a warning log.

### L-02: stale `ChallengeExpiresAt` after challenge resolution

`HandleHomeChannelCheckpointed()` clears a challenged home channel by changing `channel.Status` from `core.ChannelStatusChallenged` to `core.ChannelStatusOpen`, but it does not clear `channel.ChallengeExpiresAt`. Since `HandleHomeChannelChallenged()` persists `channel.ChallengeExpiresAt` when a challenge begins, a later checkpoint that resolves the challenge can leave the local database with an open channel that still reports a stale challenge expiry.

This does not appear to block normal state submission, because `CheckOpenChannel()` uses the channel status rather than `ChallengeExpiresAt` when deciding whether a home channel is open. However, the stale timestamp is exposed through the channel API and can mislead clients, dashboards, or operators into believing an open channel still has an active or historical challenge deadline attached to it.

**Fix.** `HandleHomeChannelCheckpointed()` now sets `channel.ChallengeExpiresAt = nil` alongside the status flip when clearing `Challenged`. `HandleHomeChannelClosed()` mirrors the same clear on terminal close so a lingering expiry never trails a closed channel.

### I-01: misleading `HandleNodeBalanceUpdated` doc comment

The comment for `HandleNodeBalanceUpdated()` states that the handler updates the user's staked balance for the specified blockchain. The implementation actually calls `SetNodeBalance()` with `event.BlockchainID`, `event.Asset`, and `event.Balance`, which stores the node's onchain liquidity metric under `node_balance`. This can mislead maintainers into thinking the event affects user staking state, while it only updates observability data for node liquidity.

**Fix.** Corrected the doc comment on `HandleNodeBalanceUpdated` to describe the actual semantics: it upserts the node's onchain liquidity for the given blockchain/asset pair, not user staking state.

## Test plan

- [x] `go vet ./...`
- [x] `go test ./...`
- [x] New unit test `TestHandleHomeChannelCheckpointed_FromChallengedWithSignedFinalize` exercises the C-01 audit scenario: status restored to `Closing` when a signed Finalize exists.
- [x] New integration test `TestDBStore_HasSignedFinalize` covers the storage primitive (present / absent / present-but-unsigned).
- [x] Existing `wasChallenged`-path tests updated to mock the new lookup and assert `ChallengeExpiresAt == nil` after resolution (L-02 coverage).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of challenged channels when disputes are resolved through checkpoints; channels now correctly transition to "Closing" if a finalization has been signed, or "Open" otherwise.
  * Fixed channel state tracking to properly clear challenge expiry information when channels transition to closed or reopened states.

[Review Change Stack](https://app.coderabbit.ai/change-stack/layer-3/nitrolite/pull/762)


nksazonov:
@coderabbitai review

nitronode/store/database/state.go
```diff
@@ -250,6 +250,25 @@ func (s *DBStore) UpdateStateSigsIfMissing(channelID string, version uint64, use
 	return nil
 }
 
+// HasSignedFinalize reports whether a Finalize state exists for the given home channel.
+// Used by event handlers to detect the post-Finalize lifecycle without loading the state
+// row: the channels.status field can be temporarily overwritten by Challenged on a stale
+// on-chain challenge, but the Finalize row persists here and is the authoritative marker.
+// Finalize rows are only ever written via SubmitState after the node signs, so a plain
+// existence check is equivalent to checking node_sig IS NOT NULL.
+func (s *DBStore) HasSignedFinalize(channelID string) (bool, error) {
```
nksazonov, line 259: If I remember correctly from the previous PR, the `channelId` has to be normalized (lower-cased, "0x"-prefixed) first before usage. This can be done by using the normalization helper I have suggested to add in the previous PR.

nitronode/store/database/state.go
```diff
@@ -250,6 +250,25 @@ func (s *DBStore) UpdateStateSigsIfMissing(channelID string, version uint64, use
...
+	var count int64
+	err := s.db.Model(&State{}).
+		Where("home_channel_id = ? AND transition_type = ?",
+			strings.ToLower(channelID), uint8(core.TransitionTypeFinalize)).
```
nksazonov: The query does not include `AND node_sig IS NOT NULL`. The function comment claims this is equivalent because `SubmitState` always sets `NodeSig` before calling `StoreUserState` — but this is a business-logic invariant, not a DB-level constraint. `StoreUserState` accepts `NodeSig = nil` without error. Any future path that stores a `TransitionTypeFinalize` row before the node signs (crash recovery, a future refactor, a test fixture) would cause `HasSignedFinalize` to return `true` incorrectly, restoring `ChannelStatusClosing` for a channel where the node has not actually committed to finalization.

Suggest adding `AND node_sig IS NOT NULL` to the `Where` clause. One line, defense-in-depth — the invariant is already true, this just makes it explicit.

nitronode/store/database/db_store_test.go
```diff
@@ -1125,6 +1125,82 @@ func TestDBStore_SumUnsignedReceiverStateAmountsAfterVersion(t *testing.T) {
...
+		setupChannel(t, store)
+
+		storeState(t, store, 6, core.TransitionTypeTransferReceive, true)
+		storeState(t, store, 7, core.TransitionTypeTransferReceive, true)
+
+		got, err := store.HasSignedFinalize(homeChannelID)
+		require.NoError(t, err)
+		assert.False(t, got)
+	})
+}
```
nksazonov, line 1338: A third case is missing: *Finalize row present, `node_sig IS NULL` → should return `false`*. This is the direct specification of the invariant documented in `HasSignedFinalize`'s comment. Without it, a regression that omits `AND node_sig IS NOT NULL` from the query (or a future path that stores an unsigned Finalize) goes undetected.

Suggest adding:
```go
t.Run("False when Finalize exists without node sig", func(t *testing.T) {
    db, cleanup := SetupTestDB(t)
    defer cleanup()
    store := NewDBStore(db)
    setupChannel(t, store)
    storeState(t, store, 7, core.TransitionTypeFinalize, false) // no node sig
    got, err := store.HasSignedFinalize(homeChannelID)
    require.NoError(t, err)
    assert.False(t, got)
})
```

nitronode/event_handlers/service.go
```diff
@@ -111,7 +117,22 @@ func (s *EventHandlerService) HandleHomeChannelCheckpointed(ctx context.Context,
...
+		// Reconstruct the post-Finalize Closing marker from channel_states: if the node
+		// has already signed a Finalize state for this channel, the off-chain close is
+		// still pending and the channel must not return to Open. See the doc comment on
+		// HandleHomeChannelChallenged for the round-trip rationale.
+		finalized, err := tx.HasSignedFinalize(chanID)
+		if err != nil {
+			return err
+		}
+		if finalized {
+			channel.Status = core.ChannelStatusClosing
```
nksazonov, line 152: `channel.StateVersion` is set unconditionally at line 116 to `event.StateVersion` (the checkpoint version, e.g., N+2). After this branch the `channels` row has `StateVersion = N+2` while the authoritative off-chain state is the Finalize at N+3. This is not a security issue — `Status = Closing` guards `CheckActiveChannel` correctly — but it creates an observable inconsistency.

Concretely: a subsequent `HandleHomeChannelChallenged` uses `channel.StateVersion` as its staleness threshold. With `StateVersion = N+2`, a re-challenge at exactly version N+2 passes the `event.StateVersion < channel.StateVersion` guard and is processed; N+3 as the reference would have correctly rejected it as stale.

Suggest returning the Finalize version from `HasSignedFinalize` (e.g., change the signature to `(uint64, bool, error)`) and updating `channel.StateVersion` to the Finalize version when restoring `Closing` here.

nksazonov (changes requested):
Solid fix for the core `Closing → Challenged → Open` regression — the `HasSignedFinalize` lookup in `HandleHomeChannelCheckpointed` is the right mechanism. Two concerns require attention before landing: a gap in query defensiveness (see inline comments) and a structurally identical regression path via event replay that this fix does not cover.

---

**Out-of-diff: `HandleHomeChannelCreated` can regress `Closing → Open` via event replay (`service.go:66`)**

`HandleHomeChannelCreated` unconditionally sets `channel.Status = core.ChannelStatusOpen`. If the `HomeChannelCreated` on-chain event is replayed — chain reorg, indexer restart, or block reprocessing — a channel already in `ChannelStatusClosing` is reset to `Open`, re-arming `CheckActiveChannel` and the submission gate. This is structurally the same regression path this PR closes in `HandleHomeChannelCheckpointed`. The system already treats replay as a real concern (`UpdateStateSigsIfMissing` is explicitly idempotent). The same `HasSignedFinalize` guard, or a simpler `if channel.Status >= core.ChannelStatusClosing { return nil }` early-out, should be applied here.

---

**Out-of-diff: `backfillOffChainHeadNodeSig` runs unconditionally after restoring `Closing` (`service.go:158–159`)**

`backfillOffChainHeadNodeSig` is called in the `wasChallenged` branch regardless of the `HasSignedFinalize` result. In practice this is safe — the Finalize state is the head and already has `NodeSig != nil`, so the function exits early. But the intent is non-obvious: a reader must trace two functions to confirm no post-Finalize state ever receives a node signature here. The `finalized` variable is already in scope; gating the call on `!finalized` makes the invariant explicit and eliminates the silent dependency on the Finalize always being the head.

nitronode/store/database/state.go
philanton: I cannot agree, actually, if we don't rely on business logic, then user sig, blockchain id, token address etc. should be in this filter as well)

nitronode/event_handlers/service.go
philanton: Sorry, but it sounds wrong. `event.StateVersion < channel.StateVersion` guard is correct, and writing N+3 version into `channel.StateVersion` is not correct, because it's not enforced on-chain at that moment

nitronode/store/database/state.go
philanton: Hmmm, the surrounding queries here (`UpdateStateSigsIfMissing`, `SumUnsignedReceiverStateAmountsAfterVersion`) all use bare `strings.ToLower` too — I'd rather not introduce a helper on this PR for just one call site. Happy to extract a `normalizeChannelID` across the whole `state.go` in a follow-up, since it'd touch ~5 places.

nitronode/store/database/state.go
philanton: Hmmm, you're right — even if `SubmitState` enforces it today, the function is literally called `HasSignedFinalize`, so the predicate should match the name. Tightened in [5b0ebc7](https://github.com/layer-3/nitrolite/pull/762/changes/5b0ebc763c226acaf1cae13f4494108fee7eb503), also dropped the no-op `Limit(1)` while there.

nitronode/store/database/db_store_test.go
philanton: Yep, added in [5b0ebc7](https://github.com/layer-3/nitrolite/pull/762/changes/5b0ebc763c226acaf1cae13f4494108fee7eb503) alongside the predicate change.

philanton:
Thanks for catching the `HandleHomeChannelCreated` replay path — same regression vector through a different door (Void→Open is supposed to be one-shot; reorg/indexer-restart re-fires Created over a current `Closing` and erases the Finalize marker). Guarded with `Status >= ChannelStatusOpen` + warning log in [f4a32bc](https://github.com/layer-3/nitrolite/pull/762/changes/f4a32bcd489134012eeaa4d76bdad4eeb23cce5b), with a table-driven test across Open/Challenged/Closing/Closed.

On gating `backfillOffChainHeadNodeSig` on `!finalized`: I'd leave it. The function's invariant is "node-sign the head if missing", not "node-sign unless we're in Closing" — folding the Finalize semantics into the caller spreads the lifecycle concern. The existing `head.NodeSig != nil` early-return already covers the Finalize-as-head case explicitly, and gating on `finalized` would still call `GetLastStateByChannelID` in the non-finalized branch, so the perf delta is zero.

nksazonov (approved):
Great work!

ihsraham (approved):
Approving after clarification. The future C-01 path this PR targets appears covered.

Given the only current environment is stress, and stress does not contain the `Open + signed Finalize` edge state, I do not see a migration or repair blocker here.

The `Open + signed Finalize` case can stay as a P2 hardening follow-up for corrupted or future persistent state.

PR #763, opened 2026-05-15
MF2-L01: gate transfer_send during channel creation

philanton:
## Summary
- `RequestCreation()` processed `TransitionTypeTransferSend` without calling `actionGateway.AllowAction`, letting a user bypass the 24h transfer allowance by piggybacking a transfer on initial channel creation (positive offchain balance from a prior `transfer_receive`, no active home channel).
- Mirror `SubmitState()` by gating the incoming transition at the top of the tx, before any state is locked, signed, stored, or receiver-side state is issued. Empty `GatedAction` (Acknowledgement / HomeDeposit / HomeWithdrawal) remains a no-op via `ActionGateway.AllowAction` map lookup.
- Added `TestRequestCreation_ActionGatewayRejection` covering the rejection path and asserting no Lock/Get/Create/Store/Record side effects occur.

## Test plan
- [x] `go test ./nitronode/api/channel_v1/...`
- [x] `go test ./nitronode/action_gateway/...`
- [x] `go vet ./nitronode/...`

## Summary by CodeRabbit

* **Bug Fixes**
  * Channel creation requests are now validated against action permissions before processing, preventing creation when conditions such as transfer allowances are not met.

[Review Change Stack](https://app.coderabbit.ai/change-stack/layer-3/nitrolite/pull/763)

### DIFF
--- a/nitronode/api/channel_v1/request_creation.go
+++ b/nitronode/api/channel_v1/request_creation.go
@@ -83,6 +83,14 @@ func (h *Handler) RequestCreation(c *rpc.Context) {
 
 	var nodeSig string
 	err = h.useStoreInTx(func(tx Store) error {
+		// Gate the incoming transition through the action gateway before any state
+		// is signed, stored, or receiver-side state is issued. Mirrors SubmitState so
+		// gated actions (e.g. transfer_send) cannot bypass the 24h allowance by
+		// piggybacking on channel creation.
+		if err := h.actionGateway.AllowAction(tx, incomingState.UserWallet, incomingState.Transition.Type.GatedAction()); err != nil {
+			return rpc.NewError(err)
+		}
+
 		_, err := tx.LockUserState(incomingState.UserWallet, incomingState.Asset)
 		if err != nil {
 			return rpc.Errorf("failed to lock user state: %v", err)
--- a/nitronode/api/channel_v1/request_creation_test.go
+++ b/nitronode/api/channel_v1/request_creation_test.go
@@ -2,6 +2,7 @@ package channel_v1
 
 import (
 	"context"
+	"errors"
 	"strconv"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


nitronode/api/channel_v1/request_creation.go
```diff
@@ -83,6 +83,14 @@ func (h *Handler) RequestCreation(c *rpc.Context) {
 
 	var nodeSig string
 	err = h.useStoreInTx(func(tx Store) error {
+		// Gate the incoming transition through the action gateway before any state
+		// is signed, stored, or receiver-side state is issued. Mirrors SubmitState so
+		// gated actions (e.g. transfer_send) cannot bypass the 24h allowance by
+		// piggybacking on channel creation.
```
nksazonov, line 89: `AllowAction` reads and records the action count before the per-user serialization lock is held. Two concurrent `RequestCreation` calls for the same wallet can both read the count, both pass the check, and both proceed — temporarily exceeding the 24h allowance. This race also exists in `SubmitState`, so it's not a regression, but the new call site inherits it. The fix is to move `LockUserState` before `AllowAction`, or have the action-count update acquire its own row-level lock.

nitronode/api/channel_v1/request_creation_test.go
```diff
@@ -620,3 +621,103 @@ func TestRequestCreation_NonClosedChannelRejection(t *testing.T) {
...
+	assert.Contains(t, respErr.Error(), "transfer allowance exhausted")
+
+	// Gate fires before any state effect: nothing should be locked, stored, or recorded.
+	mockTxStore.AssertNotCalled(t, "LockUserState", mock.Anything, mock.Anything)
+	mockTxStore.AssertNotCalled(t, "HasNonClosedHomeChannel", mock.Anything, mock.Anything)
+	mockTxStore.AssertNotCalled(t, "GetLastUserState", mock.Anything, mock.Anything, mock.Anything)
+	mockTxStore.AssertNotCalled(t, "CreateChannel", mock.Anything)
+	mockTxStore.AssertNotCalled(t, "StoreUserState", mock.Anything, mock.Anything)
+	mockTxStore.AssertNotCalled(t, "RecordTransaction", mock.Anything, mock.Anything)
+}
```
nksazonov, line 882: There's no test for a `TransferSend` initial state that succeeds through the gateway. Without a passing-path test, a regression that accidentally blocks all `TransferSend` channel creations — or silently skips the gate — won't be caught. `TestRequestCreation_Success` exercises `HomeDeposit` but not `TransferSend`. Add `TestRequestCreation_TransferSend_Success` mirroring the success test but with a `TransferSend` transition and receiver-side mock wiring (`StoreUserState` for receiver, `RecordTransaction`).

nitronode/api/channel_v1/request_creation_test.go
```diff
@@ -620,3 +621,103 @@ func TestRequestCreation_NonClosedChannelRejection(t *testing.T) {
...
+	respErr := ctx.Response.Error()
+	require.Error(t, respErr)
+	assert.Contains(t, respErr.Error(), "transfer allowance exhausted")
+
+	// Gate fires before any state effect: nothing should be locked, stored, or recorded.
+	mockTxStore.AssertNotCalled(t, "LockUserState", mock.Anything, mock.Anything)
+	mockTxStore.AssertNotCalled(t, "HasNonClosedHomeChannel", mock.Anything, mock.Anything)
+	mockTxStore.AssertNotCalled(t, "GetLastUserState", mock.Anything, mock.Anything, mock.Anything)
+	mockTxStore.AssertNotCalled(t, "CreateChannel", mock.Anything)
+	mockTxStore.AssertNotCalled(t, "StoreUserState", mock.Anything, mock.Anything)
```
nksazonov, line 879: Three gaps here: (1) `AssertNotCalled` verifies the call never happened but doesn't encode the *ordering* constraint — use `mockTxStore.AssertNumberOfCalls(t, "LockUserState", 0)` to make it structurally enforced. (2) `mockMemoryStore.AssertExpectations(t)` is never called, so the `.Once()` expectation on `IsAssetSupported` is silently ignored if that call never happens. (3) `MockActionGateway.AllowAction` discards all three arguments with `_`, so the test can't verify that `AllowAction` is called with `core.GatedActionTransfer` and the correct wallet — a bug where `GatedAction()` returns the wrong value would pass silently. Use a testify/mock-based implementation that records and asserts arguments.

nksazonov (commented):
Good job addressing MF2-L01 cleanly. The gate is placed correctly (before any state effects), wraps errors consistently with `SubmitState`, and the rejection test verifies side-effect isolation. A pre-existing race condition in the action gateway and a few test coverage gaps are worth addressing.

nitronode/api/channel_v1/request_creation.go
philanton: Yeah, real race, but it's pre-existing in `SubmitState` too — `AllowAction` reads/records before the per-user lock regardless of which entrypoint calls it. I'd keep this PR scoped to closing the bypass and not retrofit `action_gateway` ordering here.

nitronode/api/channel_v1/request_creation_test.go
philanton: Added `TestRequestCreation_TransferSend_Success` in [1120ae1](https://github.com/layer-3/nitrolite/pull/763/commits/1120ae1d5b341e48783220ceacd56c8c7ad8ddb9) — mirrors the rejection test with a real `TransferSend` transition, receiver-side wiring (`StoreUserState` for receiver, `RecordTransaction`) and node-signature verification.

nitronode/api/channel_v1/request_creation_test.go
philanton: Addressed (2) in [1120ae1](https://github.com/layer-3/nitrolite/pull/763/commits/1120ae1d5b341e48783220ceacd56c8c7ad8ddb9) — added `mockMemoryStore.AssertExpectations(t)` so the `.Once()` is enforced.

For (1), I'd keep `AssertNotCalled` — testify treats both as presence checks, so `AssertNumberOfCalls(..., 0)` doesn't actually encode call ordering, just rephrases the same assertion.

For (3), `MockActionGateway` is shared across ~10 call sites in `submit_state_test.go` with zero-config usage. Swapping to a testify/mock-based version with arg capture is a package-wide refactor, and the `GatedAction()` return is already unit-covered in `pkg/core`. I'd skip here.

nksazonov (approved):

ihsraham (approved):
Approved for L-01. The latest head gates `RequestCreation` before channel creation, signing, state storage, receiver-state issuance, and transaction recording, and the added rejection plus `TransferSend` success coverage closes the original channel-creation bypass.

One P2 follow-up I would track outside this PR: `AllowAction` still does count-then-record without a wallet/action lock, so the allowance can race under concurrent actions. That matches the existing `SubmitState` behavior and does not reopen this finding for me. Thanks for tightening this path up.

PR #764, opened 2026-05-18
MF2-C02: bound ledger values to Solidity uint256/int256 ranges

philanton:
## Summary

Reject ledger states whose scaled values fall outside the Solidity ABI ranges used for signing and onchain calls. A caller could otherwise submit a `home_deposit` whose raw amount is `2^256 + X`; the node would store the inflated offchain balance, but the signed EVM message and onchain calldata would only carry the low 256 bits, allowing the inflated balance to drain the node vault through a normal `transfer_send`.

- New `core.DecimalToUint256` / `core.DecimalToInt256` wrap the (now private) scaling helper and enforce `[0, 2^256-1]` and `[-2^255, 2^255-1]` respectively.
- `Ledger.Validate` now takes `assetDecimals` and bound-checks every scaled field.
- `StateAdvancerV1` looks up token decimals from the asset store and passes them in, so inflated state is rejected before the DB write.
- `StatePackerV1` promotes `contractLedger` to package scope, runs a `Validate()` guard on it before signing, and uses the bounded helpers when converting decimals.
- `coreLedgerToContractLedger` and the remaining `DepositToNode` / `Withdraw` / `Approve` / native deposit `Value` paths swap to `DecimalToUint256`, so onchain submission is gated by the same check.

Defense in depth: every layer that materialises a `*big.Int` from a ledger decimal now refuses out-of-range values, instead of relying on a single chokepoint.

## Test plan

- [x] `go vet ./...`
- [x] `go test ./...`
- [x] New unit tests in `pkg/core/utils_test.go` for `DecimalToUint256` / `DecimalToInt256` boundaries (max, max+1, negative, scaling-induced overflow).
- [x] `pkg/core/state_packer_test.go::TestPackState_RejectsOutOfRangeValues` — packer-level regression for both uint256 and int256 fields.
- [x] `pkg/core/state_advancer_test.go::TestValidateAdvancement_RejectsOverflowDeposit` — `home_deposit` with amount `2^256` is rejected before DB write.
- [x] `pkg/core/state_advancer_test.go::TestValidateAdvancement_RejectsOverflowNetFlow` — net flow `2^255` is rejected.
- [x] Existing `app_session_v1` deposit-state tests updated to stub the new `GetTokenDecimals` call now made by the advancer.

### DIFF
--- a/pkg/core/state_advancer_test.go
+++ b/pkg/core/state_advancer_test.go
@@ -244,10 +244,10 @@ func TestValidateAdvancement_RejectsInvalidAmount(t *testing.T) {
 	}
 }
 
-// TestValidateAdvancement_RejectsOverflowDeposit regression-tests the bug
-// described in YNU-XXX: a home_deposit whose scaled amount exceeds uint256
-// would previously be accepted offchain while the signed payload truncated to
-// the low 256 bits. ValidateAdvancement now rejects such states before storage.
+// TestValidateAdvancement_RejectsOverflowDeposit ensures a home_deposit whose
+// scaled amount exceeds uint256 is rejected before storage. Without the bound
+// check, the offchain ledger would record the full amount while the signed
+// payload truncates to the low 256 bits.
 func TestValidateAdvancement_RejectsOverflowDeposit(t *testing.T) {
 	t.Parallel()
 
@@ -315,4 +315,57 @@ func TestValidateAdvancement_RejectsOverflowNetFlow(t *testing.T) {
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "invalid home ledger")
 	assert.Contains(t, err.Error(), "int256")
+}
+
+// TestValidateAdvancement_RejectsOverflowEscrowLedger covers the escrow branch
+// of the bound check: a state whose EscrowLedger already carries an out-of-range
+// UserBalance is rejected even when the home ledger and the new transition are
+// otherwise valid.
+func TestValidateAdvancement_RejectsOverflowEscrowLedger(t *testing.T) {
+	t.Parallel()
+
+	store := newMockAssetStore()
+	store.AddToken(1, "0xHomeToken", 0)
+	store.AddToken(2, "0xEscrowToken", 0)
+	advancer := NewStateAdvancerV1(store)
+
+	userWallet := "0xUser"
+	asset := "USDC"
+	homeChanID := "0xHomeChannelId"
+	escrowChanID := "0xEscrowChannelId"
+	sig := "0xSig"
+
+	// 2^256, one above the uint256 range.
+	overflow := new(big.Int).Lsh(big.NewInt(1), 256)
+	overflowDec := decimal.NewFromBigInt(overflow, 0)
+
+	current := NewVoidState(asset, userWallet)
+	current.HomeChannelID = &homeChanID
+	current.EscrowChannelID = &escrowChanID
+	current.ID = GetStateID(userWallet, asset, 0, 0)
+	current.HomeLedger.TokenAddress = "0xHomeToken"
+	current.HomeLedger.BlockchainID = 1
+	// Last transition is acknowledgement so NextState carries the escrow ledger
+	// forward and the new transition is not constrained.
+	current.Transition = *NewTransition(TransitionTypeAcknowledgement, "0x0", "0x0", decimal.Zero)
+	// Escrow balances sum to net flows so the equality check passes; the
+	// uint256 bound is what should reject the state.
+	current.EscrowLedger = &Ledger{
+		BlockchainID: 2,
+		TokenAddress: "0xEscrowToken",
+		UserBalance:  overflowDec,
+		UserNetFlow:  overflowDec,
+		NodeBalance:  decimal.Zero,
+		NodeNetFlow:  decimal.Zero,
+	}
+
+	proposed := current.NextState()
+	_, err := proposed.ApplyHomeDepositTransition(decimal.NewFromInt(1))
+	require.NoError(t, err)
+	proposed.UserSig = &sig
+
+	err = advancer.ValidateAdvancement(*current, *proposed)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid escrow ledger")
+	assert.Contains(t, err.Error(), "uint256")
 }
--- a/pkg/core/state_advancer_test.go
+++ b/pkg/core/state_advancer_test.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"math/big"
 	"testing"
 
 	"github.com/shopspring/decimal"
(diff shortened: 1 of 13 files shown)


pkg/core/state_advancer_test.go
```diff
@@ -242,3 +243,76 @@ func TestValidateAdvancement_RejectsInvalidAmount(t *testing.T) {
 		}
 	}
 }
+
+// TestValidateAdvancement_RejectsOverflowDeposit regression-tests the bug
+// described in YNU-XXX: a home_deposit whose scaled amount exceeds uint256
+// would previously be accepted offchain while the signed payload truncated to
+// the low 256 bits. ValidateAdvancement now rejects such states before storage.
```
nksazonov: This comment should be reformulated to not mention the bug number.

nksazonov (commented):
Good job! A few missing coverage spots in the new test suite worth addressing before landing.

---

**Out-of-diff note — `pkg/blockchain/evm/locking_client.go:229`:** The two call sites in this file were migrated to `core.DecimalToUint256`, but the original `decimalToBigInt(amount decimal.Decimal, decimals int32) *big.Int` was left in place with no callers. It is now dead code. Remove it to avoid accidental reuse — its different signature (`int32` decimals, no error return) makes it visually distinct from the bounded helpers, but the potential for confusion remains.

**Out-of-diff note — `pkg/blockchain/evm/utils_test.go:139`:** `TestCoreLedgerToContractLedger_Success` covers the happy path after `coreLedgerToContractLedger` was updated to use `DecimalToUint256`/`DecimalToInt256`. There is no rejection test. A test asserting that a `Ledger` with `UserBalance = 2^256` causes `coreLedgerToContractLedger` to return an error containing `"uint256"` would give the blockchain-layer conversion the same coverage as the core-layer packer and advancer tests.

pkg/core/state_packer_test.go
```diff
@@ -103,3 +105,60 @@ func TestPackState(t *testing.T) {
...
+		state := newBaseState()
+		// balance fits uint256 but net flow exceeds int256
+		state.HomeLedger.UserBalance = decimal.NewFromBigInt(overflow255, 0)
+		state.HomeLedger.UserNetFlow = decimal.NewFromBigInt(overflow255, 0)
+
+		_, err := NewStatePackerV1(store).PackState(state)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "int256")
+	})
+}
```
nksazonov, line 207: `TestPackState_RejectsOutOfRangeValues` only exercises `HomeLedger`. `packSigningData` runs the same `DecimalToUint256`/`DecimalToInt256` conversions on `EscrowLedger` fields, but no subtest sets an escrow ledger field beyond its ABI type bound. Consider adding subtests with a populated `EscrowLedger` whose `UserBalance = 2^256` and `UserNetFlow = 2^255` to cover the escrow conversion path.

pkg/core/state_packer_test.go
```diff
@@ -103,3 +105,60 @@ func TestPackState(t *testing.T) {
...
+
+		state := newBaseState()
+		// balance fits uint256 but net flow exceeds int256
+		state.HomeLedger.UserBalance = decimal.NewFromBigInt(overflow255, 0)
+		state.HomeLedger.UserNetFlow = decimal.NewFromBigInt(overflow255, 0)
+
+		_, err := NewStatePackerV1(store).PackState(state)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "int256")
+	})
```
nksazonov, line 206: The lower bound (`< -2^255`) is verified in `TestDecimalToInt256/one_under_min_rejected` at the utility level, but there is no packer-level or advancer-level test exercising `NodeNetFlow = -(2^255 + 1)`. Consider adding a subtest with a negative net flow below `minInt256` to ensure the full path (decimal → `DecimalToInt256` → `contractLedger.Validate`) also rejects values below the lower bound.

pkg/core/state_advancer_test.go
```diff
@@ -242,3 +243,76 @@ func TestValidateAdvancement_RejectsInvalidAmount(t *testing.T) {
...
+	proposed := current.NextState()
+	_, err := proposed.ApplyHomeDepositTransition(amount)
+	require.NoError(t, err)
+	proposed.UserSig = &sig
+
+	err = advancer.ValidateAdvancement(*current, *proposed)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid home ledger")
+	assert.Contains(t, err.Error(), "int256")
+}
```
nksazonov, line 318: These two tests cover `HomeLedger` overflow via `ApplyHomeDepositTransition`. The escrow validation branch in `ValidateAdvancement` — `proposedState.EscrowLedger.Validate(escrowDecimals)` — has no overflow test. A test using a mutual-lock base state with `EscrowLedger.UserBalance = 2^256` would cover the `"invalid escrow ledger"` error path.

pkg/core/state_advancer_test.go
philanton: Done in [4426d54](https://github.com/layer-3/nitrolite/pull/764/commits/4426d548ca3c901bc203961da8547a1c72fb4914) — reworded the comment to describe the behavior without referencing a ticket.

pkg/core/state_packer_test.go
philanton: Yep, good catch — added `escrow_user_balance_overflows_uint256` and `escrow_user_net_flow_overflows_int256` subtests in [4426d54](https://github.com/layer-3/nitrolite/pull/764/commits/4426d548ca3c901bc203961da8547a1c72fb4914).

pkg/core/state_packer_test.go
philanton: I think the util-level `TestDecimalToInt256/one_under_min_rejected` is enough here — the packer just delegates to `DecimalToInt256`, so a packer-level subtest would mostly re-exercise the same code path. Happy to add it if you feel strongly though)

pkg/core/state_advancer_test.go
philanton: Added `TestValidateAdvancement_RejectsOverflowEscrowLedger` in [4426d54](https://github.com/layer-3/nitrolite/pull/764/commits/4426d548ca3c901bc203961da8547a1c72fb4914). Routes through an `Acknowledgement`-tailed current state + `HomeDeposit` proposed transition so the home ledger stays valid and the escrow `Validate` is what fires.

philanton:
@nksazonov on the out-of-diff notes:

1) Dropped the dead `decimalToBigInt` in `pkg/blockchain/evm/locking_client.go` (plus the now-unused `math/big` import) in [4426d54](https://github.com/layer-3/nitrolite/pull/764/commits/4426d548ca3c901bc203961da8547a1c72fb4914).
2) Added `TestCoreLedgerToContractLedger_RejectsOutOfRangeValues` covering both the uint256 and int256 paths in the same commit.

nksazonov (approved):

ihsraham (approved):
Approved for C-02. The latest head cleanly bounds ledger values before signing/storage and EVM conversion, and the added tests cover the important closure paths. Thanks for tightening this up.

PR #765, opened 2026-05-18
MF2-M02: accept pre-finalize escrow version in ongoing check

philanton:
## Summary
- `EnsureNoOngoingEscrowOperation` now status-gates the one-behind branch for `TransitionTypeEscrowDeposit`. The signed N+1 finalize state sits off-chain while the on-chain channel stays at INITIATE version N until the purge queue fires; that is the protocol-intended steady state, so we cannot require `Closed`. The branch now allows status ∈ {Open, Closed} and blocks otherwise.
- Fixes the stall reported in MF2-M02: `HandleEscrowDepositsPurged` marks the escrow channel `Closed` while preserving `StateVersion` at the initiate version, which previously caused `EnsureNoOngoingEscrowOperation` to report `escrow deposit finalization is still ongoing` and block transfer receives and app-session releases.
- Addresses #discussion_r3264742604: `Challenged` one-behind is now blocked. Co-signed N+1 is value-binding between parties, but on-chain resolution is still racing — finalize tx may not land, the escrow chain may settle at INITIATE, and replaying N+1 later could violate engine invariants. Stacking receiver-side state on N+1 in that window risks encumbering a credit that may rewind.
- Comparison uses `*v + 1 == StateVersion` to avoid uint underflow when version is 0.
- No change to `HandleEscrowDepositsPurged` or `StateVersion` handling. Other transitions (escrow_lock, mutual_lock, escrow_withdraw) untouched.

## Test plan
- [x] `go vet ./nitronode/store/database/...`
- [x] `go test ./nitronode/store/database/... -count=1`
- [x] `TestDBStore_EnsureNoOngoingEscrowOperation` cases:
  - chain caught up → allow
  - Open one-behind (pre-purge happy path) → allow
  - Closed one-behind (post-purge) → allow
  - Challenged one-behind → block
  - chain more than one version behind → block
  - escrow channel missing from DB → block

### DIFF
--- a/nitronode/store/database/db_store.go
+++ b/nitronode/store/database/db_store.go
@@ -245,9 +245,13 @@ func (s *DBStore) EnsureNoOngoingStateTransitions(wallet, asset string) error {
 // Validation logic by latest signed transition type:
 //   - escrow_lock / mutual_lock: always considered ongoing (no finalization yet)
 //   - escrow_deposit: considered settled when the on-chain escrow channel version
-//     equals the signed state version (finalize landed) or is exactly one behind
-//     (initiate state; finalize or on-chain purge has not landed yet, but the purge
-//     queue makes it terminal — do not block receiver-side state issuance)
+//     equals the signed state version (finalize landed). When the chain is exactly
+//     one behind, the signed N+1 finalize state lives off-chain and is gated on
+//     the escrow channel status: allowed for Open (protocol-intended steady state
+//     before the purge queue fires) and Closed (post-purge or post-finalize);
+//     blocked for Challenged (on-chain resolution still racing — finalize tx may
+//     not land, escrow chain may settle at INITIATE, and replaying N+1 later
+//     could violate engine invariants).
 //   - escrow_withdraw: considered ongoing while the on-chain escrow channel state
 //     version has not caught up with the signed state version
 //   - any other transition: not an escrow operation, allow
@@ -258,6 +262,7 @@ func (s *DBStore) EnsureNoOngoingEscrowOperation(wallet, asset string) error {
 		TransitionType       core.TransitionType
 		StateVersion         uint64
 		EscrowChannelVersion *uint64
+		EscrowChannelStatus  *core.ChannelStatus
 	}
 
 	var result escrowCheck
@@ -265,7 +270,8 @@ func (s *DBStore) EnsureNoOngoingEscrowOperation(wallet, asset string) error {
 		SELECT
 			s.transition_type as transition_type,
 			s.version as state_version,
-			ec.state_version as escrow_channel_version
+			ec.state_version as escrow_channel_version,
+			ec.status as escrow_channel_status
 		FROM channel_states s
 		LEFT JOIN channels ec ON ec.channel_id = s.escrow_channel_id
 		WHERE s.user_wallet = ?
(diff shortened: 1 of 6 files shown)


nksazonov (commented):
Good job! The `EnsureNoOngoingEscrowOperation` fix correctly closes the blocking path described in the audit finding, and the new test cases cover both the purge scenario and the in-flight finalize case.

**Out-of-diff note — secondary finding from the audit (not in this diff):**

The audit finding also noted: "`state_version` returned by `GetEscrowChannel()` and `GetChannels()` will not be up to date with the finalize state while the escrow channel is closed." `HandleEscrowDepositsPurged` (`event_handlers/service.go`) still only sets `channel.Status = ChannelStatusClosed` without advancing `StateVersion`, and the doc comment confirms this is intentional. However, since API consumers querying a purged escrow channel via `GetEscrowChannel` or `GetChannels` will see `state_version = N` (initiate version) rather than the finalize version `N+1`, it would be worth adding a note to the handler's doc comment or the `GetEscrowChannel` response docs explaining that `state_version` on a purged escrow reflects the initiate state, not the finalize state. That way callers aren't silently surprised by the discrepancy.

philanton:
Yeah, fair point. Added a doc note on the nitronode `GetEscrowChannel` handler, the Go and TS SDK client methods, and the `get_escrow_channel` entry in `docs/api.yaml` calling out that `state_version` stays at the initiate version (N) on a purge-closed escrow. In [40caa39](https://github.com/layer-3/nitrolite/pull/765/commits/40caa39dd329d11ee88f4e85474ee0aa6949da40).

nksazonov (approved):

nitronode/store/database/db_store.go
```diff
@@ -287,7 +291,13 @@ func (s *DBStore) EnsureNoOngoingEscrowOperation(wallet, asset string) error {
 		return fmt.Errorf("mutual lock is still ongoing")
 
 	case core.TransitionTypeEscrowDeposit:
-		if result.EscrowChannelVersion == nil || result.StateVersion != *result.EscrowChannelVersion {
+		// Accept escrow channel at signed state version (finalize landed) or one
+		// behind (initiate state; finalize/purge has not landed yet, but is terminal
+		// per on-chain purge queue, so do not block receiver-side state issuance).
+		// Compare via *v + 1 == StateVersion to avoid uint underflow when version is 0.
+		if result.EscrowChannelVersion == nil ||
```
ihsraham: [P1/blocking] This accepts the one-version-behind `escrow_deposit` case without checking the escrow channel is actually terminal. An active `Open` or `Challenged` escrow at version `N` with a signed finalize state at `N+1` has the same shape as the purged case, so receiver-side issuance can proceed while the escrow finalization is still ongoing and then hide that escrow state behind a newer receiver state.

Could we include `ec.status` in this query and only allow the `N+1` vs `N` case when the escrow channel is `Closed`? Exact version match can stay allowed, and tests should keep `Open`/`Challenged` one-behind cases blocking while `Closed` one-behind allows.

ihsraham (changes requested):
Looks close, but I’d hold approval on one boundary in the escrow gate. The one-version-behind case needs to be tied to a terminal escrow channel.

nitronode/store/database/db_store.go
philanton: Agree the shape-collision is real for one status, disagree on the fix.

`Closed`-only would block receiver-side issuance for ~3h after every escrow deposit, since the happy path is `Open` at N + signed N+1 sitting off-chain until the on-chain purge queue fires. Per the protocol, finalize is not enforced on-chain in the happy path — the co-signed N+1 is terminal off-chain, and `HandleEscrowDepositsPurged` deliberately preserves `state_version` at N.

The risky case is `Challenged`: on-chain dispute is live, node is racing to submit N+1 finalize on the escrow chain, and several failure modes (tx doesn't land in time, reorg, post-expiry replay against an INITIATE-settled escrow chain) can rewind real funds even though both parties co-signed N+1. Receiver-side state stacked on that baseline encumbers a credit that may not hold.

Pushed f7b4a598 — `ec.status` is now in the query and the one-behind branch allows `{Open, Closed}` and blocks `Challenged` (and any other status). Test coverage split accordingly.

ihsraham (approved):
Thanks, this addresses the escrow status boundary and the added coverage pins the intended Open/Closed vs Challenged behavior.

PR #766, opened 2026-05-18
MF2-I03: reject reused home channel ID explicitly

philanton:
## Summary
- After a challenged channel close, `issueChallengeRescue()` emits a fresh-epoch state with `HomeChannelID == nil`. The same-home-channel-id check in `request_creation()` was gated on `currentState.HomeChannelID != nil`, so when the ledger head was a post-rescue state the explicit reject was skipped and the invariant fell back to the `channels.channel_id` primary key on `CreateChannel`.
- No direct funds risk — duplicate creation still aborts on the PK collision inside the transaction — but the user-facing error is opaque (`failed to create channel: duplicate key`) and the business rule lives in the storage layer instead of the handler.
- Patch looks up the channel by computed `homeChannelID` before `CreateChannel` and rejects with `cannot use same home channel id` regardless of prior-state shape. The `channel is already initialized` guard for non-final prior states with a non-nil `HomeChannelID` is preserved.

## Test plan
- [x] `go test ./nitronode/api/channel_v1/...`
- [x] `go vet ./nitronode/api/channel_v1/...`
- [x] New regression `TestRequestCreation_RejectReusedHomeChannelID` covers post-rescue head (`HomeChannelID == nil`) with a prior channel record still in `channels` — must return `cannot use same home channel id` before reaching `CreateChannel`/`StoreUserState`.

### DIFF
--- a/nitronode/api/channel_v1/request_creation.go
+++ b/nitronode/api/channel_v1/request_creation.go
@@ -128,14 +128,19 @@ func (h *Handler) RequestCreation(c *rpc.Context) {
 			return rpc.Errorf("incoming state home_channel_id is invalid")
 		}
 
-		if currentState.HomeChannelID != nil {
-			isFinal := currentState.IsFinal()
-			if !isFinal {
-				return rpc.Errorf("channel is already initialized")
-			}
-			if isFinal && strings.EqualFold(*incomingState.HomeChannelID, *currentState.HomeChannelID) {
-				return rpc.Errorf("cannot use same home channel id")
-			}
+		// Reject reuse of an already-known channel ID. Enforced at the handler rather than
+		// relying on the channels.channel_id primary key, so the invariant holds even when
+		// the prior state has no HomeChannelID (e.g., after a ChallengeRescue squash).
+		existingChannel, err := tx.GetChannelByID(homeChannelID)
+		if err != nil {
+			return rpc.Errorf("failed to check existing channel: %v", err)
+		}
+		if existingChannel != nil {
+			return rpc.Errorf("cannot use same home channel id")
+		}
+
+		if currentState.HomeChannelID != nil && !currentState.IsFinal() {
+			return rpc.Errorf("channel is already initialized")
 		}
 
 		logger.Debug("processing channel creation request", "incomingVersion", incomingState.Version)
--- a/nitronode/api/channel_v1/request_creation.go
+++ b/nitronode/api/channel_v1/request_creation.go
@@ -141,7 +141,7 @@ func (h *Handler) RequestCreation(c *rpc.Context) {
 		// the prior state has no HomeChannelID (e.g., after a ChallengeRescue squash).
 		existingChannel, err := tx.GetChannelByID(homeChannelID)
 		if err != nil {
-			return rpc.Errorf("failed to check existing channel: %v", err)
+			return rpc.Errorf("failed to look up channel by computed id: %v", err)
 		}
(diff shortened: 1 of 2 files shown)


nksazonov (commented):
Good job! The audit finding is fully addressed with a clean, well-tested fix.

nitronode/api/channel_v1/request_creation.go
```diff
@@ -136,14 +136,19 @@ func (h *Handler) RequestCreation(c *rpc.Context) {
...
-			}
-			if isFinal && strings.EqualFold(*incomingState.HomeChannelID, *currentState.HomeChannelID) {
-				return rpc.Errorf("cannot use same home channel id")
-			}
+		// Reject reuse of an already-known channel ID. Enforced at the handler rather than
+		// relying on the channels.channel_id primary key, so the invariant holds even when
+		// the prior state has no HomeChannelID (e.g., after a ChallengeRescue squash).
+		existingChannel, err := tx.GetChannelByID(homeChannelID)
+		if err != nil {
+			return rpc.Errorf("failed to check existing channel: %v", err)
```
nksazonov: The string `"failed to check existing channel: %v"` is identical to the error at line 113, which wraps a `GetLastUserState` failure. When either fires in production logs, there's no way to tell which operation failed. Consider `"failed to look up channel by computed id: %v"` or similar to make the two distinct.

nitronode/api/channel_v1/request_creation.go
```diff
@@ -136,14 +136,19 @@ func (h *Handler) RequestCreation(c *rpc.Context) {
...
+		}
+		if existingChannel != nil {
+			return rpc.Errorf("cannot use same home channel id")
+		}
+
+		if currentState.HomeChannelID != nil && !currentState.IsFinal() {
+			return rpc.Errorf("channel is already initialized")
 		}
```
nksazonov, line 152: This path is preserved correctly and still reachable: fires when a user proposes a fresh channel ID (not in the DB) while `currentState.HomeChannelID != nil && !IsFinal()`. There's no dedicated test for it — neither before nor after this PR. Worth adding one to document the invariant: `currentState` with a non-nil, non-final `HomeChannelID`, `GetChannelByID` returning nil, expecting `"channel is already initialized"`. Not a blocker, pre-existing gap.

nitronode/api/channel_v1/request_creation.go
philanton: Yep, renamed to `failed to look up channel by computed id: %v` in [250b9dd4](https://github.com/layer-3/nitrolite/pull/766/changes/250b9dd4165be8bd1ec929588a5c09765843fb68).

nitronode/api/channel_v1/request_creation.go
philanton: Yeah, fair — added `TestRequestCreation_ChannelAlreadyInitialized` covering exactly that path (non-nil non-final prior `HomeChannelID`, `GetChannelByID` returning nil, expecting `channel is already initialized`) in [250b9dd4](https://github.com/layer-3/nitrolite/pull/766/changes/250b9dd4165be8bd1ec929588a5c09765843fb68).

nksazonov (approved):

ihsraham (approved):
Thanks for tightening this up. The latest head closes I-03 for me: reused computed home channel IDs are rejected before CreateChannel, including the detached-state case, and the added regression coverage covers the closure path. Approving.

PR #768, opened 2026-05-18
MF2-H03: guard challenge_rescue against post-Finalize close

philanton:
## Summary

- `HandleHomeChannelClosed` previously issued a challenge rescue for every channel that was `Challenged` before the on-chain close, including cooperative closes backed by a node-signed `Finalize`. In that path the rescue either silently overwrote a post-Finalize receiver credit at `(epoch+1, version=0)` or collided on deterministic `GetStateID` with an existing `(epoch+1, version=1)` credit — rolling back the close-event handler and leaving the channel stuck `Challenged`, which in turn blocked `request_creation` for the same `(wallet, asset)` pair.
- Guard the rescue branch with `HasSignedFinalize`. Mirrors the existing post-Finalize check in `HandleHomeChannelCheckpointed`.
- Align `NewChallengeRescueState` with the `NextState()` post-final convention: fresh-epoch state starts at version `0`, not `1`. The guard guarantees no other `(epoch+1, version=0)` row exists when the rescue actually runs.

## Failure modes prevented

Post-Finalize receiver credits are produced by `NextState()` in a fresh epoch with `HomeChannelID=nil`. Without the guard:

1. `GetLastStateByChannelID` only matches channel-attached rows → returns the Finalize state at the original epoch, missing the post-Finalize credits.
2. `SumNetTransitionAmountAfterVersion` filters by `home_channel_id` → returns zero.
3. `NewChallengeRescueState` builds a zero-amount rescue at `(Finalize.Epoch+1, version=1)`.
4. `StoreUserState` either overwrites the lone credit at `(epoch+1, version=0)` (silent balance loss) or collides with `(epoch+1, version=1)` and rolls back the entire close handler.

## Test plan

- [x] `go test ./pkg/core/...` — `TestNewChallengeRescueState` updated for version `0`.
- [x] `go test ./nitronode/event_handlers/... -run TestHandleHomeChannelClosed` — all rescue tests pass.
- [x] New `TestHandleHomeChannelClosed_PostFinalize_SkipsRescue` verifies the rescue branch fully short-circuits when `HasSignedFinalize` is true.
- [x] New `TestHandleHomeChannelClosed_PostFinalize_CollisionRegression` documents the two failure modes the guard prevents.
- [x] New `TestHandleHomeChannelClosed_PostFinalize_HasSignedFinalizeErr` confirms error propagation from the `HasSignedFinalize` lookup.
- [x] Existing rescue tests (`_Squash`, `_NoCredits`, `_NegativeNet_ClampsToZero`) updated for the new version=0 invariant and the added `HasSignedFinalize` mock.
- [x] `go vet ./nitronode/event_handlers/... ./pkg/core/...` clean.

## Follow-ups (separate tickets)

- Post-Finalize unsigned receiver credits at `(epoch+1, version=0..N)` with `HomeChannelID=nil` remain unsigned forever (`channel_v1/handler.go:130` skips node-sig when channel status is non-Open). Confirm `request_creation` folds them into the next signed ledger correctly; otherwise file a separate fix.
- Operator reconciliation for production channels already wedged at `Challenged` due to this bug pre-fix: needs a detection query (`Challenged` channels where `HasSignedFinalize=true`) and a one-shot flip to `Closed`.

### DIFF
--- a/nitronode/event_handlers/service.go
+++ b/nitronode/event_handlers/service.go
@@ -313,9 +313,15 @@ func (s *EventHandlerService) HandleHomeChannelChallenged(ctx context.Context, t
 // Once closed, no further state updates are possible for this channel.
 //
 // Additionally, when the closing path was a challenge resolution (channel was Challenged
-// and the closing state is not a finalize), the handler issues a single ChallengeRescue
-// state for the user that squashes the sum of receiver-state credits accrued during the
-// challenge window into an off-channel ledger entry tied to the closed channel's ID.
+// and no node-signed Finalize state exists for it), the handler issues a single
+// ChallengeRescue state for the user that squashes the sum of receiver-state credits
+// accrued during the challenge window into an off-channel ledger entry tied to the
+// closed channel's ID. When a node-signed Finalize already exists, the cooperative-close
+// path has already advanced the user to a fresh epoch via NextState(), and any receiver
+// credits accumulated post-Finalize live at (epoch+1, version=0..N) with HomeChannelID=nil
+// — issuing a rescue at (epoch+1, version=1) would either overwrite the lone receiver
+// credit or collide on deterministic state ID with an existing row, so the rescue is
+// skipped entirely in that case.
 func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx core.ChannelHubEventHandlerStore, event *core.HomeChannelClosedEvent) error {
 	logger := log.FromContext(ctx)
 	chanID := event.ChannelID
@@ -362,8 +368,23 @@ func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx co
 	}
 
 	if wasChallenged {
-		if err := s.issueChallengeRescue(ctx, tx, channel, event.StateVersion); err != nil {
+		// Post-Finalize cooperative close racing with an on-chain challenge: receiver
+		// credits issued after the Finalize live in a fresh epoch via NextState() and
+		// are not channel-attached, so the rescue path would either overwrite them or
+		// collide on deterministic state ID. Skip the rescue branch entirely.
+		finalized, err := tx.HasSignedFinalize(chanID)
+		if err != nil {
 			return err
+		}
+		if finalized {
+			logger.Info("skipping challenge_rescue for post-Finalize close",
+				"channelId", chanID,
+				"userWallet", channel.UserWallet,
+				"closureVersion", event.StateVersion)
+		} else {
+			if err := s.issueChallengeRescue(ctx, tx, channel, event.StateVersion); err != nil {
+				return err
+			}
 		}
 	}
 
--- a/nitronode/event_handlers/service.go
+++ b/nitronode/event_handlers/service.go
@@ -377,7 +377,7 @@ func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx co
 			return err
 		}
 		if finalized {
-			logger.Info("skipping challenge_rescue for post-Finalize close",
+			logger.Debug("skipping challenge_rescue for post-Finalize close",
 				"channelId", chanID,
 				"userWallet", channel.UserWallet,
 				"closureVersion", event.StateVersion)
--- a/nitronode/event_handlers/service_test.go
+++ b/nitronode/event_handlers/service_test.go
@@ -1888,6 +1888,10 @@ func TestHandleHomeChannelClosed_PostFinalize_SkipsRescue(t *testing.T) {
 //
 // With the guard, HasSignedFinalize=true short-circuits the entire branch before any
 // of those DB calls fire. The mock asserts none of them are reached.
+//
+// Intentional twin of TestHandleHomeChannelClosed_PostFinalize_SkipsRescue: setup and
+// AssertNotCalled set are identical at the mock level; this case exists to document the
+// specific collision failure modes the guard prevents. Do not collapse the two.
 func TestHandleHomeChannelClosed_PostFinalize_CollisionRegression(t *testing.T) {
 	mockStore := new(MockStore)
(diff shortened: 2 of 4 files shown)


nitronode/event_handlers/service.go
```diff
@@ -362,9 +368,24 @@ func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx co
...
+		// collide on deterministic state ID. Skip the rescue branch entirely.
+		finalized, err := tx.HasSignedFinalize(chanID)
+		if err != nil {
 			return err
 		}
+		if finalized {
+			logger.Info("skipping challenge_rescue for post-Finalize close",
+				"channelId", chanID,
+				"userWallet", channel.UserWallet,
+				"closureVersion", event.StateVersion)
```
nksazonov: What is the point of this log?

pkg/core/types.go
```diff
@@ -177,11 +177,16 @@ func NewVoidState(asset, userWallet string) *State {
...
+// after a challenge. The rescue state opens a fresh epoch at version 0 with no home
 // channel and no token/blockchain context (the closed channel is gone); it credits the
 // user the rescued amount as a standalone ledger entry referencing the closed channel
 // (taken from prev.HomeChannelID) via the transition's AccountID. The rescue state is
 // meant to be stored unsigned, like a credit to a user with no open home channel.
+//
+// Version 0 matches the NextState() post-final convention: any in-protocol fresh-epoch
+// state starts at version 0. Callers must guarantee no other (epoch+1, version=0) row
+// already exists for this (wallet, asset) — for the rescue path this is enforced by
+// only invoking it on non-Finalize challenge closes (see HandleHomeChannelClosed).
```
nksazonov, line 189: The invariant is sound for the current callsite (`issueChallengeRescue`, guarded by `HasSignedFinalize=false`), but it is only documented, not enforced inside the function itself. If `NewChallengeRescueState` is ever called from a second callsite that lacks the guard, `StoreUserState` will silently fail with an ID collision and leave the caller in a hard-to-diagnose state.

Suggestion: either make the function self-guarding (return an error if it detects a pre-existing row), or at minimum document the single authorized callsite in this comment so future authors know it is not a general-purpose helper.

nitronode/event_handlers/service_test.go
```diff
@@ -1808,9 +1811,173 @@ func TestHandleHomeChannelClosed_ChallengeRescue_NegativeNet_ClampsToZero(t *tes
...
+	err := service.HandleHomeChannelClosed(ctx, mockStore, event)
+	require.NoError(t, err, "post-Finalize close must succeed; rescue branch must be skipped")
+
+	mockStore.AssertExpectations(t)
+	// The collision-class operations must not be invoked once the guard fires.
+	mockStore.AssertNotCalled(t, "GetLastStateByChannelID", mock.Anything, mock.Anything)
+	mockStore.AssertNotCalled(t, "SumNetTransitionAmountAfterVersion", mock.Anything, mock.Anything, mock.Anything)
+	mockStore.AssertNotCalled(t, "StoreUserState", mock.Anything, mock.Anything)
+	mockStore.AssertNotCalled(t, "RecordTransaction", mock.Anything, mock.Anything)
+}
```
nksazonov, line 1942: `TestHandleHomeChannelClosed_PostFinalize_CollisionRegression` and `TestHandleHomeChannelClosed_PostFinalize_SkipsRescue` (lines 1827–1873) are behaviorally indistinguishable at the mock level — identical setup, identical `AssertNotCalled` constraints. The second test adds value only through its docstring.

Suggestion: collapse into one test with sub-cases (e.g. `t.Run`), or add an explicit comment to the second function noting it is an intentional documentation duplicate, so it does not silently diverge from the first over time.

nksazonov (commented):
Excellent work. The guard at `service.go:375` correctly and completely remediates all three failure modes described in the audit finding — the balance-overwrite path, the `StoreUserState` ID-collision path, and the resulting `Challenged`-channel deadlock. The version-0 correction in `NewChallengeRescueState` aligns the rescue with the `NextState()` convention. All three new regression tests are well-scoped and the existing tests are correctly updated. The two comments below are informational — neither blocks merging.

nitronode/event_handlers/service.go
philanton: Yeah, fair — it was meant as an operator breadcrumb for the rare race, but `Info` is too loud for a no-op branch. Demoted to `Debug` in [e28dddc](https://github.com/layer-3/nitrolite/pull/768/changes/e28dddc734093d37138ebcfc40f378bea515b835).

pkg/core/types.go
philanton: Good point. I went with the documentation route in [e28dddc](https://github.com/layer-3/nitrolite/pull/768/changes/e28dddc734093d37138ebcfc40f378bea515b835) — extended the docstring to call out `issueChallengeRescue` as the sole authorized callsite and spell out the `GetStateID` collision a second unguarded caller would hit. Self-guarding would mean an extra DB lookup on the rescue path just to defend against a hypothetical future misuse, which felt like the wrong trade-off here.

nitronode/event_handlers/service_test.go
philanton: Yeah, the two cases really are mock-indistinguishable — `_CollisionRegression` exists for the docstring, not the assertions. I would rather keep them as two named tests than fold into a `t.Run` and lose the named regression in stack traces, so I added a "Do not collapse" note on the twin in [e28dddc](https://github.com/layer-3/nitrolite/pull/768/changes/e28dddc734093d37138ebcfc40f378bea515b835) for whoever passes by next.

nksazonov (approved):

ihsraham (approved):
Approving for H-03. Nice fix: the signed-Finalize guard closes the destructive rescue path, while the normal challenge-rescue flow stays covered. Thanks for the focused tests and follow-up cleanup.

PR #775, opened 2026-05-19
YNU-918: treat past expires_at as session-key revocation

philanton:
## Summary
- Submit handlers for `channels.v1.submit_session_key_state` and `app_sessions.v1.submit_session_key_state` accept a past `expires_at` as a revocation, preserving the monotonic version sequence. The auth path already filters `expires_at > now`, so the key deactivates immediately; a later submit with the next version and a future `expires_at` re-activates the same session key address.
- Reject `expires_at < 0` instead of `expires_at <= now`. Defense-in-depth against `int64 → uint64` wrap in the metadata-hash packer, which would otherwise silently desynchronize the user-signed payload from the persisted row.
- `CountSessionKeysForUser` JOINs both history tables and only counts rows with `expires_at > now`, so a revoke frees the per-user cap slot. A single `now` is bound for both kind branches. The `int64` sum is bounds-checked before the `uint32` cast.
- Emit an Info log (`session key revoked` / `channel session key revoked`) when a submission deactivates the key, distinct from the existing `successfully stored` log.
- Refresh `docs/api.yaml`, `pkg/rpc/types.go`, and the Go SDK `Submit*` doc comments to describe revoke and re-activation semantics, including the explicit constraint that `session_key_sig` is required on every submit (including revocation).

## Follow-ups (out of scope for this PR)
- **Wallet-only revocation path** for lost or compromised keys. The current handler requires `session_key_sig` on every submit, so a user cannot revoke a key whose private material they no longer control. A wallet-only flow is a separate code path (asymmetric auth, replay-across-wallets threat model, version monotonicity when the key is unrecoverable) and warrants its own ticket.
- **SDK ergonomics — `Revoke*SessionKeyState` helpers** for the Go SDK and the TypeScript SDK, hiding the past-timestamp construction, version bump, and dual signing behind a first-class API. Also touches the TS SDK public-API drift snapshots, so worth a dedicated PR.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./nitronode/api/app_session_v1/... ./nitronode/api/channel_v1/... ./nitronode/store/database/...`
- [ ] CI green

## Release Notes

* **New Features**
  * Session key reactivation is now supported after revocation
  * Session key revocation via past `expires_at` timestamp

* **Documentation**
  * API documentation clarified with explicit session key lifecycle semantics (registration, update, revocation, reactivation)
  * Updated examples and guides reflecting revised revocation procedure
  * Clarified validation rules and signature requirements for session key submissions

* **Behavior Changes**
  * Per-user session key cap enforcement now applies only when activating previously inactive slots
  * Negative unix timestamps are explicitly rejected
  * Both user and session key signatures required for all submission types

### DIFF
--- a/nitronode/api/app_session_v1/submit_session_key_state_test.go
+++ b/nitronode/api/app_session_v1/submit_session_key_state_test.go
@@ -330,6 +330,88 @@ func TestSubmitSessionKeyState_RevokeWithPastExpiresAt(t *testing.T) {
 	mockStore.AssertExpectations(t)
 }
 
+// Covers the typical revocation path: an active key (latestVersion > 0) is deactivated
+// by submitting version+1 with a past expires_at. The per-user cap check is short-circuited
+// because the key already exists, so CountSessionKeysForUser must not be called.
+func TestSubmitSessionKeyState_RevokeExistingActiveKey(t *testing.T) {
+	mockStore := new(MockStore)
+	userSigner := NewMockSigner()
+	userAddress := strings.ToLower(userSigner.PublicKey().Address().String())
+	sessionKeySigner := NewMockSigner()
+	sessionKeyAddress := strings.ToLower(sessionKeySigner.PublicKey().Address().String())
+
+	handler := &Handler{
+		useStoreInTx: func(handler StoreTxHandler) error {
+			return handler(mockStore)
+		},
+		metrics:               metrics.NewNoopRuntimeMetricExporter(),
+		maxSessionKeyIDs:      10,
+		maxSessionKeysPerUser: 5,
+	}
+
+	expiresAt := time.Now().Add(-time.Hour).Truncate(time.Second)
+	reqPayload := buildSignedSessionKeyStateReq(t, userAddress, sessionKeyAddress, 2, nil, nil, expiresAt, userSigner, sessionKeySigner)
+
+	mockStore.On("LockSessionKeyState", userAddress, sessionKeyAddress, database.SessionKeyKindAppSession).Return(1, nil)
+	mockStore.On("StoreAppSessionKeyState", mock.AnythingOfType("app.AppSessionKeyStateV1")).Return(nil)
+
+	payload, err := rpc.NewPayload(reqPayload)
+	require.NoError(t, err)
+
+	ctx := &rpc.Context{
+		Context: context.Background(),
+		Request: rpc.NewRequest(1, rpc.AppSessionsV1SubmitSessionKeyStateMethod.String(), payload),
+	}
+
+	handler.SubmitSessionKeyState(ctx)
+
+	require.NotNil(t, ctx.Response)
+	assert.Nil(t, ctx.Response.Error())
+	mockStore.AssertExpectations(t)
+	mockStore.AssertNotCalled(t, "CountSessionKeysForUser", mock.Anything)
+}
+
+// Covers the re-activation path: after a revoke (latestVersion incremented twice — register
+// then revoke), submitting version+1 with a future expires_at re-activates the same session
+// key address without re-running the per-user cap check.
+func TestSubmitSessionKeyState_ReactivateAfterRevoke(t *testing.T) {
+	mockStore := new(MockStore)
+	userSigner := NewMockSigner()
+	userAddress := strings.ToLower(userSigner.PublicKey().Address().String())
+	sessionKeySigner := NewMockSigner()
+	sessionKeyAddress := strings.ToLower(sessionKeySigner.PublicKey().Address().String())
+
+	handler := &Handler{
+		useStoreInTx: func(handler StoreTxHandler) error {
+			return handler(mockStore)
+		},
+		metrics:               metrics.NewNoopRuntimeMetricExporter(),
+		maxSessionKeyIDs:      10,
+		maxSessionKeysPerUser: 5,
+	}
+
+	expiresAt := time.Now().Add(24 * time.Hour).Truncate(time.Second)
+	reqPayload := buildSignedSessionKeyStateReq(t, userAddress, sessionKeyAddress, 3, nil, nil, expiresAt, userSigner, sessionKeySigner)
+
+	mockStore.On("LockSessionKeyState", userAddress, sessionKeyAddress, database.SessionKeyKindAppSession).Return(2, nil)
+	mockStore.On("StoreAppSessionKeyState", mock.AnythingOfType("app.AppSessionKeyStateV1")).Return(nil)
+
+	payload, err := rpc.NewPayload(reqPayload)
+	require.NoError(t, err)
+
+	ctx := &rpc.Context{
+		Context: context.Background(),
+		Request: rpc.NewRequest(1, rpc.AppSessionsV1SubmitSessionKeyStateMethod.String(), payload),
+	}
+
+	handler.SubmitSessionKeyState(ctx)
+
+	require.NotNil(t, ctx.Response)
+	assert.Nil(t, ctx.Response.Error())
+	mockStore.AssertExpectations(t)
+	mockStore.AssertNotCalled(t, "CountSessionKeysForUser", mock.Anything)
+}
+
 func TestSubmitSessionKeyState_RejectsNegativeExpiresAt(t *testing.T) {
 	mockStore := new(MockStore)
 	handler := &Handler{
--- a/nitronode/api/app_session_v1/submit_session_key_state_test.go
+++ b/nitronode/api/app_session_v1/submit_session_key_state_test.go
@@ -292,7 +292,45 @@ func TestSubmitSessionKeyState_InvalidUserAddress(t *testing.T) {
 	assert.Contains(t, respErr.Error(), "invalid user_address")
 }
 
-func TestSubmitSessionKeyState_ExpiredExpiresAt(t *testing.T) {
+func TestSubmitSessionKeyState_RevokeWithPastExpiresAt(t *testing.T) {
+	mockStore := new(MockStore)
+	userSigner := NewMockSigner()
+	userAddress := strings.ToLower(userSigner.PublicKey().Address().String())
+	sessionKeySigner := NewMockSigner()
+	sessionKeyAddress := strings.ToLower(sessionKeySigner.PublicKey().Address().String())
+
+	handler := &Handler{
+		useStoreInTx: func(handler StoreTxHandler) error {
+			return handler(mockStore)
+		},
+		metrics:          metrics.NewNoopRuntimeMetricExporter(),
+		maxSessionKeyIDs: 10,
+	}
+
+	// expires_at in the past expresses a revoke: the same monotonic version sequence
+	// is preserved, the auth path filters expires_at > now so the key is deactivated.
+	expiresAt := time.Now().Add(-time.Hour).Truncate(time.Second)
+	reqPayload := buildSignedSessionKeyStateReq(t, userAddress, sessionKeyAddress, 1, nil, nil, expiresAt, userSigner, sessionKeySigner)
+
+	mockStore.On("LockSessionKeyState", userAddress, sessionKeyAddress, database.SessionKeyKindAppSession).Return(0, nil)
+	mockStore.On("StoreAppSessionKeyState", mock.AnythingOfType("app.AppSessionKeyStateV1")).Return(nil)
+
+	payload, err := rpc.NewPayload(reqPayload)
+	require.NoError(t, err)
+
+	ctx := &rpc.Context{
+		Context: context.Background(),
+		Request: rpc.NewRequest(1, rpc.AppSessionsV1SubmitSessionKeyStateMethod.String(), payload),
+	}
+
+	handler.SubmitSessionKeyState(ctx)
+
+	require.NotNil(t, ctx.Response)
+	assert.Nil(t, ctx.Response.Error())
+	mockStore.AssertExpectations(t)
+}
+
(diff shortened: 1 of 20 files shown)


nitronode/store/database/current_session_key_state.go
```diff
@@ -136,19 +136,34 @@ func (s *DBStore) LockSessionKeyState(userAddress, sessionKey string, kind Sessi
...
+	var appCount int64
+	err = s.db.Table("current_session_key_states_v1 AS c").
+		Joins("JOIN app_session_key_states_v1 h ON h.user_address = c.user_address AND h.session_key = c.session_key AND h.version = c.version").
+		Where("c.user_address = ? AND c.kind = ? AND c.version > 0 AND h.expires_at > ?", userAddress, SessionKeyKindAppSession, now).
+		Count(&appCount).Error
+	if err != nil {
+		return 0, fmt.Errorf("failed to count app session keys for user: %w", err)
+	}
+
+	return uint32(channelCount + appCount), nil
```
nksazonov: The `int64 + int64` sum is cast to `uint32` without a bounds check. If the total ever exceeds `math.MaxUint32` (unlikely under normal caps, but possible if the concurrent-new-key race described in the `TODO(MF-H01-followup)` above is hit repeatedly), the result wraps and the cap comparison `count >= uint32(h.maxSessionKeysPerUser)` sees a tiny number and passes incorrectly.

Add `if channelCount+appCount > math.MaxUint32 { return 0, fmt.Errorf("session key count overflows cap type") }` before the cast, or change the return type and comparison to `int64` throughout.

nitronode/api/app_session_v1/submit_session_key_state_test.go
```diff
@@ -292,7 +292,45 @@ func TestSubmitSessionKeyState_InvalidUserAddress(t *testing.T) {
...
+		metrics:          metrics.NewNoopRuntimeMetricExporter(),
+		maxSessionKeyIDs: 10,
+	}
+
+	// expires_at in the past expresses a revoke: the same monotonic version sequence
+	// is preserved, the auth path filters expires_at > now so the key is deactivated.
+	expiresAt := time.Now().Add(-time.Hour).Truncate(time.Second)
+	reqPayload := buildSignedSessionKeyStateReq(t, userAddress, sessionKeyAddress, 1, nil, nil, expiresAt, userSigner, sessionKeySigner)
+
+	mockStore.On("LockSessionKeyState", userAddress, sessionKeyAddress, database.SessionKeyKindAppSession).Return(0, nil)
```
nksazonov: `Return(0, nil)` makes this test exercise "register a brand-new key that already has a past `expires_at`," not "revoke an existing active key." In production the typical revocation path starts with `latestVersion > 0`.

Consider a second subtest that stubs `LockSessionKeyState` returning `(1, nil)` and sends `version: 2` with a past `expires_at`, confirming the update path succeeds and that `CountSessionKeysForUser` is not called (the `latestVersion > 0` branch short-circuits the cap check).

nitronode/api/channel_v1/submit_session_key_state_test.go
```diff
@@ -212,7 +212,47 @@ func TestChannelSubmitSessionKeyState_InvalidUserAddress(t *testing.T) {
...
+	}
+
+	// expires_at in the past expresses a revoke: the same monotonic version sequence
+	// is preserved, the auth path filters expires_at > now so the key is deactivated.
+	expiresAt := time.Now().Add(-time.Hour).Truncate(time.Second)
+	assets := []string{"USDC"}
+
+	reqPayload := buildSignedChannelSessionKeyStateReq(t, userAddress, sessionKeyAddress, 1, assets, expiresAt, userSigner, sessionKeySigner)
+
+	mockStore.On("LockSessionKeyState", userAddress, sessionKeyAddress, database.SessionKeyKindChannel).Return(0, nil)
```
nksazonov: Same concern as the app-session test: `Return(0, nil)` simulates a new-key registration with past expiry, not a revocation of an existing key. Add a complementary test with `latestVersion = 1` to cover the real revocation path in the channel handler.

sdk/go/app_session.go
```diff
@@ -290,10 +290,17 @@ func (c *Client) RebalanceAppSessions(ctx context.Context, signedUpdates []app.S
...
+// revocation, or re-activation. The state must carry both the wallet's UserSig
+// (authorizing the delegation) and the session-key holder's SessionKeySig (proving
+// possession of the key); submits without a valid SessionKeySig are rejected.
+//
+// Set state.ExpiresAt to a future time to register or update the key. Set it to a
+// value at or before time.Now() to revoke the key — the auth path filters by
+// expires_at, so the key is deactivated immediately while keeping the same monotonic
+// version sequence. A later submit with the next version and a future ExpiresAt
+// re-activates the same session key address. Negative unix timestamps are rejected
+// by the server.
```
nksazonov, line 306: The doc comment documents revocation by passing a past `ExpiresAt` to `SubmitAppSessionKeyState`, but this is a footgun for callers who must know to set a past timestamp, compute the correct next version, and re-sign manually.

Consider adding dedicated `RevokeAppSessionKeyState` and `RevokeChannelSessionKeyState` helpers (and equivalents in the TypeScript SDK) that handle the past-timestamp construction, version bump, and signing internally. The low-level submit can remain for flexibility, but revocation deserves a first-class API surface — `registerSessionKey(pastTime)` is not an intuitive revocation UX.

nksazonov (commented):
Good job — the revocation approach is clean and the DB-level cap accounting is well-tested.

**Revocation requires the compromised key's private key.** Both handlers require `session_key_sig` unconditionally, even on the revocation path (`app_session_v1/submit_session_key_state.go:95`, `channel_v1/submit_session_key_state.go:78`). This means a user cannot revoke a lost or compromised session key using their wallet signature alone — the compromised key must co-sign the revocation. Worth documenting explicitly; if wallet-only revocation is a safety requirement, a separate code path (accepting `user_sig` alone when `expires_at <= now`) would be needed.

**No handler-level test for re-activation.** The PR description documents "a later submit with `version+1` and a future `expires_at` re-activates the same session key address," but no handler test exercises this round-trip (active → revoke → re-activate). The DB test covers slot-freeing only. Silent regressions in the re-activation path wouldn't be caught.

nitronode/store/database/current_session_key_state.go
philanton: Yep, good catch. Fixed in [f826dd8](https://github.com/layer-3/nitrolite/pull/775/changes/f826dd871501e081824b0f0a74ba924af29de43a) — `CountSessionKeysForUser` now bounds-checks the `int64` sum and returns an explicit error if it would overflow `uint32`. Kept the return type as-is since the cap value is `int` upstream and `math.MaxUint32` keys-per-user is a deeply theoretical regime, but the silent wrap was the real concern.

nitronode/api/app_session_v1/submit_session_key_state_test.go
philanton: Yep, you're right — `Return(0, nil)` was the new-key path, not revocation. Fixed in [f826dd8](https://github.com/layer-3/nitrolite/pull/775/changes/f826dd871501e081824b0f0a74ba924af29de43a). Added `TestSubmitSessionKeyState_RevokeExistingActiveKey` with `LockSessionKeyState` returning `(1, nil)` and `version: 2` + past `expires_at`, and asserts `CountSessionKeysForUser` is not called (so the `latestVersion > 0` short-circuit is locked in). Also added `_ReactivateAfterRevoke` to cover the future-`expires_at` follow-up submit on the same key.

sdk/go/app_session.go
philanton: Yeah, agree on the UX — `registerSessionKey(pastTime)` isn't a great revocation surface. I'd rather scope that as a follow-up though: the Go + TS helpers (and the TS drift snapshots that go with them) are enough that they deserve their own PR for review. Will open a follow-up ticket. For now I've at least documented the requirement explicitly in the SDK doc comments and `docs/api.yaml` in [f826dd8](https://github.com/layer-3/nitrolite/pull/775/changes/f826dd871501e081824b0f0a74ba924af29de43a) so callers aren't left guessing.

nitronode/api/channel_v1/submit_session_key_state_test.go
philanton: Same fix as the app-session test, mirrored on the channel handler in [f826dd8](https://github.com/layer-3/nitrolite/pull/775/changes/f826dd871501e081824b0f0a74ba924af29de43a): added `TestChannelSubmitSessionKeyState_RevokeExistingActiveKey` (latestVersion=1, version=2, past `expires_at`) and `_ReactivateAfterRevoke` (latestVersion=2, version=3, future `expires_at`), both asserting `CountSessionKeysForUser` is not called.

philanton:
Thanks for the review @nksazonov, addressed:

1) **Re-activation handler test** — fixed in [f826dd8](https://github.com/layer-3/nitrolite/pull/775/changes/f826dd871501e081824b0f0a74ba924af29de43a). Added `TestSubmitSessionKeyState_ReactivateAfterRevoke` for both the app-session and channel handlers (latestVersion=2, version=3, future `expires_at`), and a separate `_RevokeExistingActiveKey` to cover the typical revoke path the inline comments flagged.

2) **Wallet-only revocation** — yeah, intentionally out of scope here. The current PR is just the past-`expires_at` semantics on the existing submit handler. A wallet-only path is a separate flow (asymmetric auth, threat model around replay across wallets, version monotonicity when the key is lost) and I'd rather plan that properly in its own ticket. I've documented the constraint explicitly in `docs/api.yaml` and the Go SDK doc comments in the same commit so callers know the current limitation. Will open a follow-up ticket.

Also tracking the I4 SDK `Revoke*` helpers (Go + TS) as a separate follow-up — those touch the TS drift snapshots and `ts-compat` surface, so worth a dedicated PR.

nksazonov (approved):

ihsraham:
I would keep this open for one closure issue.

`CountSessionKeysForUser` now excludes inactive keys, but the cap check only runs for brand-new pointer rows (`latestVersion == 0`) in `nitronode/api/app_session_v1/submit_session_key_state.go:133` and the mirrored channel path in `nitronode/api/channel_v1/submit_session_key_state.go:116`. That lets a user revoke key A, register key B into the freed slot, then reactivate key A with a future `expires_at` without counting active keys again, ending up above `maxSessionKeysPerUser`.

One way to close this is to run the cap check whenever the submitted state is active and the previous latest state was inactive or absent, not only when `latestVersion == 0`.

Minor/non-blocking cleanup: the TS example app still revokes by clearing `assets` while keeping the old future `expires_at` (`sdk/ts/examples/example-app/README.md:296`, `sdk/ts/examples/example-app/src/components/WalletDashboard.tsx:503`), so under the new semantics it stores an active latest state. Some public comments/docs are also stale (`sdk/ts/src/client.ts`, `pkg/rpc/api.go`, and `nitronode/api/app_session_v1/README.md`) and still describe submit as registration/update only or say `expires_at` must be future. These are docs/example cleanup items, but worth fixing so integrators do not follow the old behavior.

philanton:
Yep, you're right about the bypass — fixed in [f6e7f34](https://github.com/layer-3/nitrolite/pull/775/changes/f6e7f34c). Extended `LockSessionKeyState` to also return the latest history `expires_at` so the handler can tell rotation from reactivation, and the cap check now fires whenever the submit transitions the slot from inactive to active:

```
prevActive := latestVersion > 0 && latestExpiresAt.After(now)
submittedActive := coreState.ExpiresAt.After(now)
if !prevActive && submittedActive && h.maxSessionKeysPerUser > 0 {
    // count, enforce
}
```

So new-key registration, and reactivation from a revoked state, both go through the cap; rotation against a still-active key and revoke submits skip it as before. New tests on both the app-session and channel handler — `_ReactivateAfterRevoke_AtCapRejected` and `_BelowCapAllowed` — pin the four branches.

On the cleanup: also fixed in the same commit.
1) `example-app/README.md` and `WalletDashboard.tsx` revoke flow now submits a past `expires_at` (kept `assets`/scoping unchanged), since under the new semantics clearing `assets` alone leaves the latest state active.
2) Doc comments on `submitChannelSessionKeyState` / `submitSessionKeyState` (`sdk/ts/src/client.ts`), the RPC types in `pkg/rpc/api.go`, the validation block in `nitronode/api/app_session_v1/README.md`, and the `submit_session_key_state` entry in `docs/api.yaml` all spell out the three flows (register / rotate / revoke) and that reactivation counts against the cap.

ihsraham (approved):
Approving. The closure blocker is fixed: reactivation now re-checks the active-key cap, and the new tests cover both below-cap and at-cap reactivation for app-session and channel keys.

Non-blocking cleanup: `sdk/ts/examples/example-app/src/components/WalletDashboard.tsx` still has one Auto Sign disable path that revokes by clearing `assets` while keeping the old future `expires_at`. The table/list revoke path and docs were updated correctly, so I would not block closure on this, but it is worth fixing so the example app does not report a successful on-chain revoke while leaving the latest key state active.

philanton:
Yep, good catch — missed the `handleDisableAutoSign` path. Fixed in [48b7bcc](https://github.com/layer-3/nitrolite/pull/775/changes/48b7bccd). Same shape as the table/list one: keep `assets`, set `expires_at` to `now - 1s` so the submit lands as a revocation under the new semantics and the cap slot actually frees.

PR #780, opened 2026-05-22
MF3-H01: issue challenge_rescue on Challenged → Closed unconditionally

philanton:
## Summary

- Drop `HasSignedFinalize` skip in `HandleHomeChannelClosed`. Path-1 timeout close at version Y < F (when a node-signed Finalize exists locally) was silently skipping `challenge_rescue` and under-crediting the user for any receiver/release net accrued strictly above Y.
- Source rescue `prev` via `GetLastUserState` across both channel-attached and detached chain entries. `NewChallengeRescueState` now branches on `prev.HomeChannelID`: detached prev → append at `(prev.Epoch, prev.Version+1)` inheriting the ledger; in-channel prev → wrap to `(prev.Epoch+1, 0)` as before. Eliminates the deterministic state-ID collision the old skip was working around.
- Drop the redundant `epoch` filter from `SumNetTransitionAmountAfterVersion`. A channel's in-channel rows live at a single epoch; detached rows have `home_channel_id IS NULL`, already excluded by the channel predicate.

## Bug flow

1. User+node sign cooperative-close Finalize at version `F`; channel goes `Closing`.
2. Adversary challenges on-chain with old state at `Y < F` (intent != CLOSE per contract). Local status flips to `Challenged`. Auto-response disabled — operator must counter manually.
3. Operator misses challenge window. `closeChannel` path-1 fires: chain pays `userAlloc(Y)`, emits `ChannelClosed(prevState)`.
4. Reactor forwards version `Y` only. Handler sees `wasChallenged=true` AND `HasSignedFinalize=true` → **skips** rescue.
5. Net of receivers/releases in `(Y, F]` (real off-chain credits the node validated) plus any post-Finalize detached credits → lost.

## Fix verification

| Scenario | wasChallenged | Sum >Y | prev | Placement | Outcome |
|---|---|---|---|---|---|
| Open → cooperative (no challenge race) | false | — | — | — | Skip (Status was `Closing`, not `Challenged`) |
| Plain timeout, no Finalize | true | net (Y, head] | in-channel head | `(E+1, 0)` wrap | Unchanged from pre-fix |
| **Timeout + offchain Finalize (the bug)** | true | `userAlloc(F) − userAlloc(Y)` | detached tip `(E+1, M)` | `(E+1, M+1)` append | **Fixed**: user credited |
| Cooperative close racing challenge, no post-Finalize tx | true | 0 | Finalize at `(E, F)` | `(E+1, 0)` wrap, amt=0 | Harmless zero-rescue |
| Cooperative close racing challenge, post-Finalize tx exists | true | 0 | detached tip | `(E+1, M+1)` append, amt=0 | Harmless zero-rescue |

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` all packages pass
- [x] `TestHandleHomeChannelClosed_TimeoutAfterFinalize_AppendsRescue` pins the bug-fix path: detached prev + non-zero shortfall → rescue appended at `(prev.Epoch, prev.Version+1)` with `prev.UserBalance + shortfall`
- [x] `TestNewChallengeRescueState/Success - detached prev appends...` pins the constructor's append branch
- [x] Existing `ChallengeRescue_Squash` / `_NoCredits` / `_NegativeNet_ClampsToZero` updated for new prev source + signature; behavior preserved for the in-channel branch

### DIFF
--- a/nitronode/event_handlers/service.go
+++ b/nitronode/event_handlers/service.go
@@ -312,16 +312,12 @@ func (s *EventHandlerService) HandleHomeChannelChallenged(ctx context.Context, t
 // finalized and closed on-chain. It updates the channel status to Closed and sets the final state version.
 // Once closed, no further state updates are possible for this channel.
 //
-// Additionally, when the closing path was a challenge resolution (channel was Challenged
-// and no node-signed Finalize state exists for it), the handler issues a single
-// ChallengeRescue state for the user that squashes the sum of receiver-state credits
-// accrued during the challenge window into an off-channel ledger entry tied to the
-// closed channel's ID. When a node-signed Finalize already exists, the cooperative-close
-// path has already advanced the user to a fresh epoch via NextState(), and any receiver
-// credits accumulated post-Finalize live at (epoch+1, version=0..N) with HomeChannelID=nil
-// — issuing a rescue at (epoch+1, version=1) would either overwrite the lone receiver
-// credit or collide on deterministic state ID with an existing row, so the rescue is
-// skipped entirely in that case.
+// Additionally, when the channel was locally Challenged at the time of close, the handler
+// issues a ChallengeRescue state crediting the user the net receiver-minus-sender balance
+// accrued strictly above the closure version. The rescue runs unconditionally on the
+// Challenged → Closed transition: both path-1 (timeout on a stale candidate) and path-2
+// (cooperative close on a signed Finalize) routes pass through this branch, and the
+// constructor + prev source pick the correct placement for the rescue row.
 func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx core.ChannelHubEventHandlerStore, event *core.HomeChannelClosedEvent) error {
 	logger := log.FromContext(ctx)
 	chanID := event.ChannelID
@@ -368,23 +364,8 @@ func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx co
 	}
 
 	if wasChallenged {
-		// Post-Finalize cooperative close racing with an on-chain challenge: receiver
-		// credits issued after the Finalize live in a fresh epoch via NextState() and
-		// are not channel-attached, so the rescue path would either overwrite them or
-		// collide on deterministic state ID. Skip the rescue branch entirely.
-		finalized, err := tx.HasSignedFinalize(chanID)
-		if err != nil {
+		if err := s.issueChallengeRescue(ctx, tx, channel, event.StateVersion); err != nil {
 			return err
-		}
-		if finalized {
-			logger.Debug("skipping challenge_rescue for post-Finalize close",
-				"channelId", chanID,
-				"userWallet", channel.UserWallet,
-				"closureVersion", event.StateVersion)
-		} else {
-			if err := s.issueChallengeRescue(ctx, tx, channel, event.StateVersion); err != nil {
-				return err
-			}
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


nitronode/event_handlers/service.go
```diff
@@ -368,62 +364,47 @@ func (s *EventHandlerService) HandleHomeChannelClosed(ctx context.Context, tx co
 	}
 
 	if wasChallenged {
```
ihsraham, line 366: I would block on this because it makes the rescue path too broad. The audit finding only wants `challenge_rescue` for timeout closes where the emitted final state is not `CLOSE`; cooperative `CLOSE` after a challenge should not issue a rescue.

Right now the event does not carry `ChannelClosed.finalState.intent`, and this branch runs for every local `Challenged -> Closed` transition. Please forward `finalState.intent` from the reactor into `HomeChannelClosedEvent`, gate this call on `intent != CLOSE`, and add a regression that cooperative `CLOSE` does not store or record a rescue.

ihsraham (changes requested):
I would keep this in request changes for H-01. The old `HasSignedFinalize` skip is gone, and the detached-chain placement looks directionally right, but the close event still does not carry `ChannelClosed.finalState.intent`. Without that, the handler cannot distinguish a timeout close from a cooperative `CLOSE` close.

I left the blocking note inline on the unconditional rescue branch. The path I would want before approval is: forward `finalState.intent` from the reactor into `HomeChannelClosedEvent`, gate rescue on non-`CLOSE` final state intent, and add a regression showing cooperative `CLOSE` after a local challenge does not store or record a rescue.

Checks run locally: `go test ./...`, targeted handler/store/core/reactor tests, `go vet ./...`, `go build ./...`, `forge test`, diff whitespace check, diff secret scan, `govulncheck`, `gosec`, `npm audit`, and `slither`. The security tools reported existing repo/dependency noise, but not a PR-specific issue that changes this H-01 verdict.

nitronode/event_handlers/service.go
philanton: Hmmm, I'd push back here. Cooperative `CLOSE` after challenge is already a zero-rescue under the current logic, by the SQL predicate rather than by intent:

1) `closureVersion` is `F` (the Finalize version the operator counter-submitted);
2) `SumNetTransitionAmountAfterVersion` filters on `home_channel_id = ?`. Post-Finalize rows are detached (`home_channel_id IS NULL`), so they're excluded;
3) In-channel rows for the channel live at versions ≤ `F`, so the sum strictly above `F` is `0`.

Net = 0 → `total = 0` → harmless zero-rescue. That's the "Cooperative close racing challenge" row in the PR description's table — the predicate already collapses it, the intent gate would be redundant.

Plumbing `finalState.intent` through the reactor into `HomeChannelClosedEvent` is a broader event-shape change than H-01 calls for, and I'd rather keep this PR scoped to the audit finding. The stored zero-amount row being slightly noisy in the audit trail is fair, but I don't see a correctness gap that blocks here.

ihsraham: Good point, I agree the SQL predicate makes cooperative `CLOSE` collapse to a zero-rescue under the current invariants: `closureVersion` is `F`, post-Finalize rows are detached, and `CheckActiveChannel` excludes `Closing` / `Challenged`, so normal in-channel rows above `F` are not accepted.

Non-blocking: worth adding a regression for that exact cooperative-close race so this stays pinned. Something like local `Challenged` status + cooperative `CLOSE` at `F` + detached post-Finalize rows, then assert the rescue amount is zero / no balance increase. I do not think this should block H-01.

ihsraham (approved):
Approved for H-01. Anton's explanation resolves my earlier blocker: the sum predicate already collapses cooperative `CLOSE` after challenge to zero while preserving the timeout-rescue path.

I left one non-blocking test suggestion in the thread to pin that invariant.

PR #801, opened 2026-06-01
MF3-H01: fix(nitronode): fix pagination.limit=0 DoS (divide-by-zero)

nksazonov:
## Summary

- `pagination.limit = 0` previously reached `calculatePaginationMetadata()` and triggered an integer division-by-zero panic, crashing the nitronode process. Reproducible with a single unauthenticated WebSocket request to `user.v1.get_transactions`.
- `GetOffsetAndLimit` (`pkg/core/types.go`) now coerces `limit = 0` to the server default, consistent with how other list endpoints (`get_channels`, `get_last_key_states`) already handle it.
- `calculatePaginationMetadata` (`nitronode/store/database/utils.go`) returns an explicit error instead of dividing unconditionally, protecting any future caller that bypasses `GetOffsetAndLimit`.
- Three store functions (`GetUserTransactions`, `GetApps`, `GetAppSessions`) propagate the error.
- Regression tests added for all three endpoints: `ZeroLimit` (coerces to default, succeeds) and `NilLimit` (absent limit, succeeds).
- `docs/api.yaml`: `pagination_params.limit` description updated to document the `0 = use default` behavior.

## Affected endpoints

- `user.v1.get_transactions`
- `apps.v1.get_apps`
- `app_sessions.v1.get_app_sessions`

## Summary by CodeRabbit

* **Documentation**
  * Clarified API documentation for pagination limit parameter behavior when set to 0.

* **Bug Fixes**
  * Improved pagination limit handling to treat a value of 0 as equivalent to omitting the parameter, ensuring consistent default behavior.
  * Added error handling for pagination metadata calculations.

* **Tests**
  * Added comprehensive test coverage for pagination behavior with nil and zero limit values across multiple endpoints.

### DIFF
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -494,7 +494,7 @@ types:
           optional: true
         - name: limit
           type: integer
-          description: Number of items to return
+          description: Number of items to return. A value of 0 is treated the same as omitting the field — the server default is used.
           optional: true
         - name: sort
           type: string
--- a/nitronode/api/user_v1/get_transactions_test.go
+++ b/nitronode/api/user_v1/get_transactions_test.go
@@ -119,6 +119,76 @@ func TestGetTransactions_Success(t *testing.T) {
 	mockStore.AssertExpectations(t)
 }
 
+// TestGetTransactions_NilLimit verifies that omitting pagination.limit (nil) uses the default
+// and succeeds without error.
+func TestGetTransactions_NilLimit(t *testing.T) {
+	mockStore := new(MockStore)
+
+	handler := &Handler{
+		store: mockStore,
+	}
+
+	wallet := "0x1234567890123456789012345678901234567890"
+
+	// Pagination present but Limit omitted — should reach the store with default behavior.
+	mockStore.On(
+		"GetUserTransactions",
+		wallet,
+		(*string)(nil),
+		(*core.TransactionType)(nil),
+		(*uint64)(nil),
+		(*uint64)(nil),
+		&core.PaginationParams{Limit: nil},
+	).Return([]core.Transaction{}, core.PaginationMetadata{Page: 1, PerPage: 50, PageCount: 0, TotalCount: 0}, nil)
+
+	reqPayload := rpc.UserV1GetTransactionsRequest{
+		Wallet:     wallet,
+		Pagination: &rpc.PaginationParamsV1{}, // Limit is nil
+	}
+	payload, err := rpc.NewPayload(reqPayload)
+	require.NoError(t, err)
+
+	ctx := &rpc.Context{
+		Context: context.Background(),
+		Request: rpc.Message{Method: "user.v1.get_transactions", Payload: payload},
+	}
+
+	handler.GetTransactions(ctx)
+
+	require.Nil(t, ctx.Response.Error())
+	mockStore.AssertExpectations(t)
+}
+
+// TestGetTransactions_ZeroLimit verifies that pagination.limit == 0 is rejected without panicking.
+func TestGetTransactions_ZeroLimit(t *testing.T) {
+	mockStore := new(MockStore)
+
+	handler := &Handler{
+		store: mockStore,
+	}
+
+	zero := uint32(0)
+	reqPayload := rpc.UserV1GetTransactionsRequest{
+		Wallet: "0x1234567890123456789012345678901234567890",
+		Pagination: &rpc.PaginationParamsV1{
+			Limit: &zero,
+		},
+	}
+	payload, err := rpc.NewPayload(reqPayload)
+	require.NoError(t, err)
+
+	ctx := &rpc.Context{
+		Context: context.Background(),
+		Request: rpc.Message{Method: "user.v1.get_transactions", Payload: payload},
+	}
+
+	handler.GetTransactions(ctx)
+
+	require.NotNil(t, ctx.Response.Error())
+	assert.Contains(t, ctx.Response.Error().Error(), "limit must be greater than 0")
+	mockStore.AssertExpectations(t) // store must not be called
+}
+
 // TestGetTransactions_NormalizesWallet verifies the wallet is normalized before the store call.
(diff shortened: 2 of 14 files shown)


nksazonov:
@coderabbitai review

pkg/core/types.go
```diff
@@ -1222,7 +1223,7 @@ func (p *PaginationParams) GetOffsetAndLimit(defaultLimit, maxLimit uint32) (off
 		if p.Offset != nil {
 			offset = *p.Offset
 		}
-		if p.Limit != nil {
+		if p.Limit != nil && *p.Limit > 0 {
```
philanton, line 1226: This `> 0` check is the actual fix, but `TestPaginationParams_GetOffsetAndLimit` wasn't updated to cover it. I'd add a case here: `Limit: &0` → expect `defaultLimit` back. That's the one line that really defends the node, so it'd be good to pin it.

philanton (commented):
Nice job. Fix looks solid — root cause is gone, and the other divide-by-`limit` paths (`get_channels`, both `get_last_key_states`) already guard `limit > 0`, so the scope is right.

One thing on the tests though: the `*_ZeroLimit` cases mock the store, so they don't actually run the code that used to panic. `GetOffsetAndLimit` coercion and `calculatePaginationMetadata`'s divide both live below the mock, so these tests would pass even without the fix. Left a few inline notes on where I'd add real coverage.

nitronode/store/database/utils.go
```diff
@@ -172,18 +172,20 @@ func toCoreTransaction(dbTx *Transaction) *core.Transaction {
...
-// calculatePaginationMetadata computes pagination metadata from total count, offset, and limit
-func calculatePaginationMetadata(totalCount int64, offset, limit uint32) core.PaginationMetadata {
-	pageCount := uint32((totalCount + int64(limit) - 1) / int64(limit))
-	currentPage := uint32(1)
-	if limit > 0 {
-		currentPage = offset/limit + 1
+// calculatePaginationMetadata computes pagination metadata from total count, offset, and limit.
+// Returns an error if limit is 0, since a zero-sized page has no valid metadata representation.
+func calculatePaginationMetadata(totalCount int64, offset, limit uint32) (core.PaginationMetadata, error) {
+	if limit == 0 {
```
philanton, line 178: This error branch has no test, and it's actually unreachable from the three handlers now — `GetOffsetAndLimit` already coerces to a non-zero limit before we get here. Fine to keep as defense-in-depth, but a direct `calculatePaginationMetadata(_, _, 0)` test would lock in the no-panic guarantee.

nitronode/api/user_v1/get_transactions_test.go
```diff
@@ -119,6 +119,89 @@ func TestGetTransactions_Success(t *testing.T) {
...
+	mockStore := new(MockStore)
+
+	handler := &Handler{
+		store: mockStore,
+	}
+
+	wallet := "0x1234567890123456789012345678901234567890"
+	zero := uint32(0)
+
+	// limit=0 passes through to the store; GetOffsetAndLimit coerces it to DefaultLimit.
```
philanton, line 174: This one (and the other two `_ZeroLimit` tests) mocks the store, so the divide path never runs here — the test passes regardless of the fix. I'd either drop a store-level test in `store/database/test` hitting the real path with `limit=0`, or unit-test `GetOffsetAndLimit`/`calculatePaginationMetadata` directly. As is, they don't really guard against the regression coming back.

pkg/core/types.go
ihsraham: +1, this is the key behavior to pin for H-01 closure.

nitronode/store/database/utils.go
ihsraham: +1, the guard is useful defense in depth, but it should have a direct no-panic/error test.

nitronode/api/user_v1/get_transactions_test.go
ihsraham: +1, the code path looks safe, but this coverage is what would prove the original panic cannot regress.

docs/api.yaml
```diff
@@ -494,7 +494,7 @@ types:
           optional: true
         - name: limit
           type: integer
-          description: Number of items to return
+          description: Number of items to return. A value of 0 is treated the same as omitting the field — the server default is used.
```
ihsraham, line 497: This behavior is fine for the DoS fix, since `0` becomes a bounded default page. Since the finding asked to reject `0`, could we call out in the closure response that this is intentional normalization instead of rejection?

nksazonov: Intentional — we reviewed both options and chose normalization over rejection for consistency with the existing  and  endpoints, which already silently coerce `limit=0` to the default. Rejecting on three endpoints but coercing on the other two would give callers inconsistent behavior depending on which endpoint they hit. The PR description notes the trade-off. The `calculatePaginationMetadata` error return (unreachable via the normal path) still guarantees no panic if any future caller bypasses `GetOffsetAndLimit`.

pkg/core/types.go
nksazonov: 8764bd60

nitronode/store/database/utils.go
nksazonov: 8764bd60

nitronode/api/user_v1/get_transactions_test.go
nksazonov: 8764bd60

philanton (approved):

ihsraham (approved):
Approving for H-01. The zero-limit panic path is closed now: limit=0 is normalized before the store divide path, calculatePaginationMetadata has a defensive zero-limit error, and the docs call out the normalization behavior.

The added direct helper tests cover the regression point that was missing from the first pass. The mocked handler tests are fine as API-shape coverage now, since they are no longer the only proof for the original crash.

PR #802, opened 2026-06-01
MF3-M04: fix(nitronode): fix uint8 overflow in quorum weight accumulators

nksazonov:
## Summary

- `totalWeights` in `CreateAppSession` and `achievedQuorum` in `verifyQuorum` were declared as `uint8`, causing silent modulo-256 wrap-around when participant weights summed above 255
- Widened both accumulators to `uint16` (sufficient for up to 32 participants × 255 weight = 8160)
- No external types changed — `SignatureWeight` and `Quorum` remain `uint8` at the wire/storage layer

## Test plan

- [ ] `TestCreateAppSession_TotalWeightsOver255` — creation succeeds when real total > 255 (previously false-rejected)
- [ ] `TestCreateAppSession_TotalWeightsWrapToZero` — 128+128=256 wrap-to-zero boundary
- [ ] `TestCreateAppSession_QuorumExceedsRealTotalOver255_Rejected` — guard still rejects quorum > real total
- [ ] `TestSubmitAppState_VerifyQuorumWeightOver255` — valid coalition with combined weight > 255 passes
- [ ] `TestSubmitAppState_VerifyQuorumWrapToZero` — 128+128 boundary through SubmitAppState
- [ ] `TestSubmitDepositState_VerifyQuorumWeightOver255` — same coverage through SubmitDepositState
- [ ] `TestRebalanceAppSessions_VerifyQuorumWeightOver255` — same coverage through RebalanceAppSessions

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed quorum verification to correctly handle participant signature weights exceeding 255, ensuring app sessions are properly validated regardless of total weight magnitude.

* **Tests**
  * Added comprehensive test coverage for quorum verification edge cases and weight overflow scenarios across multiple operations.

### DIFF
--- a/nitronode/api/app_session_v1/create_app_session.go
+++ b/nitronode/api/app_session_v1/create_app_session.go
@@ -75,7 +75,7 @@ func (h *Handler) CreateAppSession(c *rpc.Context) {
 	// derivation, and stored participant list all see the canonical (lowercase, 0x-prefixed)
 	// representation regardless of how the caller cased the address or whether they
 	// included the 0x prefix.
-	var totalWeights uint8
+	var totalWeights uint16
 	participantWeights := make(map[string]uint8)
 	for i, participant := range reqPayload.Definition.Participants {
 		participantWallet, err := core.NormalizeHexAddress(participant.WalletAddress)
@@ -89,12 +89,12 @@ func (h *Handler) CreateAppSession(c *rpc.Context) {
 			c.Fail(rpc.Errorf("duplicate participant address: %s", participant.WalletAddress), "")
 			return
 		}
-		totalWeights += participant.SignatureWeight
+		totalWeights += uint16(participant.SignatureWeight)
 		participantWeights[participantWallet] = participant.SignatureWeight
 		appDef.Participants[i].WalletAddress = participantWallet
 	}
 
-	if reqPayload.Definition.Quorum > totalWeights {
+	if uint16(reqPayload.Definition.Quorum) > totalWeights {
 		c.Fail(rpc.Errorf("target quorum (%d) cannot be greater than total sum of weights (%d)",
 			reqPayload.Definition.Quorum, totalWeights), "")
 		return
--- a/nitronode/api/app_session_v1/create_app_session_test.go
+++ b/nitronode/api/app_session_v1/create_app_session_test.go
@@ -1170,6 +1170,88 @@ func TestCreateAppSession_AppRegistryDisabled(t *testing.T) {
 	mockStore.AssertExpectations(t)
 }
 
+// TestCreateAppSession_TotalWeightsOver255 verifies that session creation succeeds when the
+// real sum of participant weights exceeds 255 but the quorum is still achievable. With a
+// uint8 accumulator the sum wraps modulo 256, which can make a valid quorum look unreachable
+// at creation time (e.g. weights 200+200=400 wraps to 144, and quorum=200 falsely appears
+// to exceed total). The accumulator must be at least uint16 to avoid this.
+func TestCreateAppSession_TotalWeightsOver255(t *testing.T) {
+	mockStore := new(MockStore)
+	storeTxProvider := func(fn StoreTxHandler) error {
+		return fn(mockStore)
+	}
+
+	mockSigner := NewMockChannelSigner()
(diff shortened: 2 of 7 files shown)


nksazonov:
@coderabbitai review

nitronode/api/app_session_v1/create_app_session.go
```diff
@@ -75,7 +75,7 @@ func (h *Handler) CreateAppSession(c *rpc.Context) {
 	// derivation, and stored participant list all see the canonical (lowercase, 0x-prefixed)
 	// representation regardless of how the caller cased the address or whether they
 	// included the 0x prefix.
-	var totalWeights uint8
+	var totalWeights uint16
```
philanton, line 78: `uint16` only holds the sum up to 257 participants (257×255 = 65535). `NITRONODE_MAX_PARTICIPANTS` is a plain `int`, default 32, with no upper bound — set it ≥ 258 and `totalWeights` wraps again, same bug class. This is the second half of the finding's recommendation ("use such maxParticipants value that the maximum weight fits uint16"), and it's not enforced anywhere. Default-safe + operator-only, so low severity, but I'd add a startup validation that `MaxParticipants` ≤ 257 (or just ≤ 32) over in `runtime.go`/`main.go` so the bound can't drift. Cheap to close and it's exactly what the finding asked for.

nitronode/api/app_session_v1/create_app_session_test.go
```diff
@@ -1231,3 +1313,139 @@ func TestCreateAppSession_DuplicateParticipantAcrossCases(t *testing.T) {
...
+		t.Fatalf("total weights 256 (128+128) with quorum 128 must be accepted, got: %v", respErr)
+	}
+	assert.Equal(t, rpc.MsgTypeResp, ctx.Response.Type)
+	mockStore.AssertExpectations(t)
+}
+
+// TestCreateAppSession_QuorumExceedsRealTotalOver255_Rejected verifies that when total
+// weights exceed 255, a quorum genuinely larger than the real total is still rejected.
+// (The fix must not disable the quorum>total guard for the >255 range.)
+func TestCreateAppSession_QuorumExceedsRealTotalOver255_Rejected(t *testing.T) {
```
philanton: The name says `Over255` but this uses total 100+100=200 — and as the inline comment admits, it can't actually test the >255 range because `Quorum` is `uint8` (capped at 255). I'd drop `Over255` from the name. Worth noting the underlying thing: since `Quorum` stays `uint8` while `totalWeights` can now reach 8160, the `quorum > totalWeights` guard can never fire once total ≥ 255 — not a bug, just dead in that range, and it means you can't require a threshold above 255 even with large weights. That's a pre-existing `uint8` `Quorum` limit, not something this PR introduces, but can you confirm it's intended?

philanton (commented):
Nice job, fix is correct and tests cover the real path. Two notes inline.

ihsraham (commented):
Looks close overall. The main accumulator fix is doing the right thing: `CreateAppSession` and shared `verifyQuorum` now sum in `uint16`, I did not find another production `SignatureWeight` quorum accumulator, and the real overflow regressions fail on the base branch and pass here across create, submit app state, submit deposit state, and rebalance.

The only thing I would keep open for M-04 closure is Anton's `MaxParticipants` note. The audit asked for the accumulator type and the participant bound to agree, so this should either reject `NITRONODE_MAX_PARTICIPANTS > 257` at startup or use a wider accumulator. The default/charts are at 32, so I do not see a default-path exploit, but I would still close that before marking the audit item fixed.

Tooling-wise, Snyk is green and there are no manifest changes in this PR. Local `go test ./...`, `go test -race ./nitronode/api/app_session_v1`, `go vet ./nitronode/...`, `go mod verify`, `gofmt -l`, and `git diff --check` were clean.

nitronode/api/app_session_v1/create_app_session.go
nksazonov: 36fa562f

nitronode/api/app_session_v1/create_app_session_test.go
nksazonov: 36fa562f

PR #803, opened 2026-06-01
MF-L02: fix(nitronode): lock user balance row in HandleHomeChannelCreated + backfill unsigned receiver states

nksazonov:
## Summary

- Adds `LockUserState` to `HandleHomeChannelCreated` before the channel status flip, serializing it against concurrent receiver/release issuance RPCs and closing the race where an unsigned receiver state could land on a newly-opened channel
- Adds `backfillOffChainHeadNodeSig` at the end of the handler to node-sign any unsigned `transfer_receive`/`release` head that arrived while the channel was still Void
- Updates the comment and warning log in `backfillOffChainHeadNodeSig` to be caller-agnostic (previously described only the challenge-clearance path; function now has two callers)
- Adds 5 tests: lock ordering, backfill with signature verification, no-backfill when head already signed, updated success and replay tests

## Test plan

- [ ] `go test ./nitronode/event_handlers/...` passes
- [ ] `TestHandleHomeChannelCreated_AcquiresUserLockBeforeMutation` — pins lock-before-mutation ordering
- [ ] `TestHandleHomeChannelCreated_BackfillsUnsignedReceiverStateOnOpen` — verifies node sig produced and valid
- [ ] `TestHandleHomeChannelCreated_HeadAlreadySigned_NoBackfill` — verifies no double-sign

### DIFF
--- a/nitronode/event_handlers/service_test.go
+++ b/nitronode/event_handlers/service_test.go
@@ -44,6 +44,7 @@ func TestHandleHomeChannelCreated_Success(t *testing.T) {
 
 	// Mock expectations
 	mockStore.On("GetChannelByID", channelID).Return(channel, nil)
+	mockStore.On("LockUserState", userWallet, "usdc").Return(decimal.Zero, nil)
 	mockStore.On("UpdateChannel", mock.MatchedBy(func(ch core.Channel) bool {
 		return ch.ChannelID == channelID &&
 			ch.Status == core.ChannelStatusOpen &&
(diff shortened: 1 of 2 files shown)


nksazonov:
@coderabbitai review

philanton (approved):
lgtm. Good catch that the lock alone wouldn't fully fix it — pairing it with the node-sig backfill is the right call, and the tests exercise the real path.

nitronode/event_handlers/service_test.go
```diff
@@ -111,6 +116,207 @@ func TestHandleHomeChannelCreated_IgnoresReplayOnInitializedChannel(t *testing.T
...
+
+	// Unsigned transfer_receive credit stored before the channel opened.
+	headState := &core.State{
+		ID:            core.GetStateID(userWallet, asset, 0, headVersion),
+		Asset:         asset,
+		UserWallet:    userWallet,
+		Epoch:         0,
+		Version:       headVersion,
+		HomeChannelID: &homeChannelIDPtr,
+		Transition:    core.Transition{Type: core.TransitionTypeTransferReceive},
```
ihsraham: non-blocking: this test covers the `transfer_receive` head, which is the main path for L-02. since `backfillOffChainHeadNodeSig` also allows `TransitionTypeRelease`, it would be worth table-driving this with a `release` head too so both allowed backfill transitions are pinned by the created-event regression test.

ihsraham (approved):
approved. @philanton already called out the key part: the lock alone was not enough, and the node-sig backfill closes the remaining window.

no blockers from me. i left one non-blocking coverage note for the `release` case.

nitronode/event_handlers/service_test.go
nksazonov: 597dbd076f36098232d95440329c56f72820af2d

PR #806, opened 2026-06-02
MF3-M01, M02, M03, L09: feat(app-sessions): remove rebalance feature

philanton:
## Summary

Removes the app-session **rebalancing** feature end to end. The endpoint was already disabled in every deployed environment (`NITRONODE_MAX_SIGNED_UPDATES=0`, route gated on `>= 2`), so this is effectively dead-code cleanup at runtime plus full removal of the API surface, SDK methods, types, and docs.

## Changes by layer

**nitronode**
- Delete `rebalance_app_sessions.go` handler + its 1358-line test
- Drop `MaxSignedUpdates` / `MaxRebalanceSignedUpdates` config and the route gate; remove `NITRONODE_MAX_SIGNED_UPDATES` from the 3 helm configs
- Fix `NewHandler` signature + 35 call sites

**pkg/**
- `rpc`: remove `rebalance_app_sessions` method const, request/response types, client wrapper
- `app`: remove `AppStateUpdateIntentRebalance`, `AppSessionVersionV1`, `GenerateRebalanceBatchIDV1`, `GenerateRebalanceTransactionIDV1`
- `core`: remove `TransactionTypeRebalance` (42)

**SDKs**
- `sdk/go`, `sdk/ts`: drop client methods, types, packing helpers, enum members, examples, READMEs
- Regenerate `rpc-drift` and `public-api-drift` snapshots

**Docs / misc**
- `docs/api.yaml`, `docs/data_models.mmd`, `llms-full.txt`, MCP categorizer regex, playground history view

## Notes

- The intent (`4`) and transaction-type (`42`) enum values were **terminal**, so removal shifts no other wire values.
- `TransactionTypeRebalance=42` is removed entirely — pre-existing ledger rows tagged `42` (if any) now render as `unknown`. Chosen deliberately over keeping a display-only mapping.

## Verification

- `go vet ./...` clean · `go test ./...` all pass
- TS SDK: 184 tests pass, snapshots regenerated · `tsc` + ts-compat typecheck OK
- MCP build OK · playground typecheck OK

## Release Notes

* **Breaking Changes**
  * Removed `rebalance_app_sessions` RPC endpoint for atomic multi-session rebalancing
  * Removed rebalance methods from Go and TypeScript SDKs
  * Removed "rebalance" as a supported transaction type and app state update intent
  * Removed related configuration parameters

* **Documentation**
  * Updated API specifications, SDK documentation, and deployment guides to reflect removal of rebalancing functionality


nksazonov (changes requested):
Clean approach and thorough removal across all layers. However, the `NewHandler` signature change was partially applied — 5 `NewHandler` call sites and 1 struct literal still reference the removed `maxSignedUpdates` parameter, causing `go test ./nitronode/api/app_session_v1/...` to fail to compile. These must be fixed before this can land.

---

**[Bug — compile failure] `create_app_session_test.go` lines 1198, 1339, 1413**

`TestCreateAppSession_TotalWeightsOver255`, `TestCreateAppSession_TotalWeightsWrapToZero`, and `TestCreateAppSession_QuorumExceedsTotalWeights_Rejected` all pass five integer arguments to `NewHandler` — `32, 1024, 256, 16, 100` — but the signature now takes four (`maxParticipants`, `maxSessionData`, `maxSessionKeyIDs`, `maxSessionKeysPerUser`). The `maxSignedUpdates` arg (`16`) was not removed here during the bulk update. Compiler error: `too many arguments in call to NewHandler`.

**Suggestion:** Remove the `16` from each call so they read `32, 1024, 256, 100`.

---

**[Bug — compile failure] `submit_app_state_test.go` lines 2160, 2272**

`TestSubmitAppState_VerifyQuorumWeightOver255` and `TestSubmitAppState_VerifyQuorumWeightWrapToZero` have the same stale five-argument `NewHandler` call. Same compiler error.

**Suggestion:** Remove the `16` from each call so they read `32, 1024, 256, 100`.

---

**[Bug — compile failure] `submit_deposit_state_test.go` line 1023**

`TestSubmitDepositState_VerifyQuorumWeightOver255` initialises `Handler` via a struct literal rather than `NewHandler`, so it was not caught by the bulk update. `maxSignedUpdates` no longer exists as a field on `Handler`. Compiler error: `unknown field maxSignedUpdates in struct literal of type Handler`.

**Suggestion:** Remove the `maxSignedUpdates: 16,` line from the struct literal.

---

**[Doc] `nitronode/README.md` line 34**

The `app_session_v1` entry in the API groups section still lists "Rebalancing" as a capability. Line 9 was correctly cleaned up, but this one was missed.

**Suggestion:** Change "Advanced application session management (Creation, Deposits, Rebalancing)" to remove "Rebalancing".

docs/api.yaml
```diff
@@ -297,7 +297,6 @@ types:
         - escrow_lock
         - escrow_withdraw
         - migrate
-        - rebalance
         - finalize
```
ihsraham, line 300: Runtime removal looks right, but this enum is also the response contract for old transaction rows. After `TransactionTypeRebalance` is removed, legacy `tx_type=42` rows stringify as `unknown`, so strict clients generated from this schema can reject historical `get_transactions` responses.

Could we either document `unknown` for response rows here, or split request and response transaction types so request filtering stays limited while legacy rows remain valid?

ihsraham (commented):
The rebalance removal itself looks like the right closure path for M-01/M-02/M-03/L-09. I did not find a remaining callable rebalance RPC, SDK, or MCP surface. +1 to Nikita's compile blocker; leaving one extra release-contract note inline.

dimast-x (approved):

docs/api.yaml
philanton: Yeah, good catch. I went with adding `unknown` to the `transaction_type` enum (your first option) rather than splitting request/response — the split is cleaner, but felt a bit heavy for a single terminal fallback, and the filter side just treats `unknown` as a no-op. Fixed in [5129c80](https://github.com/layer-3/nitrolite/pull/806/changes/5129c805f65257d70dcd0845f6a23f4f10df46d3).

philanton:
@nksazonov you were right — after I rebased onto main the `NewHandler` cleanup turned out incomplete and the package didn't compile. Fixed in [6a77ba5](https://github.com/layer-3/nitrolite/pull/806/changes/6a77ba53173735e0f0f67bb1ec3e420c67844394): dropped the stale `maxSignedUpdates` arg from the 5 call sites and the struct literal in `submit_deposit_state_test.go`. While there I also removed `unmapSignedAppStateUpdateV1`, which was left dangling after the rebalance handler deletion.

README rebalance line on L34 fixed in [5129c80](https://github.com/layer-3/nitrolite/pull/806/changes/5129c805f65257d70dcd0845f6a23f4f10df46d3).

`go build/vet ./...` clean and the app_session_v1 tests pass locally now. Mind taking another look?)

nksazonov (approved):

PR #807, opened 2026-06-02
MF3-L01: fix(core): align transaction ID with referencing transition TxID

philanton:
## Finding (MF3-L01)

Each transition stores a `txID` pointing to its transaction. The ID `NewTransactionFromTransition` generated did **not** match that `txID` for `HomeDeposit`, `TransferSend`, and `TransferReceive` (and, missed by the finding, `EscrowDeposit`).

## Root cause

Two ID conventions collided:

- **Transition `TxID`** (set by `Apply*`, validated in `state_advancer.go` via `Equal`) = `getTransactionID(transition.AccountID, stateID)`.
- **Transaction `ID`** (recomputed in `NewTransactionFromTransition`) = `getTransactionID(toAccount|fromAccount, stateID)`.

They matched only when `AccountID` equalled `toAccount`/`fromAccount`. They diverged for:

| Transition | transition.TxID | old recomputed ID |
|---|---|---|
| HomeDeposit | (HomeChannelID, sID) | (UserWallet, sID) |
| EscrowDeposit | (EscrowChannelID, sID) | (UserWallet, sID) |
| TransferSend | (recipient, sID) | (UserWallet, receiverID) |
| TransferReceive | shares Send's TxID | — |

TransferSend/Receive share one TxID (`channel_v1/handler.go:108`), so a transfer must produce a single transaction referenced by both. The old code produced an ID neither transition referenced → dangling `tx_id`.

## Fix

`NewTransactionFromTransition` now uses `transition.TxID` directly as the transaction ID (rejecting empty). The `TxID` is canonicalised and validated in the state advancer, so it is the single source of truth and cannot be forged by a client. Fixes all four cases by construction and drops the fragile parallel recompute.

- `pkg/core/types.go` — use `transition.TxID`.
- `nitronode/store/database/transaction.go` — doc comment updated to the new convention.
- `pkg/core/types_test.go` — assert `tx.ID == transition.TxID` for HomeDeposit / EscrowDeposit / TransferSend, plus empty-TxID rejection.

## Notes

- Pre-fix rows carry the old (wrong) IDs. No backfill included — history intentionally out of scope.

## Testing

- `go test ./pkg/core/... ./nitronode/...` — pass
- `go vet` — clean

### DIFF
--- a/pkg/core/types.go
+++ b/pkg/core/types.go
@@ -928,16 +928,18 @@ func NewTransactionFromTransition(senderState *State, receiverState *State, tran
 	}
 
 	var receiverStateID *string
-	var txID string
-	var err error
 	if receiverState != nil {
 		receiverStateID = &receiverState.ID
-		txID, err = GetReceiverTransactionID(fromAccount, receiverState.ID)
-	} else {
-		txID, err = GetSenderTransactionID(toAccount, senderState.ID)
 	}
-	if err != nil {
-		return nil, err
+
+	// The transaction ID must equal the transition's TxID so that the transition
+	// row references this transaction. For transfers, the sender's TransferSend and
+	// the receiver's TransferReceive share the same TxID, so both must point at this
+	// single transaction. The TxID is canonicalised and validated in the state
+	// advancer, making it the single source of truth.
+	txID := transition.TxID
+	if txID == "" {
+		return nil, fmt.Errorf("transition has empty txID")
 	}
 
 	return NewTransaction(
--- a/pkg/core/types_test.go
+++ b/pkg/core/types_test.go
@@ -580,6 +580,14 @@ func TestNewTransactionFromTransition(t *testing.T) {
 	assert.Equal(t, TransactionTypeTransfer, tx.TxType)
 	assert.Equal(t, transferTxID, tx.ID)
 
+	// The receiver's TransferReceive transition must carry the same TxID as the
+	// sender's TransferSend, so both states reference the single transfer
+	// transaction recorded above.
+	receiveTransition, err := receiverState.ApplyTransferReceiveTransition(senderState.UserWallet, decimal.NewFromInt(10), transferTxID)
+	require.NoError(t, err)
+	assert.Equal(t, transferTxID, receiveTransition.TxID)
+	assert.Equal(t, transition.TxID, receiveTransition.TxID)
+
 	// TransferSend error: nil receiverState
 	_, err = NewTransactionFromTransition(senderState, nil, transition)
 	assert.Error(t, err)
--- a/pkg/core/types_test.go
+++ b/pkg/core/types_test.go
@@ -550,23 +550,46 @@ func TestNewTransactionFromTransition(t *testing.T) {
 	senderState.EscrowChannelID = new(string)
 	*senderState.EscrowChannelID = "EC"
 
-	// HomeDeposit
-	transition := Transition{Type: TransitionTypeHomeDeposit, Amount: decimal.NewFromInt(10)}
+	// HomeDeposit: transaction ID must equal the transition's TxID,
+	// computed over the HomeChannelID (AccountID), not the UserWallet.
+	homeDepositTxID, err := GetSenderTransactionID(*senderState.HomeChannelID, senderState.ID)
+	require.NoError(t, err)
+	transition := Transition{Type: TransitionTypeHomeDeposit, Amount: decimal.NewFromInt(10), AccountID: *senderState.HomeChannelID, TxID: homeDepositTxID}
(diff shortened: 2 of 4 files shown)


pkg/core/types.go
```diff
@@ -928,16 +928,18 @@ func NewTransactionFromTransition(senderState *State, receiverState *State, tran
...
 	}
-	if err != nil {
-		return nil, err
+
+	// The transaction ID must equal the transition's TxID so that the transition
+	// row references this transaction. For transfers, the sender's TransferSend and
+	// the receiver's TransferReceive share the same TxID, so both must point at this
+	// single transaction. The TxID is canonicalised and validated in the state
+	// advancer, making it the single source of truth.
+	txID := transition.TxID
```
ihsraham, line 940: This is the right source of truth for L-01. The submit path validates the proposed transition before this helper records the transaction, so using `transition.TxID` here makes the transaction row match the transition row instead of recomputing a different ID convention.

pkg/core/types_test.go
```diff
@@ -550,24 +550,47 @@ func TestNewTransactionFromTransition(t *testing.T) {
...
+	assert.Equal(t, TransactionTypeEscrowDeposit, tx.TxType)
+	assert.Equal(t, escrowDepositTxID, tx.ID)
 
-	// TransferSend
-	transition = Transition{Type: TransitionTypeTransferSend, Amount: decimal.NewFromInt(10), AccountID: "REC"}
+	// TransferSend: sender and receiver transitions share this TxID, computed
+	// over the recipient (AccountID) and the sender's state ID.
+	transferTxID, err := GetSenderTransactionID("REC", senderState.ID)
+	require.NoError(t, err)
+	transition = Transition{Type: TransitionTypeTransferSend, Amount: decimal.NewFromInt(10), AccountID: "REC", TxID: transferTxID}
```
ihsraham, line 576: Non-blocking: this catches the old transfer ID mismatch, and it would be even tighter if the test also exercised the receiver state via `ApplyTransferReceiveTransition()` or added an advancer case where a forged `TxID` is rejected. That would pin the validation boundary now that `NewTransactionFromTransition` intentionally trusts the already-validated `transition.TxID`.

ihsraham (commented):
No blockers from me. The important part looks right: `NewTransactionFromTransition` now uses the validated transition `TxID`, so future transaction rows line up with the `tx_id` stored on the transition.

I left one non-blocking test note. The remaining historical-row question looks like a separate data/reporting decision, since this PR intentionally changes future writes only.

pkg/core/types_test.go
philanton: Yeah, good call — added both in [17451d3](https://github.com/layer-3/nitrolite/pull/807/changes/17451d36).

1) `TestValidateAdvancement_RejectsForgedTxID` in `state_advancer_test.go` proposes a valid `home_deposit`, forges only the `TxID`, and asserts the advancer rejects it with `transaction ID mismatch` — that pins the boundary `NewTransactionFromTransition` now trusts;
2) extended the helper test to run `ApplyTransferReceiveTransition()` and check the receiver's `TxID` equals the sender's, so the shared-TxID assumption is covered both sides.

nksazonov (approved):
Excellent work!

The fix is correct and well-tested. The trust chain is sound — `ValidateAdvancement` re-derives the canonical TxID via `Apply*Transition()`, then `Equal()` at `state_advancer.go:171` compares it against the proposed value before `NewTransactionFromTransition` ever runs, making `transition.TxID` fully justified as the source of truth. All four divergent cases (HomeDeposit, EscrowDeposit, TransferSend, TransferReceive) are addressed, and the forged-TxID and receiver-TxID-parity tests pin the invariants cleanly.

pkg/core/types.go
nksazonov: Fully agree. `transition.TxID` is the canonical value here — the advancer's `Equal()` check at `state_advancer.go:171` already rejected any forged or mismatched TxID before this code runs, so trusting it is safe by construction.

pkg/core/types_test.go
nksazonov: Both were added in `17451d3` and they're important. `TestValidateAdvancement_RejectsForgedTxID` pins that the advancer enforces TxID canonicity (the invariant `NewTransactionFromTransition` now relies on), and the receiver TxID parity check pins that `TransferSend` and `TransferReceive` converge on the same transaction row. Good call flagging those.

ihsraham (approved):
approved, the follow-up commit resolves my previous test note. I rechecked head `17451d3`; the forged-TxID advancer test pins the validation boundary before `NewTransactionFromTransition` trusts `transition.TxID`, and the receiver-side transfer check covers the shared-TxID path.

PR #808, opened 2026-06-02
MF3-C01: fix(nitronode): require enforcing deposit states before crediting off-chain

nksazonov:
## Summary

Fixes **MF3-C01** — a version-0 epoch/channel sentinel collision that let a node-signed `HomeDeposit` be credited off-chain before any deposit event landed.

After a `Finalize`, `NextState()` opened the new epoch at `version=0`. Newly created channels also persist as `state_version=0`. `EnsureNoOngoingStateTransitions` treats a `HomeDeposit` as settled when `state.version == channel.state_version`, so the two zeros collided and the gate let the user spend phantom funds.

## Changes

**Primary fix** — reserve `version=0` as the "no on-chain state materialised yet" sentinel:
- `pkg/core/types.go`: `NextState()` post-Finalize and `NewChallengeRescueState()` in-channel branch both emit `Version: 1` (was `0`).
- `sdk/ts/src/core/state.ts`: ma)`.

**Defence in depth** — gate mustl to be `Open`:
- `EnsureNoOngoingStateTransitions` SQL now selects `hc.status`; the `HomeDeposit` / `HomeWithdrawal` cases require `o the version match.
- `RefreshUserEnforcedBalance` SQL: `state_version > 0` → `status = 1` (removes the magic-number sentinel).
- New `Channel.IsEnforced()` helper encodes the convention.

### DIFF
--- a/nitronode/event_handlers/service_test.go
+++ b/nitronode/event_handlers/service_test.go
@@ -1633,7 +1633,9 @@ func TestHandleHomeChannelClosed_ChallengeRescue_Squash(t *testing.T) {
 	require.True(t, rescueAmount.Equal(capturedState.Transition.Amount))
 	require.Nil(t, capturedState.HomeChannelID, "rescue state must be off-channel")
 	require.Equal(t, prevState.Epoch+1, capturedState.Epoch)
-	require.Equal(t, uint64(0), capturedState.Version)
+	// Fresh-epoch in-channel rescue lands at version=1: version=0 is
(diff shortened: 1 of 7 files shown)


philanton (approved):

ihsraham (approved):
Approving for C-01 closure. The patch reserves version 0 for the unmaterialized channel sentinel and now requires HomeDeposit/HomeWithdrawal to match the home channel state version with status Open.

No blockers from me.

PR #809, opened 2026-06-02
MF3-M05: allow wallet-only revocation of session keys

philanton:
### Problem

`SubmitSessionKeyState()` (both app-session and channel paths) treated a submitted `expires_at` not after `now` as a revocation, but still **required and validated `session_key_sig`** before storing. A user could not unilaterally revoke a delegated key: if the session key was lost, unavailable, compromised, or held by a delegate who refused to sign the revocation payload, the user had to wait until the existing `expires_at` passed while the key stayed usable. Requiring `session_key_sig` is correct when activating or extending authority, but on revocation it handed the delegate a veto.

### Fix

Split the signature requirement by intent:

- **`expires_at > now`** (activation / extension / rotation) — still requires **both** `user_sig` and `session_key_sig`, validated as before.
- **`expires_at <= now`** (revocation) — requires **only `user_sig`**. Any `session_key_sig` present is ignored.

The packed payload binds `user_address`, `session_key`, `version`, and `expires_at`, so a wallet-only revocation is authorized for exactly that key/version and cannot be replayed. Revocation still flows through `LockSessionKeyState` (ownership binding) and the monotonic version check.

### Changes

- **`pkg/app`, `pkg/core`** — new `ValidateAppSessionKeyStateUserSigV1` / `ValidateChannelSessionKeyStateUserSigV1`; the both-signature validators now reuse them.
- **nitronode handlers** — branch on `expires_at` to pick the validator. `now` is captured once and reused for the in-transaction cap/version logic so the active/inactive decision is consistent across validation and storage.
- **Migration `20260602000000`** — drops the `*_session_key_sig_present_chk` CHECK constraints added in `20260508000000` (they rejected an empty `session_key_sig`). The ownership guarantee is preserved in application code: active rows are still co-signed, and owner/auth lookups filter `expires_at > now`, so revocation rows with an empty signature are never trusted as a key's owner. Down re-adds the constraints as `NOT VALID`.
- **SDK** — wallet-only helpers: `RevokeChannelSessionKey` / `RevokeAppSessionKey` (Go), `revokeChannelSessionKey` / `revokeSessionKey` (TS). Updated the docstrings that previously stated wallet-only revocation was unsupported.
- **`docs/api.yaml`** — `session_key_sig` documented as optional on the revocation path for both methods.

### Tests

- pkg-level unit tests for the new user-sig-only validators (valid passes, wrong wallet rejected, tampered version rejected).
- Handler tests per kind: wallet-only revoke succeeds; a mismatched `session_key_sig` is ignored on revoke; an invalid `user_sig` is still rejected on revoke. Existing both-signature, cap, and reactivation tests unchanged and passing.

### Verification

- `go build ./...`, `go vet` (affected packages), Go tests for pkg/app, pkg/core, both handlers, sdk/go — pass.
- TS typecheck + lint clean; 184 tests pass (public-API drift snapshot updated for the two new methods — intentional).

### Note for reviewers

Migration is timestamped `20260602000000`. If it should slot before another unreleased migration, say so and I'll rename.

## Release Notes

* **New Features**
  * Added wallet-only session-key revocation support. New methods `RevokeChannelSessionKey`, `RevokeAppSessionKey` (Go SDK) and `revokeChannelSessionKey`, `revokeSessionKey` (TypeScript SDK) enable simplified revocation without requiring session-key holder signatures.

* **Documentation**
  * Clarified signature requirements for session-key operations: activation/extension requires both wallet and session-key signatures; revocation requires wallet signature only.


nitronode/api/channel_v1/submit_session_key_state.go
```diff
@@ -75,19 +75,34 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
...
+		if coreState.SessionKeySig == "" {
+			c.Fail(rpc.Errorf("invalid_session_key_state: session_key_sig is required"), "")
+			return
+		}
+		// Validate both signatures: wallet's user_sig and session-key holder's session_key_sig.
+		if err := core.ValidateChannelSessionKeyStateV1(coreState); err != nil {
+			c.Fail(rpc.Errorf("invalid_session_key_state: %v", err), "")
+			return
+		}
+	} else {
```
ihsraham, line 97: Good direction on splitting active vs revoke validation, but I would block on one ownership edge case here. A wallet can submit `version=1` with `expires_at <= now` for any arbitrary `session_key`; this branch only checks `user_sig`, then `LockSessionKeyState` creates the permanent `(session_key, kind)` owner row and the later `latestVersion + 1` check accepts it.

Could we only allow the user-only path after the lock proves `latestVersion > 0` for this wallet? For first submit, reject the expired state or require both signatures before the pointer row can be created.

nitronode/api/app_session_v1/submit_session_key_state.go
```diff
@@ -92,19 +92,34 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
...
+		if coreState.SessionKeySig == "" {
+			c.Fail(rpc.Errorf("invalid_session_key_state: session_key_sig is required"), "")
+			return
+		}
+		// Validate both signatures: wallet's UserSig and session-key holder's SessionKeySig.
+		if err := app.ValidateAppSessionKeyStateV1(coreState); err != nil {
+			c.Fail(rpc.Errorf("invalid_session_key_state: %v", err), "")
+			return
+		}
+	} else {
```
ihsraham, line 114: Same issue applies here for app session keys. A first submit with `expires_at <= now` can reserve an unowned `session_key` under the caller because the user-only branch runs before the lock proves an existing delegation.

Please gate wallet-only revocation on `latestVersion > 0`. First submits should either be future-dated with both signatures or rejected.

ihsraham (changes requested):
Good direction overall: future-dated registration and re-activation still require both signatures, and the new user-only revoke path addresses the lost or uncooperative session-key case.

I would keep this open for now. The remaining blocker is that an expired first submit can reserve an unowned `session_key` before any delegate signature is required. I left inline notes on both handler paths.

Also worth picking up the existing notes on clearing `SessionKeySig` before storing revocation rows, marking `session_key_sig` optional in the API schema, and aligning the Go helper time check to Unix seconds.

sdk/ts/src/client.ts
```diff
@@ -1776,6 +1777,32 @@ export class Client {
...
+   * session-key holder cannot or will not co-sign — a lost, unavailable, or compromised
+   * delegate. The supplied state must carry the next monotonic version (latest + 1) and an
+   * expires_at at or before now; this method signs it with the wallet (user_sig) and submits
+   * with an empty session_key_sig. The server accepts user-only signatures only on the
+   * revocation path (expires_at <= now). For registration, rotation, or extension use
+   * submitChannelSessionKeyState with both signatures.
+   *
+   * @param state - The channel session key state to revoke (version = latest + 1, expires_at <= now)
+   */
+  async revokeChannelSessionKey(state: ChannelSessionKeyStateV1): Promise<void> {
```
ihsraham: This helper signs and clears the signature fields itself, but the input type still requires callers to provide `user_sig` and `session_key_sig`. That means TS callers need dummy values for the wallet-only revoke path.

Could we use a dedicated revoke input type, or `Omit<ChannelSessionKeyStateV1, 'user_sig' | 'session_key_sig'>`, and apply the same shape to `revokeSessionKey`?

sdk/go/app_session.go
```diff
@@ -446,3 +447,44 @@ func SignAppSessionKeyOwnership(state app.AppSessionKeyStateV1, sessionKeySigner
...
+//	    UserAddress:    client.GetUserAddress(),
+//	    SessionKey:     lostSessionKey,
+//	    Version:        latestVersion + 1,
+//	    ApplicationIDs: []string{},
+//	    AppSessionIDs:  []string{},
+//	    ExpiresAt:      time.Now(),
+//	}
+//	err := client.RevokeAppSessionKey(ctx, state)
+func (c *Client) RevokeAppSessionKey(ctx context.Context, state app.AppSessionKeyStateV1) error {
+	if state.ExpiresAt.After(time.Now()) {
```
ihsraham: Same precision issue applies to the app helper. This guard uses full `time.Time` precision, but `PackAppSessionKeyStateV1` signs `ExpiresAt.Unix()`, so a value inside the current Unix second can be rejected locally even though the server treats it as `expires_at <= now`.

Please compare Unix seconds here too, matching the channel helper fix.

nitronode/api/channel_v1/submit_session_key_state.go
philanton: Yeah, good catch — this was the real hole. Fixed in [bf08526](https://github.com/layer-3/nitrolite/pull/809/changes/bf0852691587f4d570fc5ad07bb92dc1d66e0a22).

I went with rejecting a revoke at `version == 1` before `LockSessionKeyState` runs, rather than gating on `latestVersion > 0` after the lock. The reason: the seed row is written inside the same tx, so a `version >= 2` revoke on an unowned key would seed, then fail the `latestVersion + 1` check and roll the seed back with the tx. The only revoke that can actually commit a claim for a key with no prior delegation is `version == 1` (passes `0 + 1`), so blocking that is enough — and it never even seeds. A legitimate revoke is always `version >= 2` since registration at v1 is future-dated and requires both sigs.

nitronode/api/app_session_v1/submit_session_key_state.go
philanton: Same fix here — [bf08526](https://github.com/layer-3/nitrolite/pull/809/changes/bf0852691587f4d570fc5ad07bb92dc1d66e0a22). Reject `version == 1` revoke before the lock, same reasoning as the channel handler (the seed only persists if the tx commits, and only a v1 revoke commits without a prior delegation).

sdk/ts/src/client.ts
philanton: Yep, makes sense — switched both `revokeChannelSessionKey` and `revokeSessionKey` to `Omit<…, 'user_sig' | 'session_key_sig'>` so callers don't pass dummy sigs. [bf08526](https://github.com/layer-3/nitrolite/pull/809/changes/bf0852691587f4d570fc5ad07bb92dc1d66e0a22)

sdk/go/app_session.go
philanton: Fixed in [bf08526](https://github.com/layer-3/nitrolite/pull/809/changes/bf0852691587f4d570fc5ad07bb92dc1d66e0a22) — comparing `ExpiresAt.Unix()` here too, matching the channel helper.

docs/api.yaml
```diff
@@ -661,15 +664,16 @@ api:
...
+                `user_sig` (wallet) and `session_key_sig` (session-key holder, proving possession).
+                Revocation (`expires_at` at or before now) requires only `user_sig`: a wallet may
+                revoke unilaterally so a lost, unavailable, or compromised session key cannot block
+                revocation of its own delegation. Any `session_key_sig` supplied on the revocation
+                path is ignored.
               request:
                 - field_name: state
                   type: channel_session_key_state
-                  description: Session key metadata and delegation information. Set `expires_at` to a past value to revoke; future to (re-)activate.
+                  description: Session key metadata and delegation information. Set `expires_at` to a past value to revoke (user_sig only); future to (re-)activate (both signatures).
```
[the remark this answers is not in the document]
philanton: Done in [bf08526](https://github.com/layer-3/nitrolite/pull/809/changes/bf0852691587f4d570fc5ad07bb92dc1d66e0a22) — updated both field descriptions to "at or before now".

nitronode/api/app_session_v1/submit_session_key_state.go
```diff
@@ -92,19 +92,34 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
...
+		// Revocation: validate only the wallet's UserSig. Any session_key_sig present is ignored.
+		if err := app.ValidateAppSessionKeyStateUserSigV1(coreState); err != nil {
+			c.Fail(rpc.Errorf("invalid_session_key_state: %v", err), "")
+			return
+		}
 	}
```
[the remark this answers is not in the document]
philanton: Fixed in [bf08526](https://github.com/layer-3/nitrolite/pull/809/changes/bf0852691587f4d570fc5ad07bb92dc1d66e0a22) — clearing `SessionKeySig` on the revoke branch before persisting, with a regression assertion in the revoke test.

nitronode/api/channel_v1/submit_session_key_state.go
```diff
@@ -75,19 +75,34 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
...
+		// Revocation: validate only the wallet's user_sig. Any session_key_sig present is ignored.
+		if err := core.ValidateChannelSessionKeyStateUserSigV1(coreState); err != nil {
+			c.Fail(rpc.Errorf("invalid_session_key_state: %v", err), "")
+			return
+		}
 	}
```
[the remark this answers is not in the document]
philanton: Fixed in [bf08526](https://github.com/layer-3/nitrolite/pull/809/changes/bf0852691587f4d570fc5ad07bb92dc1d66e0a22) — cleared `SessionKeySig` on the revoke branch before persisting, plus a regression assertion.

sdk/go/channel.go
```diff
@@ -1071,6 +1072,46 @@ func SignChannelSessionKeyOwnership(state core.ChannelSessionKeyStateV1, session
...
+//	    SessionKey:  lostSessionKey,
+//	    Version:     latestVersion + 1,
+//	    Assets:      []string{},
+//	    ExpiresAt:   time.Now(),
+//	}
+//	err := client.RevokeChannelSessionKey(ctx, state)
+func (c *Client) RevokeChannelSessionKey(ctx context.Context, state core.ChannelSessionKeyStateV1) error {
+	if state.ExpiresAt.After(time.Now()) {
+		return fmt.Errorf("revocation requires expires_at at or before now, got %s", state.ExpiresAt)
+	}
```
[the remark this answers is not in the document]
philanton: Fixed in [bf08526](https://github.com/layer-3/nitrolite/pull/809/changes/bf0852691587f4d570fc5ad07bb92dc1d66e0a22) — comparing Unix seconds now, same as the app helper.

ihsraham (approved):
Approving for M-05. The latest commit closes the revoke-veto path and the first-submit ownership issue: v1 revokes are rejected before the pointer lock, v2+ first-submit revokes fail the version check in the transaction, and revoke rows are canonicalized by clearing SessionKeySig.

Only non-blocking follow-up I would consider is a DB-backed regression for the v2+ first-submit rollback case, since the current tests already cover the v1 pre-lock rejection.

nksazonov:
After reading this PR, I have some thoughts:

> Design question: who should be allowed to revoke a session key?

### What the PR does now

Revocation (`expires_at <= now`) requires **only `user_sig`** (wallet owner). Any `session_key_sig` present is silently ignored and cleared before storage.

---

### Industry analogues

| Domain | Who can revoke? | Notes |
|--------|-----------------|-------|
| **OAuth 2.0 RFC 7009** | **Owner OR token holder** | Token surrender is an explicit, recommended pattern — client calls `/revoke` authenticated with its own token. |
| **GitHub PATs** | **Owner OR key itself** | `DELETE /applications/{client_id}/grant` works with the token as auth. |
| **JWT** | Owner only (server-side) | JWTs can't self-revoke — but they're stateless. Not analogous since we have a DB. |
| **SSH authorized_keys** | Owner only | Key can't deauthorize itself, but SSH keys are passive and don't run code. |
| **Power of Attorney (banking)** | **Principal OR attorney** | Attorney can resign ("surrender") independently of the principal revoking. |

The closest analogue is **OAuth RFC 7009**. A session key is a delegated access token that runs code — exactly what RFC 7009 covers.

---

### The core argument for EITHER

A session key self-revoking is **strictly a reduction of authority**. It can never escalate privilege. This makes it unconditionally safe.

Practical use cases the current design blocks:

- **Autonomous agent graceful shutdown** — a bot that finishes a job wants to clean up its own delegation without requiring the wallet to come online.
- **Compromised key first-response** — a leaked session key can self-revoke before the attacker uses it (the attacker can't *delay* revocation by refusing to sign).
- **SDK ergonomics** — `RevokeSessionKey()` that works from the session key side, not just the wallet side. Much more natural for server-to-server integrations.

---

### Security consideration on session-key-initiated revocation

Could an attacker who steals a session key cause harm by self-revoking it?

- Self-revocation only removes authority — the attacker gains nothing.
- Worst case: the user has to submit `version + 1` to re-activate (a minor inconvenience).
- The signed payload already binds `(user_address, session_key, version, expires_at)` — no replay across keys, wallets, or versions.

The one non-obvious risk is a **racing revocation**: an attacker steals the key and immediately self-revokes it before the owner notices. But this is a minor DoS at most — the owner can re-register at `version + 1`. And if the key is already stolen, the user has bigger problems regardless.

---

### My personal recommendation: accept EITHER `user_sig` OR valid `session_key_sig` on revocation

Logic for the revocation branch:
1. If `session_key_sig` is present and valid → accept (session key surrenders itself).
2. Else if `user_sig` is valid → accept (owner forcibly revokes).
3. Else → reject.

`user_sig` still remains the **escape hatch** for lost/unavailable/malicious delegate scenarios — the original motivation for this PR. Session-key self-revocation is purely additive and matches OAuth 2.0 RFC 7009, GitHub PAT, and Power-of-Attorney patterns.

---

Note: the current code already silently clears an invalid `session_key_sig` on the revocation path (line 113 in the handler), so the only gap is that a **session-key-only revocation** (no `user_sig`) is not supported at all. The change to support it would be small and additive.

philanton:
Yeah, this is a really solid writeup — the RFC 7009 framing makes sense, and I agree self-revocation is "strictly a reduction of authority," so it's safe in principle.

That said, I'd keep it out of this PR. MF3-M05 is specifically "the wallet can't unilaterally revoke" — and wallet-only revoke closes exactly that. Session-key self-revocation is a separate, additive capability (the _delegate_ side rather than the _owner_ side), so I'd rather not widen the scope of the audit fix here.

nksazonov:
Fair point — scope discipline on an audit fix makes sense. Tracked the session-key self-revocation idea in #823 so it doesn't get lost.

nksazonov (approved):
LGTM

PR #811, opened 2026-06-03
MF3-L04: fix(nitronode): enforce canonical allocations in deposit state

philanton:
## What

Remediation for **MF3-L04** — `SubmitDepositState()` only partially validated `AppStateUpdate.Allocations`, so an accepted deposit update could carry a non-canonical or incomplete allocation set relative to the node's actual ledger.

Two gaps:
1) the participant check sat *inside* the amount-increase branch, so a non-participant could be included with a zero amount and bypass it;
2) the final completeness check skipped `asset == userState.Asset`, so existing nonzero allocations for the deposited asset could be omitted from the signed update.

Neither moves or erases funds (omitted balances aren't modified, zero allocations produce no ledger entries) — but the signed deposit state could diverge from the ledger, which is misleading for any client/app/verifier treating it as a full allocation snapshot.

## Changes

- Hoist the participant check to the top of the allocation loop so it runs regardless of amount — non-participant allocations are now rejected even at zero amount.
- Drop the deposited-asset skip in the completeness check. Every existing nonzero allocation must now be present in the update:
  - **deposited asset** — may only grow (the deposit itself; the increase is still validated and recorded in the forward loop), never shrink;
  - **all other assets** — must match the current state exactly (unchanged).

Valid deposit increases keep working; the signed allocation set becomes a complete, canonical snapshot.

## Tests

- `NonParticipantZeroAllocation_Rejected` — zero-amount non-participant allocation is rejected, no ledger entry recorded for it.
- `MissingDepositedAssetAllocation_Rejected` — omitting another participant's existing balance in the deposited asset is rejected, session is not updated.

All `app_session_v1` tests pass; `go vet` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved participant validation during deposit state submission to reject invalid allocations earlier in processing
  * Strengthened balance verification to ensure all existing asset allocations are accounted for in updates
  * Enforced stricter requirements for deposited asset allocations to maintain non-decreasing amounts

### DIFF
--- a/nitronode/api/app_session_v1/submit_deposit_state.go
+++ b/nitronode/api/app_session_v1/submit_deposit_state.go
@@ -185,6 +185,13 @@ func (h *Handler) SubmitDepositState(c *rpc.Context) {
 
 		incomingAllocations := make(map[string]map[string]decimal.Decimal)
 		for _, alloc := range appStateUpd.Allocations {
+			// Validate participant before any amount-diff logic so non-participant
+			// allocations are rejected even when the amount is zero (which would
+			// otherwise bypass the check below).
+			if _, ok := participantWeights[alloc.Participant]; !ok {
+				return rpc.Errorf("allocation to non-participant %s", alloc.Participant)
+			}
+
 			if alloc.Amount.IsNegative() {
 				return rpc.Errorf("negative allocation: %s for asset %s", alloc.Amount, alloc.Asset)
 			}
@@ -216,11 +223,6 @@ func (h *Handler) SubmitDepositState(c *rpc.Context) {
 			}
 
 			if alloc.Amount.GreaterThan(currentAmount) {
-				// Validate participant
-				if _, ok := participantWeights[alloc.Participant]; !ok {
-					return rpc.Errorf("allocation to non-participant %s", alloc.Participant)
-				}
-
 				// Validate that allocation asset matches user state asset
 				if alloc.Asset != userState.Asset {
 					return rpc.Errorf("app session deposit allocation for asset '%s' does not match user channel state asset '%s'", alloc.Asset, userState.Asset)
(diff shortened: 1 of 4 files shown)


ihsraham (approved):
approving for the L-04 fix. the important parts are covered: non-participant allocations are rejected before amount-diff handling, and the completeness check now includes the deposited asset, so existing nonzero balances cannot disappear from the signed snapshot. the added regression tests also cover both original bypasses.

one non-blocking cleanup note: this still accepts surplus zero allocations for valid participants/assets with no current balance. that does not look like ledger or state impact, but if we want a strict canonical allocation list later, it is worth tightening. i also saw only existing scanner noise from `gosec` / `govulncheck`, with no findings in the changed files and no manifest changes.

nitronode/api/app_session_v1/submit_deposit_state.go
```diff
@@ -216,11 +223,6 @@ func (h *Handler) SubmitDepositState(c *rpc.Context) {
 			}
 
 			if alloc.Amount.GreaterThan(currentAmount) {
```
nksazonov, line 232: **Informational (not in MF3-L04 scope):** When `alloc.Amount.IsZero()` and the participant has no existing balance for that asset (`currentAmount == 0`), neither the `LessThan` nor the `GreaterThan` branch executes — the zero allocation is stored in `incomingAllocations` without any validation. A valid participant can include a spurious zero allocation for a brand-new `(participant, asset)` pair and it passes through silently. No funds move, but the signed allocation set isn't perfectly canonical.

This is the same class of canonicality issue as MF3-L04 but scoped to valid participants. Suggest rejecting zero allocations where no existing balance exists — e.g. after the `IsNegative` guard, add a check for `alloc.Amount.IsZero() && currentAmount.IsZero()`.

nksazonov (approved):
Excellent work. Both MF3-L04 gaps are correctly closed, and the two regression tests (`NonParticipantZeroAllocation_Rejected` and `MissingDepositedAssetAllocation_Rejected`) cover exactly the described scenarios.

PR #812, opened 2026-06-03
MF3-L10: reject missing last user state in submit_deposit_state

philanton:
## Summary

`SubmitDepositState()` synthesized `core.NewVoidState()` when `GetLastUserState()` returned `nil` and continued validation. But deposit is **not** a channel initialization path: after `CheckActiveChannel()` confirms an active `Void`/`Open` home channel, the `channel_states` row for `(user_wallet, asset)` is guaranteed to exist (home channels are created atomically with their first state in `request_creation.go`).

A `nil` result there means the local-state materialization is inconsistent. Continuing from a synthetic void state evaluates a `Commit` against state history that was never persisted, yielding a misleading downstream validation error instead of a clean missing-state rejection.

## Changes

- Replace the `NewVoidState()` fallback with an explicit `last user state not found` rejection.
- `EnsureNoOngoingStateTransitions` now runs unconditionally — `currentState` is guaranteed non-nil past the guard.
- Add `TestSubmitDepositState_MissingLastUserState_Rejected`: active channel passes but `GetLastUserState` returns nil → clean rejection, no advancement past the guard.

## Impact

Hardening / fail-fast. The void path was unreachable under correct operation (no exploit, no incorrect accept) — worst prior consequence was a confusing error on DB inconsistency.

## Testing

- `go test ./nitronode/api/app_session_v1/` — pass
- `go vet ./nitronode/api/app_session_v1/` — clean

### DIFF
--- a/nitronode/api/app_session_v1/submit_deposit_state.go
+++ b/nitronode/api/app_session_v1/submit_deposit_state.go
@@ -122,16 +122,19 @@ func (h *Handler) SubmitDepositState(c *rpc.Context) {
 			return rpc.Errorf("missing user signature on user state")
 		}
 
+		// An active home channel (confirmed by CheckActiveChannel above) always has a
+		// persisted latest state; deposit is not an initialization path. A nil result here
+		// means the local state materialization is inconsistent, so reject instead of
+		// synthesizing a void state and validating against history that was never persisted.
 		currentState, err := tx.GetLastUserState(userState.UserWallet, userState.Asset, false)
 		if err != nil {
 			return rpc.Errorf("failed to get last user state: %v", err)
 		}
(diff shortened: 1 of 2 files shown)


ihsraham (approved):
approved for L-10. the fix now rejects a missing latest user state after the active-channel check instead of synthesizing a void state, so the deposit path no longer validates against state that was never persisted.

the regression test covers the inconsistent active-channel/no-last-state case and confirms the handler stops before transition checks or storage. no blockers from my pass; the remaining scanner notes I saw were baseline/toolchain noise, not introduced by this PR.

dimast-x (approved):

nksazonov (approved):
Well done!

PR #813, opened 2026-06-03
MF3-L11: fix unbounded responseSinks growth on marshal failure

philanton:
## Finding (MF3-L11)

`WebsocketDialer.Call` registered the response sink **before** marshalling the request. If marshalling failed (malformed raw JSON in the payload), the function returned `ErrMarshalingRequest` without removing the sink. Repeated direct calls to the low-level `WebsocketDialer.Call` API with malformed payloads and unique request IDs could grow `responseSinks` unboundedly, degrading or crashing the calling process.

SDK methods are not affected — they build payloads via `NewPayload` before reaching `Call`.

## Fix

Move `json.Marshal` ahead of the connection check and sink registration in `Call`. A marshal failure now returns before any sink is registered, so nothing can leak.

## Test

`pkg/rpc/dialer_internal_test.go` — white-box test (no server needed, since marshalling now precedes the connection check):
- 100 calls with malformed `json.RawMessage` payloads and unique request IDs
- asserts each returns `ErrMarshalingRequest`
- asserts `responseSinks` is empty afterward

The test locks in the reordering: with the old order a disconnected dialer returned `ErrNotConnected` before reaching the marshal, so it would fail.

### DIFF
--- a/pkg/rpc/dialer.go
+++ b/pkg/rpc/dialer.go
@@ -284,6 +284,13 @@ func (d *WebsocketDialer) Call(ctx context.Context, req *Message) (*Message, err
 		return nil, ErrNilRequest
 	}
 
+	// Marshal the request before registering a response sink so a marshal
+	// failure cannot leak entries in responseSinks.
+	reqJSON, err := json.Marshal(req)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrMarshalingRequest, err)
+	}
+
 	// Check connection and register response channel atomically
 	d.mu.Lock()
 	if d.dialCtx == nil || d.dialCtx.ctx.Err() != nil {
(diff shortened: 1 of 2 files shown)


ihsraham (approved):
approving for L-11. The important fix is in the right place: `Call` marshals the request before checking the connection and before registering `responseSinks[req.RequestID]`, so malformed `json.RawMessage` returns before the pending sink map is touched. The regression test also covers repeated malformed calls and checks the sink map stays empty.

non-blocking: the test uses an unconnected dialer, so it proves the new ordering and no-leak invariant but does not recreate the original connected-state leak shape directly. I would keep that as optional follow-up because the production order closes the leak.

the duplicate `RequestID` and shutdown sink race concerns are existing dialer hardening items, not blockers for this finding. local tests and race checks passed, and the scanner output I saw was existing/toolchain noise rather than PR-introduced risk.

nksazonov (approved):
Excellent work! The fix is correct, minimal, and the test enforces the ordering rather than just observing the symptom.

PR #814, opened 2026-06-03
fix(nitronode): remediate MF3-I* audit findings

philanton:
Remediates the MF3-I* batch of audit findings (informational severity). Each finding is its own commit so they can be reviewed independently.

## Findings

**MF3-I01 — `MAX_SESSION_KEYS_PER_USER=0` silently disables the per-user cap**
The cap is skipped whenever `maxSessionKeysPerUser <= 0`, but that wasn't documented and `0` could be read as "invalid" or "zero keys allowed". Documented `<= 0` as the unlimited mode on the config field and in the app-session submit README. Kept it as a documented behavior rather than a startup-reject, since that matches the existing `WsBytesPerSec` "<0 to disable" / `MaxSignedUpdates` `0` conventions and avoids breaking anyone already running `0`.

**MF3-I02 — duplicate asset symbols accepted case-insensitively**
`verifyAssetsConfig()` accepted both `usdt` and `USDT`, which is ambiguous because the in-memory registry matches symbols case-sensitively while the persistence layer lowercases them. Now tracks enabled symbols by their lowercased form and fails startup on a case-insensitive (or exact) duplicate. Disabled assets are skipped; request-time exact-symbol semantics are unchanged.

**MF3-I03 — stale `err` in missing-user-signature error**
The missing-`user_sig` guard formatted an `err` that is always `nil` at that point, polluting the message with `<nil>`. Dropped the format arg.

**MF3-I04 — nil-deref in `NewTransactionFromTransition` release branch**
The `release` case dereferenced `receiverState.UserWallet` before the nil check, so a nil receiver state panicked instead of returning an error. Moved the nil check ahead of any field access. No RPC path passes nil today, but the helper is now safe regardless of caller. Added a regression test.

**MF3-I05 — pagination offset can wrap negative on 32-bit**
`GetOffsetAndLimit` returned a caller-controlled `uint32` offset that every store query converts via `int(offset)` before GORM's `Offset()`; on a 32-bit target a large offset wraps negative and GORM treats it as "no offset", silently returning the first page. Clamped offset to `MaxInt32` in `GetOffsetAndLimit`, then routed the handlers that still computed offset/limit by hand through the same helper so the clamp is enforced everywhere (`channels.v1.get_channels`, `channels.v1.get_last_key_states`, `app_sessions.v1.get_last_key_states`). An explicit `limit: 0` still falls back to the endpoint default so page-count math never divides by zero.

**MF3-I07 — session-key revocation mislogged**
Both `SubmitSessionKeyState` handlers logged a revocation whenever the submitted `expires_at` was not in the future, ignoring whether the previous state was active — so inactive-to-inactive updates (and already-expired new versions) were logged as revocations. Now gated on the real transition: `revoked := prevActive && !submittedActive`.

**MF3-I08 — unvalidated channel / app-session IDs**
`GetAppSessions` accepted a non-nil but empty `app_session_id`, which the store treated as "no filter", silently returning an unfiltered paginated list instead of honoring the documented either-id-or-participant contract. Added `core.HashRegex` + `core.IsValidHash` (a `0x`-prefixed 32-byte hash — the canonical form of channel and app-session IDs), mirroring `app.IsValidApplicationID`, and reject malformed IDs at the RPC boundary in `GetAppSessions` and `GetEscrowChannel`. `HomeChannelID` is left as-is — it has no bare-id read endpoint and is already verified by exact derivation match or existence lookup.

**MF3-I09 — `UpdateChannel` rewrites immutable config**
`UpdateChannel` rewrote immutable config (`blockchain_id`, `token`, `nonce`) on every lifecycle update, widening the blast radius if a future caller passes a partial/stale `core.Channel`. Restricted it to mutable lifecycle fields (`status`, `state_version`, `challenge_expires_at`, `updated_at`); immutable config is set once in `CreateChannel`.

## Test plan

- `go build ./...`
- `go test ./pkg/core/... ./nitronode/store/database/... ./nitronode/api/channel_v1/... ./nitronode/api/app_session_v1/... ./nitronode/event_handlers/...`
- `go vet ./...` on touched packages

## Summary by CodeRabbit

* **Bug Fixes**
  * Session-key revocation logging now only fires when a key actually transitions from active to inactive
  * Error message for missing state signature corrected
  * Prevented a potential crash when handling a specific transaction case

* **New Features**
  * Validation added for session and channel identifiers to reject malformed IDs

* **Improvements**
  * Centralized, more consistent pagination and safer offset handling across endpoints
  * Asset symbol validation now treats duplicates as case-insensitive
  * Channel updates no longer modify immutable configuration fields

* **Documentation**
  * Clarified per-user active session-key cap semantics and disabling behavior

### DIFF
--- a/pkg/core/types.go
+++ b/pkg/core/types.go
@@ -879,12 +879,12 @@ func NewTransactionFromTransition(senderState *State, receiverState *State, tran
 		toAccount = transition.AccountID
 
 	case TransitionTypeRelease:
-		txType = TransactionTypeRelease
-		fromAccount = transition.AccountID
-		toAccount = receiverState.UserWallet
 		if receiverState == nil {
 			return nil, fmt.Errorf("receiver state must not be nil for 'release' transition")
 		}
+		txType = TransactionTypeRelease
+		fromAccount = transition.AccountID
+		toAccount = receiverState.UserWallet
 
 	case TransitionTypeMutualLock:
 		if senderState.EscrowChannelID == nil || senderState.HomeChannelID == nil {
--- a/pkg/core/types.go
+++ b/pkg/core/types.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 	"time"
@@ -1226,6 +1227,12 @@ func (p *PaginationParams) GetOffsetAndLimit(defaultLimit, maxLimit uint32) (off
 			limit = min(*p.Limit, maxLimit)
 		}
 	}
+
+	// Callers convert offset to int before handing it to GORM's Offset(). On a
+	// 32-bit target int(offset) wraps to a negative value for large uint32s,
+	// which GORM treats as "no offset" and silently returns the first page.
+	// Clamp to MaxInt32 so the conversion is always a non-negative int.
+	offset = min(offset, math.MaxInt32)
 
 	return offset, limit
 }
--- a/pkg/core/types.go
+++ b/pkg/core/types.go
@@ -1202,8 +1202,8 @@ type EscrowWithdrawalDataResponse struct {
 
 // BalanceEntry represents a balance entry for an asset
 type BalanceEntry struct {
-	Asset   string          `json:"asset"`   // Asset symbol
-	Balance decimal.Decimal `json:"balance"` // Balance amount
+	Asset    string          `json:"asset"`    // Asset symbol
+	Balance  decimal.Decimal `json:"balance"`  // Balance amount
 	Enforced decimal.Decimal `json:"enforced"` // On-chain enforced balance
 }
 
--- a/nitronode/api/app_session_v1/README.md
+++ b/nitronode/api/app_session_v1/README.md
@@ -715,7 +715,7 @@ ORDER BY created_at DESC;
 - `app_session_ids` entries must be lowercase strings (non-lowercase values are rejected before signature verification)
 - Version must be sequential (latest_version + 1)
 - Signature must recover to `user_address`
-- The per-user cap (`NITRONODE_MAX_SESSION_KEYS_PER_USER`, default 100) is enforced whenever the submit transitions the slot from inactive to active: a brand-new key (no prior state) or a reactivation (previous latest state's `expires_at` was already in the past). Rotation/update against a still-active key, and revocation submits, are not subject to the cap.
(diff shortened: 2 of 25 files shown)


ihsraham (approved):
approving for the MF3-I batch. the fixes cover the audit asks: the session-key cap behavior is documented, asset symbols are checked case-insensitively at config load, the stale signature error is cleaned up, the release transaction nil check is before the dereference, ID filters are rejected before store reads, revocation logging now keys off an active-to-inactive transition, and `UpdateChannel` no longer rewrites immutable config fields.

one non-blocking follow-up: the I-05 offset clamp is enforced at the current RPC callers before they reach the store, which closes the runtime path I reviewed. the lower-level store methods still accept raw `uint32 offset` values and cast them with `int(offset)`, so if we want the store API itself to carry the invariant, it would be worth adding the same clamp there too.

also worth running `gofmt` on `nitronode/runtime.go`, `nitronode/store/database/channel_test.go`, `pkg/core/types.go`, and `pkg/core/utils_test.go` before merge. my local checks otherwise passed, and the scanner output I saw was existing/toolchain noise rather than PR-introduced risk.

pkg/core/types.go
```diff
@@ -1227,6 +1228,12 @@ func (p *PaginationParams) GetOffsetAndLimit(defaultLimit, maxLimit uint32) (off
 		}
 	}
 
+	// Callers convert offset to int before handing it to GORM's Offset(). On a
+	// 32-bit target int(offset) wraps to a negative value for large uint32s,
+	// which GORM treats as "no offset" and silently returns the first page.
+	// Clamp to MaxInt32 so the conversion is always a non-negative int.
+	offset = min(offset, math.MaxInt32)
```
nksazonov, line 1248: The clamp enforces the invariant at the RPC handler layer. However, six store methods still do `int(offset)` directly without a guard: `app.go:79`, `app_session.go:119`, `channel.go:192`, `transaction.go:109`, `app_session_key_state.go:142`, `channel_session_key_state.go:123`. Any caller that reaches the store without routing through `GetOffsetAndLimit` — admin tooling, future internal endpoints, tests passing raw values — can still trigger the silent first-page bug. Consider either adding the clamp in the store query helpers, or adding a doc comment on each store method requiring callers to pre-clamp.

nksazonov (approved):
Good work — all eight addressed findings are correctly implemented and backed by targeted regression tests.

PR #818, opened 2026-06-05
MF3-L13: document asset-symbol equivalence operator invariant

philanton:
## Audit finding MF3-L13

The state validator lets off-chain credit with no `HomeChannelID` bind to any chain/token the user proposes at first channel creation, enforcing only that the **asset symbol** matches (`pkg/core/state_advancer.go:34-55`). Consequently credit can be redeemed from any configured token inventory sharing that symbol, independent of which inventory originally backed it.

## Assessment

This is **intended behaviour** of the unified asset model — a unified asset is identified solely by its symbol, and all chain-specific tokens configured under that symbol are fungible 1:1 representations. Leaving the chain/token open until the user binds it is exactly what enables cross-chain redemption. Binding credit to its origin token would remove that feature.

The harmful scenario requires the **node operator** to map economically non-equivalent tokens (e.g. a test token and production USDC) under a single symbol. In that case the operator drains their own inventory through their own misconfiguration — no external trust boundary is crossed. Token equivalence cannot be verified programmatically.

Severity: informational / config hardening. No validation or provenance change.

## Change

Make the operator obligation explicit:
- `docs/protocol/security-and-limitations.md` — new **Asset-symbol equivalence** trust assumption.
- `nitronode/store/memory/asset_config.go` — equivalence warning on the `AssetConfig` doc comment, cross-referencing the protocol doc.


docs/protocol/security-and-limitations.md
```diff
@@ -64,6 +64,7 @@ In the current protocol version, participants MUST trust nodes for:
...
+- **Asset-symbol equivalence** — the node operator controls which chain-specific tokens are configured under a single unified asset symbol. The protocol treats all tokens sharing a symbol as fully fungible 1:1 representations of the same asset, so off-chain credit denominated in that asset can be redeemed from any of those token inventories regardless of which one originally backed it (the validator binds unchanneled credit to the chain/token chosen at first channel creation, enforcing only that the asset symbol matches). This is intended behaviour that enables cross-chain redemption. Operators MUST therefore configure only economically equivalent (1:1 redeemable) tokens under one symbol; grouping non-equivalent tokens (e.g. a test token and production USDC) under the same symbol would let credit sourced from the cheap inventory be redeemed against the valuable one. Token equivalence cannot be verified programmatically and is an operator configuration responsibility.
```
ihsraham, line 67: Good call documenting the unified-asset assumption here. This explains why binding detached credit to an origin token would conflict with intended cross-chain redemption.

Missing piece for L-13 closure: the same warning is not present where operators actually configure this risk, such as the `nitronode/README.md` `config/assets.yaml` example and the Helm `assets.yaml` path. Since the mitigation is docs-only, could we also put the 1:1 equivalence requirement on those asset-config surfaces, or explicitly mark the audit item as accepted/documented operator risk?

ihsraham (changes requested):
Directionally this is the right framing if L-13 is being handled as an accepted operator-configuration risk. I left one closure note: since the mitigation is documentation-only, the invariant should also be visible in the asset configuration path before we call this closed.

docs/protocol/security-and-limitations.md
philanton: Yeah, good catch — the warning was only in the protocol doc and the `AssetConfig` godoc, neither of which an operator editing YAML actually sees.

Added the 1:1 equivalence warning to the config surfaces in [17af52e](https://github.com/layer-3/nitrolite/pull/818/changes/17af52e415d81a6e3fab08eb8dd381c005fdcdb0): the `config/assets.yaml` example in `nitronode/README.md` and the Helm prod/stress/sandbox `assets.yaml` files, each linking back to security-and-limitations. So L-13 stays a documented operator-config risk, just now visible at the point of configuration.

ihsraham (approved):
Approving for L-13. Treating this as a documented operator-config risk, the warning is now in the protocol security doc, the `config/assets.yaml` README example, and the Helm asset config surfaces, so operators see the 1:1 equivalence requirement at configuration time.

No runtime enforcement was added, but that matches the docs-only mitigation path here. No blockers from me.

nksazonov (changes requested):
Please, add the same information in more detailed and shortened version to `contracts/SECURITY.md` and `protocol-description.md` respectively.

philanton:
@nksazonov done in [81c134e](https://github.com/layer-3/nitrolite/pull/818/changes/81c134ef). Added the detailed version as an `Asset-symbol equivalence` trust assumption in `contracts/SECURITY.md`, and a shortened bullet in the security model summary of `protocol-description.md`. Both link back / align with the full statement in `security-and-limitations.md`.

nksazonov (approved):
Well done!

PR #819, opened 2026-06-05
MF3-L14: enforce request-rate limit at the frame layer

philanton:
## Audit finding (MF3-L14)

The per-connection request-rate limiter (~10 rps) ran as RPC **middleware** — after a frame was parsed and matched to a registered method. Malformed frames and unknown-method frames return error responses *before* the middleware chain executes, so they never debited the request-rate bucket. A flood of tiny invalid / unknown-method frames was bounded only by the per-connection byte limiter and queue limits — well above the intended request rate, and multiplied across connections.

## Fix

Move request-rate limiting to the connection (**frame**) layer, beside the existing byte budget, so every inbound frame is counted *before* parsing and routing — malformed and unknown-method frames included.

- Add `RequestTokenBucket` (one token per frame, size-independent) and `CompositeFrameRateLimiter` (admits only if every member admits) to `pkg/rpc`.
- Build the per-connection `FrameRateLimiter` as a composite of the byte + request buckets in `runtime.go`; fail fast at startup if `RATE_LIMIT_BURST < 1` while enabled (mirrors the byte-burst check).
- Remove `RateLimitMiddleware` and its router/config plumbing.

Net **−73 lines** including new tests.

## Behavior changes

- **Overrun now closes the connection** (consistent with the byte limiter) instead of returning a `"rate limit exceeded"` RPC error with the connection kept open. Appropriate for an anti-flood control; `burst=20` gives legitimate clients headroom.
- Coverage is now **pre-parse** — even the JSON unmarshal cost of invalid frames is avoided once the budget is exhausted.
- Config knobs `NITRONODE_RATE_LIMIT_PER_SEC` / `NITRONODE_RATE_LIMIT_BURST` (default 10 / 20) are unchanged — enforced one layer lower.

## Tests

- `pkg/rpc/rate_limiter_test.go`: burst / refill / cap / sub-unit-burst for `RequestTokenBucket`; composite reject + nil/empty members.
- `go build ./...`, `go vet`, and tests for `pkg/rpc`, `nitronode/api`, `nitronode` all green.

## Release Notes

* **New Features**
  * Added validation for app session and escrow channel identifiers
  * Introduced request-based rate limiting alongside byte-rate limiting
  * Enhanced pagination handling with improved limit parameter support
  * Added case-insensitive duplicate asset symbol detection

* **Bug Fixes**
  * Fixed quorum weight validation to prevent integer overflow
  * Improved deposit state validation and error handling
  * Enhanced session key revocation tracking for active-to-inactive transitions
  * Fixed pagination metadata computation to reject invalid limits

* **Improvements**
  * Migrated rate limiting from RPC layer to WebSocket connection level
  * Added user state locking during home channel operations
  * Enhanced off-chain signature backfill on channel opening
  * Strengthened channel status validation for state transitions


nitronode/runtime.go
```diff
@@ -307,17 +305,43 @@ func InitBackbone() *Backbone {
 			)
 		}
 	}
+
+	// Per-connection request-count budget. Enforced at the frame layer alongside
+	// the byte budget so a flood of tiny frames — malformed or unknown-method
+	// frames included, which never reach the handler chain — is throttled before
```
ihsraham, line 289: Good move getting request limiting out of the handler chain; non-empty malformed JSON and unknown-method frames now hit this bucket before parse/route.

One remaining L-14 bypass: `readMessages` skips zero-length WS messages before calling `conn.frameRateLimiter.Admit(...)` (`pkg/rpc/connection.go:403-408`). Empty payloads are malformed RPC frames too, so they can still be flooded without spending request-count tokens or byte budget. Please either run the request limiter before the empty-message skip, or close on empty/protocol-invalid frames.

ihsraham (changes requested):
Requesting changes for one closure gap. Moving request limiting out of the handler chain is the right direction and it covers non-empty malformed JSON plus unknown-method frames before parse/route lookup. The remaining issue is that zero-length WS messages still bypass the limiter.

nitronode/runtime.go
philanton: Yep, good catch — empty frames are malformed RPC too, so they should spend a token like anything else. Moved `Admit` ahead of the empty-message skip in [5711409](https://github.com/layer-3/nitrolite/pull/819/changes/5711409da7c52af8b4ad383a430915892d6e080e), and added a regression test that floods an empty frame and asserts the limiter still closes the connection.

ihsraham (approved):
Approving for L-14. `Admit` now runs before the empty-message skip, and the added empty-frame regression test covers the gap from my previous review. Non-empty malformed JSON and unknown-method frames were already covered by the frame-layer move, so the finding looks closed to me.

I am leaving the current merge conflict to the branch update path, not treating it as an audit-closure blocker.

nksazonov (approved):
Excellent work. The PR correctly addresses both attack vectors from the audit finding. `Admit()` now runs before JSON parsing and routing, so empty frames, malformed frames, and valid-JSON unknown-method frames are all charged before reaching the processing pipeline. Connection closure on overrun is appropriate for anti-flood control and consistent with the existing byte limiter.

PR #820, opened 2026-06-05
MF3-H02: lock home channel before reading status in event handlers

philanton:
## Finding

Home-channel event handlers read the channel via `GetChannelByID` **before** acquiring `LockUserState`, then mutated/persisted that pre-lock snapshot. `LockUserState` locks the user *balance* row, not the channel row — so a concurrent `submit_state` Finalize could flip the channel `Open→Closing` in the read→lock window.

`HandleHomeChannelCheckpointed`'s non-challenged path persists the channel snapshot verbatim (it only reassigns `Status` on the `Challenged` branch). A stale `Open` snapshot therefore overwrote the committed `Closing`, **reopening a finalized channel** — bypassing the C-01 remediation and enabling the post-finalize epoch-rebind / fake-`HomeDeposit` spend chain.

## Fix

New `DBStore.LockUserStateForHomeChannel(channelID)`:
- Derives the `(wallet, asset)` lock key from the channel **inside SQL** — no pre-lock Go snapshot exists.
- Postgres: ensures the balance row, then `SELECT c.* FROM channels c JOIN user_balances b ... WHERE c.channel_id = ? FOR UPDATE OF b` — locks the balance row (same row/strength as `LockUserState`) and reads the channel in one statement. The stale-snapshot window is structurally impossible, not just guarded.
- sqlite (tests): resolve → `LockUserState` → read.

All four home-channel handlers (`Created`, `Checkpointed`, `Challenged`, `Closed`) now lock-then-read via this method. Escrow handlers are unchanged (different status surface, out of scope).

Wired into `core.ChannelHubEventHandlerStore`, `evm.ChannelHubReactorStore`, `database.DatabaseStore` + both mocks.

## Tests

- Rewired 25 home-handler tests to the new mock; escrow tests untouched.
- `TestHandleHomeChannelCheckpointed_DoesNotReopenFinalizedChannel` — post-lock `Closing` snapshot must persist `Closing`, never reopen to `Open`.
- `TestDBStore_LockUserStateForHomeChannel` — returns channel + ensures balance row; nil for missing channel.

`go build ./...`, `go vet`, and `go test` (event_handlers, evm, store/database) pass.

> Note: the Postgres branch (`FOR UPDATE OF` on the join) is covered by the gated `store/database/test` integration suite, not the default run — same convention as the existing `LockUserState` Postgres branch.

### DIFF
--- a/nitronode/event_handlers/service.go
+++ b/nitronode/event_handlers/service.go
@@ -57,13 +57,9 @@ func (s *EventHandlerService) HandleNodeBalanceUpdated(ctx context.Context, tx c
 func (s *EventHandlerService) HandleHomeChannelCreated(ctx context.Context, tx core.ChannelHubEventHandlerStore, event *core.HomeChannelCreatedEvent) error {
 	logger := log.FromContext(ctx)
 	chanID := event.ChannelID
-	// Acquire the user's balance-row lock and read the channel under it before mutating
-	// channel status or backfilling the off-chain head. Receiver/release issuance paths
-	// lock the same row up front and then re-check Status via CheckActiveChannel; without
-	// this lock an in-flight RPC can read Status=Void, decide to store an unsigned receiver
-	// row, and commit before we flip to Open — leaving the latest head unsigned on a
-	// technically opened channel. The lock+read is a single store call so the channel
-	// snapshot is consistent with the lock. See HandleHomeChannelCheckpointed and
+	// Acquire the user's balance-row lock and read the channel under it before mutating status:
+	// the lock guards against a concurrent submit_state flipping channel status (e.g. Void→Open
+	// receiver issuance) between the read and our write. See HandleHomeChannelCheckpointed and
 	// HandleHomeChannelClosed for the same pattern.
 	channel, err := tx.LockUserStateForHomeChannel(chanID)
 	if err != nil {
--- a/nitronode/event_handlers/service.go
+++ b/nitronode/event_handlers/service.go
@@ -57,7 +57,15 @@ func (s *EventHandlerService) HandleNodeBalanceUpdated(ctx context.Context, tx c
 func (s *EventHandlerService) HandleHomeChannelCreated(ctx context.Context, tx core.ChannelHubEventHandlerStore, event *core.HomeChannelCreatedEvent) error {
 	logger := log.FromContext(ctx)
 	chanID := event.ChannelID
-	channel, err := tx.GetChannelByID(chanID)
+	// Acquire the user's balance-row lock and read the channel under it before mutating
+	// channel status or backfilling the off-chain head. Receiver/release issuance paths
+	// lock the same row up front and then re-check Status via CheckActiveChannel; without
+	// this lock an in-flight RPC can read Status=Void, decide to store an unsigned receiver
+	// row, and commit before we flip to Open — leaving the latest head unsigned on a
+	// technically opened channel. The lock+read is a single store call so the channel
+	// snapshot is consistent with the lock. See HandleHomeChannelCheckpointed and
+	// HandleHomeChannelClosed for the same pattern.
+	channel, err := tx.LockUserStateForHomeChannel(chanID)
 	if err != nil {
 		return err
 	}
@@ -68,18 +76,6 @@ func (s *EventHandlerService) HandleHomeChannelCreated(ctx context.Context, tx c
 	if channel.Type != core.ChannelTypeHome {
 		logger.Warn("channel type mismatch during HomeChannelCreated event", "channelId", chanID, "expectedType", core.ChannelTypeHome, "actualType", channel.Type)
 		return nil
-	}
-
-	// Acquire the user's balance-row lock before mutating channel status or
-	// backfilling the off-chain head. Receiver/release issuance paths lock
-	// the same row up front and then re-check Status via CheckActiveChannel;
-	// without this lock an in-flight RPC can read Status=Void, decide to store
-	// an unsigned receiver row, and commit before we flip to Open — leaving
-	// the latest head unsigned on a technically opened channel. See
-	// HandleHomeChannelCheckpointed and HandleHomeChannelClosed for the same
-	// pattern.
-	if _, err := tx.LockUserState(channel.UserWallet, channel.Asset); err != nil {
-		return err
 	}
 
 	if channel.Status >= core.ChannelStatusOpen {
@@ -137,7 +133,17 @@ func (s *EventHandlerService) HandleHomeChannelMigrated(ctx context.Context, tx 
 func (s *EventHandlerService) HandleHomeChannelCheckpointed(ctx context.Context, tx core.ChannelHubEventHandlerStore, event *core.HomeChannelCheckpointedEvent) error {
 	logger := log.FromContext(ctx)
 	chanID := event.ChannelID
-	channel, err := tx.GetChannelByID(chanID)
+	// Acquire the user's balance-row lock and read the channel under it before mutating
+	// channel status or backfilling the off-chain head. Reading the channel before the
+	// lock would race: a concurrent submit_state can co-sign a Finalize and flip the
+	// channel Open→Closing between the read and the lock, and the non-challenged path
+	// below persists the channel snapshot verbatim — a pre-lock Open snapshot would
+	// silently reopen the finalized channel. Receiver issuance paths lock the same row
+	// and then re-check Status; without this lock an RPC can read Status=Challenged,
+	// decide to store an unsigned receiver row, but commit after we flip to Open and
+	// backfill the prior head — leaving the latest head unsigned on an Open channel. See
+	// HandleHomeChannelClosed for the same pattern.
+	channel, err := tx.LockUserStateForHomeChannel(chanID)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


nitronode/store/database/db_store.go
```diff
@@ -145,6 +145,63 @@ func (s *DBStore) LockUserState(wallet, asset string) (decimal.Decimal, error) {
...
+		return nil, fmt.Errorf("failed to ensure user balance row for channel %s: %w", channelID, err)
+	}
+
+	// Lock only the balance row (FOR UPDATE OF b) and read the channel in the same statement.
+	var dbChannel Channel
+	result := s.db.Raw(`
+		SELECT c.* FROM channels c
+		JOIN user_balances b ON b.user_wallet = c.user_wallet AND b.asset = c.asset
+		WHERE c.channel_id = ?
+		FOR UPDATE OF b
```
ihsraham: I would block on this because `FOR UPDATE OF b` only locks rows from `user_balances`; the joined `channels c` row is still read as part of the same statement snapshot. A checkpoint handler can start while `c.status` is still `Open`, wait here on `b` while a concurrent Finalize commits `Closing`, and still return the stale `Open` channel to `HandleHomeChannelCheckpointed`, which then persists it.

To close H-02, please acquire the `(wallet, asset)` balance lock first, then issue a separate channel read in the same transaction after the lock has been acquired. It would also be worth adding a Postgres concurrency regression test for that wait-and-refresh case, since the current test mocks the desired post-lock return rather than exercising this SQL branch.

ihsraham (changes requested):
Thanks, the handler-side change is moving in the right direction: the home-channel event handlers now go through one helper before mutating status, and the production reactor path is transaction-wrapped.

I am requesting changes on one blocker because the Postgres helper still does not guarantee a post-lock channel refresh, so H-02 is not fully closed yet. I left the specific note inline.

I am not treating the current SDK/MCP audit failure as part of this review, since that dependency fix is being handled separately and should be pulled in before merge.

nitronode/store/database/db_store.go
philanton: Yep, you're correct — `FOR UPDATE OF b` only re-evaluates the locked `user_balances` row after the wait, the joined `channels` row is still served from the statement-start snapshot, so a Finalize that commits `Closing` while we're blocked on the balance lock slips through as a stale `Open`.

Fixed in [8496903](https://github.com/layer-3/nitrolite/pull/820/changes/84969032fc6cdbf38b87471de77edb1a2883bc7b): acquire the `(wallet, asset)` balance lock first, then read the channel in a separate statement so it takes a fresh snapshot after the lock is held.

Also added the Postgres concurrency regression you asked for — it holds the balance lock, flips status to `Closing`, and commits while the call is blocked on the lock, then asserts the returned status is `Closing`. I verified it fails against the old single-statement query (returns `Open`) and passes with the fix.

One heads-up: it only runs under `TEST_DB_DRIVER=postgres`, and our `test-go.yml` runs bare `go test ./...` with no driver set, so this (and the other pg-gated tests) skip in CI today — I ran it locally against the testcontainers Postgres. Wiring a pg lane into CI is its own thing, out of scope here.

ihsraham (approved):
Approving for H-02 closure. The new commit fixes the stale snapshot issue by acquiring the `(wallet, asset)` balance lock first and reading the channel in a separate statement after the lock is held. The added Postgres regression covers the wait-and-refresh case Anton described.

Non-blocking follow-up: that regression only runs with `TEST_DB_DRIVER=postgres`, so a dedicated pg CI lane would make this easier to keep covered. I am not treating that as a blocker for this fix.

nitronode/event_handlers/service.go
```diff
@@ -57,7 +57,15 @@ func (s *EventHandlerService) HandleNodeBalanceUpdated(ctx context.Context, tx c
...
 	chanID := event.ChannelID
-	channel, err := tx.GetChannelByID(chanID)
+	// Acquire the user's balance-row lock and read the channel under it before mutating
+	// channel status or backfilling the off-chain head. Receiver/release issuance paths
+	// lock the same row up front and then re-check Status via CheckActiveChannel; without
+	// this lock an in-flight RPC can read Status=Void, decide to store an unsigned receiver
+	// row, and commit before we flip to Open — leaving the latest head unsigned on a
+	// technically opened channel. The lock+read is a single store call so the channel
+	// snapshot is consistent with the lock. See HandleHomeChannelCheckpointed and
+	// HandleHomeChannelClosed for the same pattern.
```
nksazonov: I think this comment needs shortening. Specifying that "user's balance-row lock guards against race-conditions of channel status being changed in the meantime" should be enough.

nitronode/store/database/db_store.go
```diff
@@ -145,6 +145,62 @@ func (s *DBStore) LockUserState(wallet, asset string) (decimal.Decimal, error) {
...
+	if _, err := s.LockUserState(key.UserWallet, key.Asset); err != nil {
+		return nil, err
+	}
+
+	// Read the channel only after the lock is held. A single SELECT ... FOR UPDATE OF b that
+	// joins channels would return c.* from the statement-start snapshot — a Finalize that flips
+	// status to Closing while we wait on the balance lock would not be reflected. This separate
+	// statement takes a fresh snapshot once the lock is acquired, so the returned status reflects
+	// any such committed transition.
+	return s.GetChannelByID(channelID)
```
nksazonov, line 205: The "fresh snapshot" guarantee here only holds under **READ COMMITTED** isolation (Postgres default). Under `REPEATABLE READ` or `SERIALIZABLE`, every statement in the same transaction sees the snapshot taken at transaction start — meaning the post-lock `GetChannelByID` would return the same stale status as before the lock was acquired, silently defeating the fix.

The comment currently reads as though this always works without stating the prerequisite. Suggest adding:
```
// N.B. requires READ COMMITTED isolation (Postgres default). Under REPEATABLE READ
// the post-lock statement still sees the transaction-start snapshot, negating the fix.
```

nitronode/store/database/channel_test.go
```diff
@@ -194,6 +197,132 @@ func TestDBStore_GetChannelByID(t *testing.T) {
...
+		}
+		got := make(chan lockResult, 1)
+		go func() {
+			ch, err := store.LockUserStateForHomeChannel(channel.ChannelID)
+			got <- lockResult{channel: ch, err: err}
+		}()
+
+		// Give the call time to start and block on the balance lock, then let the competing
+		// transaction commit the Closing status.
+		time.Sleep(200 * time.Millisecond)
```
nksazonov: `time.Sleep(200ms)` is non-deterministic: under CI load the goroutine might not yet be executing `SELECT … FOR UPDATE` when the sleep expires, so the competing transaction commits *before* the lock is ever contended. The test then passes, but the race is never exercised and the regression pin becomes a no-op.

A deterministic alternative: have the competing transaction signal only after it has confirmed lock contention — e.g., poll `pg_stat_activity` for a waiting query, or use a `pg_advisory_lock` rendezvous before `close(proceed)`.

pkg/core/interface.go
```diff
@@ -156,6 +156,13 @@ type ChannelHubEventHandlerStore interface {
 	// in tests.
 	LockUserState(wallet, asset string) (decimal.Decimal, error)
 
+	// LockUserStateForHomeChannel locks the balance row of the user owning channelID and
+	// returns the channel read under that lock, deriving the lock key from the channel in
+	// SQL so no stale pre-lock snapshot is used. Event handlers must use this instead of a
+	// GetChannelByID + LockUserState pair, which reads channel status before the lock and
+	// races a concurrent submit_state finalization. Returns nil if the channel is absent.
+	LockUserStateForHomeChannel(channelID string) (*Channel, error)
```
nksazonov: "no stale pre-lock snapshot is used" overstates the guarantee. The non-postgres implementation (the sqlite path used by all unit tests) reads the channel *before* the lock and returns that pre-lock snapshot — see `db_store.go` lines 165–173. The guarantee holds only on Postgres; unit tests backed by sqlite cannot detect a regression to the old racy read order.

Suggest narrowing the wording to something like: `// returns the channel read under that lock on postgres; on non-postgres the snapshot is taken before the lock for test compatibility.`

Same note applies to `nitronode/store/database/interface.go:31–33`.

nksazonov (approved):
Clean implementation of the lock-before-read pattern across all four home-channel event handlers. The new regression test `TestHandleHomeChannelCheckpointed_DoesNotReopenFinalizedChannel` directly pins the race described in the audit finding.

nitronode/event_handlers/service.go
philanton: Yep, trimmed it down in [7f6d275](https://github.com/layer-3/nitrolite/pull/820/changes/7f6d27562a4509fc77f503de88ba599c1f0c8ada) — now just says the lock guards against a concurrent submit_state flipping status between the read and our write, plus the see-also pointer.

nitronode/store/database/db_store.go
philanton: Good catch — added the caveat in [7f6d275](https://github.com/layer-3/nitrolite/pull/820/changes/7f6d27562a4509fc77f503de88ba599c1f0c8ada). Worth noting we never override the isolation level (every tx goes through GORM `db.Transaction(...)` with no `sql.TxOptions`), so we're on Postgres' READ COMMITTED default and the refresh holds. But you're right that it's an implicit prerequisite, so I documented that bumping to REPEATABLE READ/SERIALIZABLE would silently negate the fix.

nitronode/store/database/channel_test.go
philanton: Yeah, fair — the sleep could let the competitor commit before contention and turn the pin into a no-op. Replaced it in [7f6d275](https://github.com/layer-3/nitrolite/pull/820/changes/7f6d27562a4509fc77f503de88ba599c1f0c8ada) with a `require.Eventually` that polls `pg_locks WHERE NOT granted`, so we only release `proceed` once a backend is genuinely blocked on the lock. Verified against testcontainers Postgres — passes, and `Eventually` observing the ungranted lock confirms the wait-and-refresh path is actually contended.

pkg/core/interface.go
philanton: Yep, you're correct — the sqlite path reads before the lock, so the old wording overstated it. Narrowed in [7f6d275](https://github.com/layer-3/nitrolite/pull/820/changes/7f6d27562a4509fc77f503de88ba599c1f0c8ada): now says the channel is read under the lock on postgres, while non-postgres takes the snapshot before the lock for test compatibility. Applied the same fix to `store/database/interface.go` too.

PR #822, opened 2026-06-05
MF3-I12, I13, I14: nitronode audit remediations

philanton:
Remediates three audit findings in nitronode. All informational/defense-in-depth — no exploit in the current healthy flow; each hardens an implicit invariant to fail closed.

## MF3-I12 — channel session-key asset validation
`SubmitSessionKeyState` for channel session keys only length-checked `coreState.Assets`. Now, per asset, `validateSessionKeyAssets` rejects:
- **non-canonical** (`asset != strings.ToLower(asset)`) — assets are part of the user-signed metadata hash, so the node rejects rather than rewrites, keeping the signed payload identical to what is stored;
- **duplicates** (after normalization);
- **unsupported** (`GetAssetDecimals` errors).

An empty assets list is accepted only as a revocation/deactivation state: it is rejected when `expires_at` is in the future (it would leave an active current version authorizing no usable asset) and allowed only when `expires_at <= now`. All checks run before signature validation/storage.

## MF3-I13 — reject submit_state on a stateless active channel
`SubmitState`'s default branch silently bootstrapped a synthetic `Void` state when no stored state existed for an active home channel. An active channel always has its initial state stored atomically by `request_creation`, so a nil state is an invariant violation, not a first-submit. Now fails closed with `active channel has no stored state`. The `NewVoidState` fallback stays in the channel-creation path.

## MF3-I14 — assert user_balances update affects one row
`StoreUserState` checked only the GORM `Error` from the `user_balances` `Updates()` and ignored `RowsAffected`. A caller that stored a state without first calling `LockUserState` would silently match zero rows, leaving `channel_states` and the cached balance divergent without failing the transaction. Now requires exactly one affected row. Tests that used `StoreUserState` without locking are updated to lock-first via a `storeLocked` helper, plus a negative test for the new error.

All production call sites already lock-before-store / store-initial-state, so no behavior change in the healthy flow.

### DIFF
--- a/nitronode/api/channel_v1/submit_session_key_state.go
+++ b/nitronode/api/channel_v1/submit_session_key_state.go
@@ -2,6 +2,7 @@ package channel_v1
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 	"time"
 
@@ -69,6 +70,13 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
 	}
 	if len(coreState.Assets) > h.maxSessionKeyIDs {
 		c.Fail(rpc.Errorf("invalid_session_key_state: assets array exceeds maximum length of %d", h.maxSessionKeyIDs), "")
+		return
+	}
+	// An empty assets list authorizes no usable asset, so it is only meaningful as a
+	// revocation/deactivation state (expires_at <= now). Rejecting it for a future
+	// expires_at prevents an active current version that authorizes nothing.
+	if len(coreState.Assets) == 0 && coreState.ExpiresAt.After(time.Now()) {
+		c.Fail(rpc.Errorf("invalid_session_key_state: assets must be non-empty unless expires_at is in the past"), "")
 		return
 	}
 	if err := h.validateSessionKeyAssets(coreState.Assets); err != nil {
@@ -190,15 +198,15 @@ func (h *Handler) validateSessionKeyAssets(assets []string) error {
 	for _, asset := range assets {
 		normalized := strings.ToLower(asset)
 		if asset != normalized {
-			return rpc.Errorf("non-canonical asset '%s', expected '%s'", asset, normalized)
+			return fmt.Errorf("non-canonical asset '%s', expected '%s'", asset, normalized)
 		}
 		if _, ok := seen[normalized]; ok {
-			return rpc.Errorf("duplicate asset '%s'", asset)
+			return fmt.Errorf("duplicate asset '%s'", asset)
 		}
 		seen[normalized] = struct{}{}
 
 		if _, err := h.memoryStore.GetAssetDecimals(normalized); err != nil {
-			return rpc.Errorf("unsupported asset '%s'", asset)
+			return fmt.Errorf("unsupported asset '%s'", asset)
 		}
 	}
 	return nil
--- a/nitronode/api/channel_v1/submit_session_key_state.go
+++ b/nitronode/api/channel_v1/submit_session_key_state.go
@@ -71,6 +71,10 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
 		c.Fail(rpc.Errorf("invalid_session_key_state: assets array exceeds maximum length of %d", h.maxSessionKeyIDs), "")
 		return
 	}
+	if err := h.validateSessionKeyAssets(coreState.Assets); err != nil {
+		c.Fail(rpc.Errorf("invalid_session_key_state: %v", err), "")
+		return
+	}
 	if coreState.UserSig == "" {
 		c.Fail(rpc.Errorf("invalid_session_key_state: user_sig is required"), "")
 		return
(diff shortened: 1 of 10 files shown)


nitronode/api/channel_v1/submit_session_key_state.go
```diff
@@ -174,3 +178,28 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
 		"sessionKey", coreState.SessionKey,
 		"version", coreState.Version)
 }
+
+// validateSessionKeyAssets rejects an assets list that contains a non-canonical asset, a
+// duplicate, or an asset the node does not support. Assets must already be lowercased because
+// they are persisted lowercased (StoreChannelSessionKeyState) and looked up lowercased
+// (ValidateChannelSessionKeyForAsset); rejecting a non-canonical asset keeps the user-signed
+// payload identical to what is stored. Empty assets lists are allowed; the per-asset checks
```
ihsraham, line 194: This leaves one I-12 case open. The audit asked for `assets` to be empty only when `expires_at <= now`, so an empty list is an explicit revocation/deactivation state. With this helper taking only `assets`, a signed state with `assets: []` and a future `expires_at` still passes validation and can become the active current version while authorizing no usable asset.

Before approval, please reject `len(coreState.Assets) == 0` when `coreState.ExpiresAt` is in the future, and add the matching regression test. The duplicate, canonical, and supported-asset checks already look good.

ihsraham (changes requested):
Thanks Anton. I-13 and I-14 look closed to me, and the new I-12 per-asset checks for canonical, duplicate, and unsupported assets are the right direction. I am requesting changes on the remaining I-12 case because the audit also required empty asset lists to be accepted only as revocation/deactivation states.

nitronode/api/channel_v1/submit_session_key_state.go
```diff
@@ -71,6 +71,10 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
 		c.Fail(rpc.Errorf("invalid_session_key_state: assets array exceeds maximum length of %d", h.maxSessionKeyIDs), "")
 		return
 	}
+	if err := h.validateSessionKeyAssets(coreState.Assets); err != nil {
+		c.Fail(rpc.Errorf("invalid_session_key_state: %v", err), "")
```
nksazonov, line 83: `validateSessionKeyAssets` returns `rpc.Errorf(...)`, which is then passed as the `%v` arg to another `rpc.Errorf(...)`. If `rpc.Error`'s `.Error()` method ever includes more than the bare message string (error codes, type prefix, etc.), the double-wrap will leak internal structure into the JSON-RPC error payload.

Every other call site in this file that wraps an external error (e.g. `core.ValidateChannelSessionKeyStateV1` at line 88) wraps a standard `error` — not an `rpc.Error`. `validateSessionKeyAssets` is the exception.

**Suggestion:** Have `validateSessionKeyAssets` return `fmt.Errorf(...)` instead of `rpc.Errorf(...)`. The call site stays identical and the outer `rpc.Errorf("invalid_session_key_state: %v", err)` then cleanly formats a plain error string.

nitronode/store/database/state_test.go
```diff
@@ -46,7 +57,7 @@ func TestDBStore_StoreUserState_ApplicationID(t *testing.T) {
 		_, err := store.LockUserState("0xuserapp", "USDC")
 		require.NoError(t, err)
 
-		require.NoError(t, store.StoreUserState(newState("state-app"), "my-app"))
+		require.NoError(t, storeLocked(t, store, newState("state-app"), "my-app"))
```
nksazonov: `storeLocked` already calls `LockUserState` internally (lines 22–23 of the helper). The explicit `LockUserState` at line 57 is now redundant — it creates the row, then `storeLocked` immediately re-locks it. In the SQLite path `LockUserState` queries then re-creates nothing; in the PostgreSQL path it re-runs `INSERT ON CONFLICT DO NOTHING` + `SELECT FOR UPDATE`. Harmless today, but it obscures whether the pre-lock is intentional setup or just boilerplate.

The same pattern appears throughout `db_store_test.go`: lines 39–42, 95–101, 167–171, 223–227, 306–309, 362–365, 442–445, 521–524, 586–589, 639–642, 720–722, 885–929.

**Suggestion:** Remove the explicit `store.LockUserState(...)` / `require.NoError(t, err)` pair wherever it immediately precedes a `storeLocked` call for the same `(wallet, asset)`. Let `storeLocked` own the lock. Apply the same cleanup in `db_store_test.go`.

nksazonov (commented):
Good work — all three audit remediations are correctly implemented and the logic is sound. Two minor things below worth cleaning up before landing.

nitronode/api/channel_v1/submit_session_key_state.go
philanton: Yeah, good catch — fixed in [468e128](https://github.com/layer-3/nitrolite/pull/822/changes/468e1280ba71916b5b3f4d04294c3901d06c0433). It now rejects an empty `assets` list when `expires_at` is in the future, so an empty list is only meaningful as a revoke/deactivation state. Added regression tests for both the rejected (future) and the allowed (past) cases.

nitronode/api/channel_v1/submit_session_key_state.go
philanton: Yep, makes sense — switched `validateSessionKeyAssets` to `fmt.Errorf` so the outer `rpc.Errorf` does the wrapping. [468e128](https://github.com/layer-3/nitrolite/pull/822/changes/468e1280ba71916b5b3f4d04294c3901d06c0433)

nitronode/store/database/state_test.go
philanton: Done — removed the redundant explicit `LockUserState` pre-locks in both `state_test.go` and `db_store_test.go` wherever `storeLocked` already locks the same `(wallet, asset)`. [468e128](https://github.com/layer-3/nitrolite/pull/822/changes/468e1280ba71916b5b3f4d04294c3901d06c0433)

nksazonov (approved):
Great!

ihsraham (approved):
Thanks, this resolves my earlier I-12 request.

I re-checked `468e128`: future `assets: []` is now rejected before signature validation or any store call, and the tests cover both the rejected future case and the allowed past-expiry revocation case. I-13 and I-14 still look closed, and the Go test plus SAST/SCA passes did not turn up a PR-introduced blocker.

Only non-blocking cleanup I noticed is wording drift: the PR body and the helper comment still read like empty asset lists are generally allowed, while the code now allows them only for expired revocation states. I would not block landing on that.

PR #826, opened 2026-06-08
MF3-L12: validate session-key scope ID formats at request boundary

philanton:
## What

`app_sessions.v1.submit_session_key_state` only validated array length and lowercasing for `application_ids` and `app_session_ids`. It did **not** validate each ID's format. As a result malformed-but-lowercase IDs passed request validation, reached signature verification, and could persist as inert, unusable scopes (never matching a real application/app-session).

This aligns the handler with the validation already applied in `create_app_session` and the RPC `application_id` query-param path.

## Changes

- `application_ids`: replaced the lowercase-only loop with `app.IsValidApplicationID()` (regex `^[a-z0-9_-]{1,66}$`, which already enforces lowercase).
- `app_session_ids`: added `core.IsValidHash()` (`^0x[0-9a-fA-F]{64}$`) before the lowercase canonical check. `IsValidHash` is case-insensitive, so the lowercase check is kept.
- Both checks run **before** signature verification — fail fast at the request boundary. Array-length caps unchanged.
- Tests: refactored the two stale lowercase-only tests into a shared `submitStateExpectingError` helper; added format-rejection cases (uppercase / illegal char / over-length application IDs; non-hash / short / no-`0x` / uppercase-hex app-session IDs).

## Severity

Low — hardening/consistency, not an exploitable vuln. The request is self-authenticated (both `user_sig` and `session_key_sig` gate persistence, so no injection into another user's state), and DB column constraints (`VARCHAR(66)` / `CHAR(66)`) already reject oversized values. Worst real case is a user signing their own malformed scope that silently never authorizes anything; this rejects it up front instead.

## Test

```
go build ./nitronode/api/app_session_v1/
go vet  ./nitronode/api/app_session_v1/
go test ./nitronode/api/app_session_v1/ -run TestSubmitSessionKeyState
```
All green.

### DIFF
--- a/nitronode/api/app_session_v1/submit_session_key_state.go
+++ b/nitronode/api/app_session_v1/submit_session_key_state.go
@@ -78,17 +78,13 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
 	}
 	for _, id := range coreState.ApplicationIDs {
 		if !app.IsValidApplicationID(id) {
-			c.Fail(rpc.Errorf("invalid_session_key_state: application_ids must match %s, got: %s", app.ApplicationIDRegex.String(), id), "")
+			c.Fail(rpc.Errorf("invalid_session_key_state: application_ids must contain only lowercase letters, digits, dashes, and underscores (max 66 chars), got: %s", id), "")
 			return
 		}
 	}
 	for _, id := range coreState.AppSessionIDs {
-		if !core.IsValidHash(id) {
-			c.Fail(rpc.Errorf("invalid_session_key_state: app_session_ids must be 0x-prefixed 32-byte hashes, got: %s", id), "")
-			return
-		}
-		if id != strings.ToLower(id) {
-			c.Fail(rpc.Errorf("invalid_session_key_state: app_session_ids must be lowercase, got: %s", id), "")
+		if !core.IsValidHash(id, true) {
+			c.Fail(rpc.Errorf("invalid_session_key_state: app_session_ids must be 0x-prefixed 32-byte lowercase hex hashes, got: %s", id), "")
 			return
 		}
 	}
--- a/nitronode/api/app_session_v1/submit_session_key_state.go
+++ b/nitronode/api/app_session_v1/submit_session_key_state.go
@@ -77,12 +77,16 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
 		return
 	}
 	for _, id := range coreState.ApplicationIDs {
-		if id != strings.ToLower(id) {
-			c.Fail(rpc.Errorf("invalid_session_key_state: application_ids must be lowercase, got: %s", id), "")
+		if !app.IsValidApplicationID(id) {
+			c.Fail(rpc.Errorf("invalid_session_key_state: application_ids must match %s, got: %s", app.ApplicationIDRegex.String(), id), "")
 			return
 		}
 	}
 	for _, id := range coreState.AppSessionIDs {
+		if !core.IsValidHash(id) {
+			c.Fail(rpc.Errorf("invalid_session_key_state: app_session_ids must be 0x-prefixed 32-byte hashes, got: %s", id), "")
+			return
+		}
 		if id != strings.ToLower(id) {
 			c.Fail(rpc.Errorf("invalid_session_key_state: app_session_ids must be lowercase, got: %s", id), "")
 			return
--- a/docs/api.yaml
+++ b/docs/api.yaml
@@ -409,12 +409,20 @@ types:
           type: array
           items:
             type: string
-          description: Application IDs associated with this session key
+          description: >
+            Application IDs associated with this session key. Each entry must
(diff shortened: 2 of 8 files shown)


nitronode/api/app_session_v1/submit_session_key_state.go
```diff
@@ -77,12 +77,16 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
...
 		}
 	}
 	for _, id := range coreState.AppSessionIDs {
+		if !core.IsValidHash(id) {
+			c.Fail(rpc.Errorf("invalid_session_key_state: app_session_ids must be 0x-prefixed 32-byte hashes, got: %s", id), "")
+			return
+		}
 		if id != strings.ToLower(id) {
 			c.Fail(rpc.Errorf("invalid_session_key_state: app_session_ids must be lowercase, got: %s", id), "")
 			return
```
nksazonov: The two guards produce two different error messages for what the protocol defines as a single rule: "must be a 0x-prefixed 32-byte lowercase hex hash." A caller with an uppercase-hex input (e.g. `0xABCD…`) passes the first check and hits the second, receiving a different error code than a caller with a non-hash string. API consumers have to handle two error surfaces for one conceptual constraint.

Suggestion: add `core.IsValidLowercaseHash()` with regex `^0x[0-9a-f]{64}$` (the exact canonical form), replace both guards with a single call and a single error message, and keep the test coverage for both uppercase-hex and non-hash inputs confirming they both fail via the same path.

nitronode/api/app_session_v1/submit_session_key_state.go
```diff
@@ -77,12 +77,16 @@ func (h *Handler) SubmitSessionKeyState(c *rpc.Context) {
 		return
 	}
 	for _, id := range coreState.ApplicationIDs {
-		if id != strings.ToLower(id) {
-			c.Fail(rpc.Errorf("invalid_session_key_state: application_ids must be lowercase, got: %s", id), "")
+		if !app.IsValidApplicationID(id) {
+			c.Fail(rpc.Errorf("invalid_session_key_state: application_ids must match %s, got: %s", app.ApplicationIDRegex.String(), id), "")
```
nksazonov: The regex string (`^[a-z0-9_-]{1,66}$`) in the error message is an internal implementation detail that callers shouldn't need to parse to understand what went wrong. If the regex ever changes, existing clients matching on the error text will silently break.

Suggestion: replace the dynamic `app.ApplicationIDRegex.String()` with a stable human-readable description, for example: `"application_ids must contain only lowercase letters, digits, dashes, and underscores (max 66 chars), got: %s"`.

nksazonov (approved):
Good job — the fix correctly addresses the audit finding by moving both `IsValidApplicationID` and `IsValidHash` checks ahead of signature verification, closing the malformed-but-lowercase bypass. Two minor polish items below.

ihsraham (approved):
No blockers from me. Approving for L-12 closure.

The fix moves both scope-ID format checks to the request boundary before signature verification and storage. `application_ids` now use `app.IsValidApplicationID`, and `app_session_ids` require a 32-byte hash shape plus lowercase canonical form. The new tests also cover malformed app IDs and app-session IDs and assert those requests do not reach `LockSessionKeyState`, which is the right proof for this finding.

What is still worth cleaning up is not a closure blocker: the public docs/API wording still describes these arrays as loose lowercase strings rather than the stricter formats now enforced. I would update that in the release/docs pass so clients do not copy the old shape.

nitronode/api/app_session_v1/submit_session_key_state.go
philanton: Yeah, agreed. Done in [59d9f15](https://github.com/layer-3/nitrolite/pull/826/changes/59d9f15b4b2b60370d0f7310caec0dfe42f5991f): added `core.IsValidHash(s, requireLowercase bool)` backed by `LowercaseHashRegex` (`^0x[0-9a-f]{64}$`), and the two guards are now one call with a single message — "app_session_ids must be 0x-prefixed 32-byte lowercase hex hashes". Tests still cover both uppercase-hex and non-hash inputs failing via the same path.

nitronode/api/app_session_v1/submit_session_key_state.go
philanton: Good point. Swapped the dynamic regex string for stable wording in [59d9f15](https://github.com/layer-3/nitrolite/pull/826/changes/59d9f15b4b2b60370d0f7310caec0dfe42f5991f): "application_ids must contain only lowercase letters, digits, dashes, and underscores (max 66 chars), got: %s".

philanton:
@ihsraham good catch on the docs — went ahead and updated them in [59d9f15](https://github.com/layer-3/nitrolite/pull/826/changes/59d9f15b4b2b60370d0f7310caec0dfe42f5991f): `api.yaml` and the app_session_v1 README now spell out the exact formats (`^[a-z0-9_-]{1,66}$` and `^0x[0-9a-f]{64}$`) instead of "loose lowercase strings".

PR #827, opened 2026-06-08
MF3-H03: fix(nitronode): skip unsupported NodeBalanceUpdated tokens instead of fatal

philanton:
## Finding (MF3-H03)

`ChannelHubReactor.HandleEvent()` treated any handler error as fatal: it returned before `StoreContractEvent()`, so `Listener.processEvents()` unsubscribed and `main()` exited via `logger.Fatal()`.

`depositToNode()` accepts any ERC20 and emits `NodeBalanceUpdated`. `handleNodeBalanceUpdated()` required the token to be configured in the node asset store, so depositing an unconfigured ERC20 was a permissionless, low-cost way to stop nitronode event processing. Since the failing event was never recorded, it replayed on restart — recovery needed manual intervention or a config/code change.

## Fix

- New `core.ErrTokenNotSupported` sentinel, returned (wrapped) by the memory asset store's `GetTokenAsset`/`GetTokenDecimals` when a token is not configured.
- `handleNodeBalanceUpdated()` checks `errors.Is(err, core.ErrTokenNotSupported)`, logs a warning, skips the balance update, and returns `nil`. `HandleEvent()` then records the event via `StoreContractEvent()`, so it is not replayed, the listener stays subscribed, and the process does not die.
- Genuine store errors still propagate — only the deterministic "not configured" case is skipped (typed error, not a catch-all).

Out of scope (per finding's optional suggestion): broader listener-level error classification so handler failures don't all collapse to `logger.Fatal`. Can follow up separately.

## Tests

- New subtest `unsupported token is skipped and recorded`: asserts `HandleEvent` returns nil, `StoreContractEvent` is called, the balance handler is NOT called, and the processed callback reports success.
- `go test ./pkg/blockchain/evm/... ./nitronode/store/memory/...` and `go vet` pass.

### DIFF
--- a/nitronode/store/memory/memory_store.go
+++ b/nitronode/store/memory/memory_store.go
@@ -238,7 +238,7 @@ func (ms *MemoryStoreV1) GetTokenDecimals(blockchainID uint64, tokenAddress stri
 
 	decimalsOnChain, ok := ms.tokenDecimals[blockchainID]
 	if !ok {
-		return 0, fmt.Errorf("blockchain with ID '%d' is not supported", blockchainID)
+		return 0, fmt.Errorf("blockchain with ID '%d' has no configured tokens: %w", blockchainID, core.ErrTokenNotSupported)
 	}
 	decimals, ok := decimalsOnChain[tokenAddress]
 	if !ok {
@@ -253,7 +253,7 @@ func (ms *MemoryStoreV1) GetTokenAsset(blockchainID uint64, tokenAddress string)
 
 	assetsOnChain, ok := ms.tokenAssets[blockchainID]
 	if !ok {
-		return "", fmt.Errorf("blockchain with ID '%d' is not supported", blockchainID)
+		return "", fmt.Errorf("blockchain with ID '%d' has no configured tokens: %w", blockchainID, core.ErrTokenNotSupported)
 	}
 	asset, ok := assetsOnChain[tokenAddress]
 	if !ok {
--- a/nitronode/store/memory/memory_store.go
+++ b/nitronode/store/memory/memory_store.go
@@ -242,7 +242,7 @@ func (ms *MemoryStoreV1) GetTokenDecimals(blockchainID uint64, tokenAddress stri
 	}
 	decimals, ok := decimalsOnChain[tokenAddress]
 	if !ok {
-		return 0, fmt.Errorf("token %s is not supported on blockchain with ID '%d'", tokenAddress, blockchainID)
+		return 0, fmt.Errorf("token %s is not supported on blockchain with ID '%d': %w", tokenAddress, blockchainID, core.ErrTokenNotSupported)
 	}
 	return decimals, nil
 }
@@ -257,7 +257,7 @@ func (ms *MemoryStoreV1) GetTokenAsset(blockchainID uint64, tokenAddress string)
 	}
 	asset, ok := assetsOnChain[tokenAddress]
 	if !ok {
-		return "", fmt.Errorf("token %s is not supported on blockchain with ID '%d'", tokenAddress, blockchainID)
+		return "", fmt.Errorf("token %s is not supported on blockchain with ID '%d': %w", tokenAddress, blockchainID, core.ErrTokenNotSupported)
 	}
 	return asset.Symbol, nil
 }
--- a/pkg/blockchain/evm/channel_hub_reactor.go
+++ b/pkg/blockchain/evm/channel_hub_reactor.go
@@ -260,13 +260,7 @@ func (r *ChannelHubReactor) handleNodeBalanceUpdated(ctx context.Context, store 
 
 	asset, err := r.assetStore.GetTokenAsset(r.blockchainID, event.Token.String())
 	if err != nil {
-		if errors.Is(err, core.ErrTokenNotSupported) {
-			// A NodeBalanceUpdated event for an unconfigured token is not fatal:
-			// anyone can deposit an arbitrary ERC20 via depositToNode(). Skip the
-			// balance update but let HandleEvent record the event so it is not replayed.
-			log.FromContext(ctx).Warn("skipping NodeBalanceUpdated for unsupported token",
-				"token", event.Token.String(), "blockchainID", r.blockchainID,
-				"blockNumber", l.BlockNumber, "txHash", l.TxHash.String(), "logIndex", l.Index)
+		if r.skipIfUnsupportedToken(ctx, err, event.Token, l) {
 			return nil
 		}
 		return errors.Wrap(err, "failed to get token asset")
@@ -274,10 +268,7 @@ func (r *ChannelHubReactor) handleNodeBalanceUpdated(ctx context.Context, store 
 
 	decimals, err := r.assetStore.GetTokenDecimals(r.blockchainID, event.Token.String())
 	if err != nil {
-		if errors.Is(err, core.ErrTokenNotSupported) {
-			log.FromContext(ctx).Warn("skipping NodeBalanceUpdated for unsupported token",
-				"token", event.Token.String(), "blockchainID", r.blockchainID,
-				"blockNumber", l.BlockNumber, "txHash", l.TxHash.String(), "logIndex", l.Index)
+		if r.skipIfUnsupportedToken(ctx, err, event.Token, l) {
 			return nil
 		}
 		return errors.Wrap(err, "failed to get token decimals")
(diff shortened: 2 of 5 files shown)


pkg/blockchain/evm/channel_hub_reactor_test.go
```diff
@@ -317,6 +318,35 @@ func TestChannelHubReactor_HandleNodeBalanceUpdated(t *testing.T) {
...
+			processedSuccess = success
+		})
+
+		err := reactor.HandleEvent(context.Background(), logEntry)
+		require.NoError(t, err)
+		assert.True(t, processedSuccess)
+		// Balance update must not be applied for an unsupported token.
+		handler.AssertNotCalled(t, "HandleNodeBalanceUpdated", mock.Anything, mock.Anything, mock.Anything)
+		store.AssertExpectations(t)
+	})
```
nksazonov, line 349: Coverage gap: this subtest only exercises the early-return branch through `GetTokenAsset`. The mirror branch on `GetTokenDecimals` (reactor.go:277-281) has identical logic but no direct coverage, so a regression that breaks only one of the two guards would slip past. Suggestion: add a second subtest (or fold both into a table-driven case) where `GetTokenAsset` returns ok and `GetTokenDecimals` returns the wrapped `ErrTokenNotSupported`.

pkg/blockchain/evm/channel_hub_reactor.go
```diff
@@ -260,11 +260,26 @@ func (r *ChannelHubReactor) handleNodeBalanceUpdated(ctx context.Context, store
...
 	}
 
 	decimals, err := r.assetStore.GetTokenDecimals(r.blockchainID, event.Token.String())
 	if err != nil {
+		if errors.Is(err, core.ErrTokenNotSupported) {
+			log.FromContext(ctx).Warn("skipping NodeBalanceUpdated for unsupported token",
+				"token", event.Token.String(), "blockchainID", r.blockchainID,
+				"blockNumber", l.BlockNumber, "txHash", l.TxHash.String(), "logIndex", l.Index)
+			return nil
+		}
```
nksazonov: Minor DRY: the `errors.Is(err, core.ErrTokenNotSupported)` + warn-log + `return nil` guard is duplicated verbatim across both call sites. With only two sites this is borderline, but extracting a helper (e.g. `skipIfUnsupportedToken(ctx, err, event.Token, l) (bool)`) would prevent the two arms drifting (one branch gaining a log field the other doesn't, etc.) and shorten the handler body. Suggestion: extract to a small method on the reactor.

pkg/blockchain/evm/channel_hub_reactor.go
```diff
@@ -260,11 +260,26 @@ func (r *ChannelHubReactor) handleNodeBalanceUpdated(ctx context.Context, store
...
 	asset, err := r.assetStore.GetTokenAsset(r.blockchainID, event.Token.String())
 	if err != nil {
+		if errors.Is(err, core.ErrTokenNotSupported) {
+			// A NodeBalanceUpdated event for an unconfigured token is not fatal:
+			// anyone can deposit an arbitrary ERC20 via depositToNode(). Skip the
+			// balance update but let HandleEvent record the event so it is not replayed.
+			log.FromContext(ctx).Warn("skipping NodeBalanceUpdated for unsupported token",
+				"token", event.Token.String(), "blockchainID", r.blockchainID,
+				"blockNumber", l.BlockNumber, "txHash", l.TxHash.String(), "logIndex", l.Index)
+			return nil
```
nksazonov: Operational consequence worth one extra line in this comment block: `StoreContractEvent` is a dedup record, not a replay queue — once a `NodeBalanceUpdated` is stored as "skipped", the balance change is **not** automatically applied if the operator later configures that token. That's correct behavior for an attacker-deposited token, but an operator legitimately adding a previously-unsupported asset will need a manual resync. Suggestion: extend the existing comment block with one line spelling out the no-replay semantics so future readers don't assume re-config triggers re-processing.

nksazonov (commented):
Excellent work — the MF3-H03 fix is correct, minimal, and the new subtest locks in the right invariants (`HandleEvent` returns nil, `processedSuccess=true`, balance handler not called, `StoreContractEvent` is called). Verified locally: targeted test passes, `go vet` clean across the touched packages, CI fully green.

nitronode/store/memory/memory_store.go
```diff
@@ -257,7 +257,7 @@ func (ms *MemoryStoreV1) GetTokenAsset(blockchainID uint64, tokenAddress string)
 	}
 	asset, ok := assetsOnChain[tokenAddress]
 	if !ok {
-		return "", fmt.Errorf("token %s is not supported on blockchain with ID '%d'", tokenAddress, blockchainID)
+		return "", fmt.Errorf("token %s is not supported on blockchain with ID '%d': %w", tokenAddress, blockchainID, core.ErrTokenNotSupported)
```
ihsraham, line 260: One H-03 closure gap remains: this wraps the missing-token entry, but the missing per-chain map branch above still returns a generic `blockchain with ID ... is not supported` error. In prod, `world_chain` (`480`) is ChannelHub-enabled while the only `blockchain_id: 480` asset entry is commented out, so an arbitrary `NodeBalanceUpdated` on that chain can still miss `ErrTokenNotSupported`, bubble out of `handleNodeBalanceUpdated`, and stop the listener.

Could we make the no-token-map case return or wrap `ErrTokenNotSupported` too, or handle it explicitly in the reactor, and add a regression case for an enabled chain with no token entries?

ihsraham (changes requested):
Requesting changes for one remaining H-03 closure gap. The typed skip is the right shape for configured chains with unknown tokens, but an enabled chain with no token map still returns a generic error and can hit the old fatal listener path. I left the exact branch inline.

pkg/blockchain/evm/channel_hub_reactor_test.go
philanton: Yep, good catch — added a mirror subtest in [887b14e](https://github.com/layer-3/nitrolite/pull/827/changes/887b14e18b148f9826a745bd0a97197642330b88) that drives the skip through `GetTokenDecimals` (asset lookup ok, decimals returns the wrapped `ErrTokenNotSupported`). Both guards covered independently now.

pkg/blockchain/evm/channel_hub_reactor.go
philanton: Agreed, extracted it in [887b14e](https://github.com/layer-3/nitrolite/pull/827/changes/887b14e18b148f9826a745bd0a97197642330b88) as `skipIfUnsupportedToken(ctx, err, token, l) bool` — both arms collapse to the same call now so they can't drift.

pkg/blockchain/evm/channel_hub_reactor.go
philanton: Good point — added the no-replay note to the helper's doc comment in [887b14e](https://github.com/layer-3/nitrolite/pull/827/changes/887b14e18b148f9826a745bd0a97197642330b88): it's a dedup record, not a replay queue, so a token configured later needs a manual resync.

nitronode/store/memory/memory_store.go
philanton: Yeah, you're right — the no-token-map branch slipped through. Fixed in [887b14e](https://github.com/layer-3/nitrolite/pull/827/changes/887b14e18b148f9826a745bd0a97197642330b88): both `GetTokenAsset` and `GetTokenDecimals` now wrap `core.ErrTokenNotSupported` when the chain has no configured tokens, so an enabled chain like `world_chain` (480) skips the event instead of stopping the listener. Added `memory_store` regression cases for the enabled-chain-no-tokens path too.

ihsraham (approved):
Approved for H-03. The no-token-map branch now returns the same `ErrTokenNotSupported` sentinel as the unknown-token branch, so the `world_chain`/chain 480 case skips cleanly and still reaches `StoreContractEvent` for dedup. The added asset/decimals and reactor tests cover the closure path; no remaining blockers from me.

nksazonov (approved):
All three review comments addressed in 887b14e1 — verified locally: mirror subtest for the `GetTokenDecimals` skip branch, `skipIfUnsupportedToken` helper de-dups both guards, helper doc comment spells out the no-replay (dedup, not replay) semantics. Bonus follow-up from @ihsraham on the enabled-chain-no-tokens path (e.g. world_chain 480) also wrapped in `core.ErrTokenNotSupported` with memory_store regressions. `go test ./pkg/blockchain/evm/... ./nitronode/store/memory/...` green. LGTM.

PR #828, opened 2026-06-08
MF3-L17, I15, I16: app-session participant docs, Void-checkpoint promotion, signature canonicalization note

philanton:
Addresses three audit findings.

## MF3-L17 — single-participant app sessions (docs)
`create_app_session` docs claimed "at least 2 participants required", but `CreateAppSession()` only enforces the *max* participant count and quorum — a single-participant session with `quorum = 1` is valid. We decided to keep single-participant app sessions, so the doc was the divergence. Updated `app_session_v1/README.md` to "at least 1 participant required" to match implementation.

## MF3-I15 — Void channel left un-promoted on checkpoint (fix)
`ChannelHub.createChannel()` can emit `ChannelCheckpointed` before `ChannelCreated` for an initial non-deposit/non-withdraw state. If the checkpoint was processed while the local row was still the `Void` seed, `HandleHomeChannelCheckpointed()` bumped `state_version` but left the channel `Void`, violating the protocol invariant until the later `ChannelCreated` event replayed and repaired it.

Fix: treat a checkpoint on a `Void` home channel as evidence it is materialized on-chain and promote it to `Open`, backfilling any unsigned concurrent receiver head (mirrors `HandleHomeChannelCreated`). The later `ChannelCreated` then no-ops via the `Status >= Open` guard. Added a regression test.

## MF3-I16 — stored signatures not canonicalized (docs)
`channel_states.user_sig` / `node_sig` are unbounded `text` and accepted signatures are persisted verbatim rather than re-encoded after verification. They may retain non-canonical / unused trailing bytes (notably around session-key payloads, and for signatures learned from on-chain events that bypass the WebSocket message-size limit). No asset-safety impact — verification still gates acceptance. Documented under Known Limitations so consumers don't assume stored signature bytes are minimal/canonical.


nitronode/event_handlers/service.go
```diff
@@ -161,6 +161,18 @@ func (s *EventHandlerService) HandleHomeChannelCheckpointed(ctx context.Context,
...
+	// sufficient evidence that the channel has been materialized: promote Void to Open
+	// here instead of leaving a bumped state_version on a Void channel until the later
+	// ChannelCreated event replays. That replay then no-ops via the Status >= Open guard
+	// in HandleHomeChannelCreated.
+	wasVoid := channel.Status == core.ChannelStatusVoid
+	if wasVoid {
+		channel.Status = core.ChannelStatusOpen
+	}
+
 	wasChallenged := channel.Status == core.ChannelStatusChallenged
```
nksazonov: `wasChallenged` is computed after `channel.Status` may already have been flipped to `ChannelStatusOpen` by the `wasVoid` branch directly above. This is currently safe because `Void` and `Challenged` are mutually exclusive, but the dependency on that invariant isn't local to either line — any future promotion path that goes `Void → Challenged`, or a refactor that moves the assignment, would silently make the `wasChallenged` branch unreachable.

Suggestion: snapshot the original status once at the top and derive both flags from it, so neither depends on the other's mutation order.

nitronode/event_handlers/service_test.go
```diff
@@ -381,6 +381,53 @@ func TestHandleHomeChannelCheckpointed_Success(t *testing.T) {
...
+		return ch.ChannelID == channelID &&
+			ch.Status == core.ChannelStatusOpen &&
+			ch.StateVersion == 1
+	})).Return(nil)
+	mockStore.On("RefreshUserEnforcedBalance", userWallet, "usdc").Return(nil)
+	mockStore.On("UpdateStateSigsIfMissing", channelID, uint64(1), "", "").Return(nil)
+	// Void→Open promotion runs the head-sig backfill (mirrors HandleHomeChannelCreated);
+	// no off-chain head present → no-op. HasSignedFinalize is only consulted on the
+	// Challenged path and must not be reached here.
+	mockStore.On("GetLastStateByChannelID", channelID, false).Return(nil, nil)
```
nksazonov, line 415: The new branch added at `service.go:218` (`if wasChallenged || wasVoid`) is only verified for the no-head case (mock returns `nil`). The interesting path — the "concurrent RPC stored an unsigned `transfer_receive`/`release` head while the channel was still `Void`" scenario explicitly called out in the production code comment at `service.go:213-218` — has no regression test. A future change that drops the `|| wasVoid` clause would still pass CI.

Suggestion: add the analog of `TestHandleHomeChannelCreated_BackfillsUnsignedReceiverStateOnOpen` (`service_test.go:173-264`) for the Checkpointed path — seed a `Void` channel plus an unsigned `transfer_receive` (and `release`) head at a version above the checkpoint version, drive `HandleHomeChannelCheckpointed`, and assert the head row is node-signed with a signature that verifies against the node key.

nksazonov (approved):
Good job — the three audit findings are addressed correctly and the fix logic is sound; two small things to tighten before landing.

ihsraham (approved):
Approving for L-17 / I-15 / I-16 closure. The docs now match the intended single-participant app-session behavior, the checkpoint path promotes only Void home channels to Open, and the stored-signature limitation is documented clearly.

Agree to Nikita's existing non-blocking notes. One small docs cleanup remains: the create-app-session README sample now shows one participant but still has two signature placeholders. That does not reopen L-17, but it is worth tightening before merge.

nitronode/event_handlers/service.go
philanton: Yeah, good catch — that cross-line dependency was too implicit. Snapshotted the status once at the top now and derive both `wasVoid` and `wasChallenged` from it, so neither branch leans on the other's mutation order. Fixed in [2dad3a2](https://github.com/layer-3/nitrolite/pull/828/changes/2dad3a202d4f4912216f827551a457ebdddaf157)

nitronode/event_handlers/service_test.go
philanton: Yep, the `|| wasVoid` arm was only exercised for the no-head case. Added `TestHandleHomeChannelCheckpointed_FromVoidBackfillsUnsignedReceiverHead` — seeds a Void channel plus an unsigned `transfer_receive`/`release` head above the checkpoint version, drives `HandleHomeChannelCheckpointed`, and asserts the head row is node-signed with a sig that verifies against the node key. Dropping the clause fails it now. Added in [2dad3a2](https://github.com/layer-3/nitrolite/pull/828/changes/2dad3a202d4f4912216f827551a457ebdddaf157)

PR #829, opened 2026-06-08
MF3-L16: fix(rpc): escape error message in NewErrorPayload

philanton:
### Problem
`NewErrorPayload()` built the error field by wrapping `errMsg` with `fmt.Sprintf(`"%s"`, errMsg)` and storing it as `json.RawMessage`. Since `json.RawMessage` is treated as already-encoded JSON, embedded quotes or other JSON-special characters in `errMsg` were not escaped during the later websocket response marshal.

A client-facing `rpc.Error` containing quotes — e.g. a duplicate-key error mentioning `"app_sessions_v1_pkey"` — produced invalid raw JSON. The websocket node then failed to marshal `ctx.Response` and returned the generic node error instead of the intended application-level error response.

### Fix
Encode `errMsg` with `json.Marshal()` in `NewErrorPayload()` and store the returned bytes. The standard encoder escapes quotes, backslashes, and control characters before the value is embedded in the final response. A defensive fallback to `""` is kept (marshaling a Go string cannot fail in practice).

### Tests
Added `pkg/rpc/error_test.go` covering embedded quotes, backslashes, newlines, control chars, leading quotes, and empty input. Each case asserts the stored value round-trips to the original string and that the whole payload marshals without error (regression guard for the websocket-marshal failure).

### DIFF
--- a/pkg/rpc/error.go
+++ b/pkg/rpc/error.go
@@ -122,5 +122,13 @@ func (e Error) Error() string {
 //
 // The resulting Params will contain: {"error": "invalid address format"}
 func NewErrorPayload(errMsg string) Payload {
-	return Payload{errorParamKey: json.RawMessage(fmt.Sprintf(`"%s"`, errMsg))}
+	// json.Marshal escapes quotes, backslashes, and control characters so the
+	// resulting RawMessage is always valid JSON. Building the string by hand
+	// would corrupt the payload whenever errMsg contains JSON-special chars.
+	encoded, err := json.Marshal(errMsg)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


nksazonov (approved):
Good work!

ihsraham (approved):
Approving for L-16. The change moves `NewErrorPayload` from hand-built JSON to `json.Marshal(errMsg)`, which closes the quoted-error failure mode without changing client error semantics.

I also checked the full `NewErrorResponse` path with a quoted duplicate-key style message: the same regression fails on the base with the invalid `json.RawMessage` marshal error and passes on this head. The committed test covers the important escaping cases at the payload level. A direct full-message regression would be a nice extra guard, but I would not block on it.

PR #830, opened 2026-06-08
MF3-L18: fix(nitronode): lock transfer balances in deterministic order

philanton:
## What

Fixes the MF3-L18 audit finding: balance-lock acquisition for `transfer_send` transitions happened in request-dependent order, so two opposite-direction transfers of the same asset could deadlock at the database layer.

`SubmitState` and `RequestCreation` locked the sender's `user_balances` row first, then `issueTransferReceiverState` locked the receiver's row later in the same tx. Concurrent A→B and B→A transfers could lock A-then-wait-B and B-then-wait-A — a cycle that Postgres breaks with `SQLSTATE 40P01`, aborting a valid transfer instead of resolving it deterministically. No balance corruption (the tx rolls back), but it's a liveness/UX problem under load.

## Change

- New `lockTransferBalances` helper: normalizes the receiver address, rejects self-transfers, and locks both `(wallet, asset)` rows in ascending lowercase-wallet order **up front** for any `TransferSend` transition. Same sequence regardless of direction → no lock cycle.
- Both handlers branch: `TransferSend` → `lockTransferBalances`; every other transition keeps the existing sender-only lock.
- `issueTransferReceiverState` no longer takes the receiver lock — it now documents that the caller must pre-lock the receiver row.

Mirrors the deterministic-ordering pattern already used in `app_session_v1` (sorted participants before locking).

### Side effect (improvement)

Self-transfer and invalid-receiver are now rejected *before* the sender lock and state store, rather than after. Less wasted work; behavior otherwise unchanged.

## Scope

Cross-subsystem ordering (channel transfers vs `app_session_v1` release locks) is a separate concern and intentionally out of scope here — different assets/rows.

## Tests

- New `lock_transfer_balances_test.go`: asserts both directions lock low→high, and that self-transfer / invalid-receiver take zero locks.
- Updated the self-transfer case in `submit_state_test.go` to reflect the earlier rejection point.
- `go build ./nitronode/...`, `go vet`, full `channel_v1` suite green.

### DIFF
--- a/nitronode/api/channel_v1/handler.go
+++ b/nitronode/api/channel_v1/handler.go
@@ -97,22 +97,16 @@ func (h *Handler) lockTransferBalances(tx Store, senderState core.State) (string
 // It reads the receiver's current state, applies a transfer_receive transition with the same
 // amount and tx hash, signs it with the node's key, and persists it.
 //
-// The receiver's (wallet, asset) balance row MUST already be locked by the caller
-// (via lockTransferBalances) before this is invoked — it does not take the lock itself.
-func (h *Handler) issueTransferReceiverState(ctx context.Context, tx Store, senderState core.State, applicationID string) (*core.State, error) {
+// receiverWallet must be the normalized receiver address returned by
+// lockTransferBalances, which the caller MUST invoke first — that call also
+// locks the receiver's (wallet, asset) row and rejects self-transfers, so this
+// function does neither itself.
+func (h *Handler) issueTransferReceiverState(ctx context.Context, tx Store, senderState core.State, receiverWallet, applicationID string) (*core.State, error) {
 	logger := log.FromContext(ctx)
 
 	incomingTransition := senderState.Transition
 	if incomingTransition.Type != core.TransitionTypeTransferSend {
 		return nil, rpc.Errorf("incoming state doesn't have 'transfer_send' transition")
-	}
-	receiverWallet, err := core.NormalizeHexAddress(incomingTransition.AccountID)
-	if err != nil {
-		return nil, rpc.Errorf("invalid receiver wallet address: %v", err)
-	}
-
-	if strings.EqualFold(senderState.UserWallet, receiverWallet) {
-		return nil, rpc.Errorf("sender and receiver wallets are the same")
 	}
 
 	logger = logger.
--- a/nitronode/api/channel_v1/handler.go
+++ b/nitronode/api/channel_v1/handler.go
@@ -59,9 +59,46 @@ func (h *Handler) getChannelSigValidator(tx Store, asset string) *core.ChannelSi
 	})
 }
 
+// lockTransferBalances normalizes the receiver address, rejects self-transfers,
+// and locks the sender's and receiver's (wallet, asset) balance rows in a
+// deterministic order (ascending lowercase wallet) to avoid database deadlocks
+// when two opposite-direction transfers run concurrently. It returns the
+// normalized receiver wallet.
+//
+// Callers MUST call this before issueTransferReceiverState for a TransferSend
+// transition — issueTransferReceiverState assumes the receiver row is already
+// locked and does not take the lock itself.
+func (h *Handler) lockTransferBalances(tx Store, senderState core.State) (string, error) {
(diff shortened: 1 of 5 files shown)


nitronode/api/channel_v1/handler.go
```diff
@@ -59,9 +59,46 @@ func (h *Handler) getChannelSigValidator(tx Store, asset string) *core.ChannelSi
...
+	if strings.ToLower(first) > strings.ToLower(second) {
+		first, second = second, first
+	}
+	if _, err := tx.LockUserState(first, senderState.Asset); err != nil {
+		return "", rpc.Errorf("failed to lock user state: %v", err)
+	}
+	if _, err := tx.LockUserState(second, senderState.Asset); err != nil {
+		return "", rpc.Errorf("failed to lock user state: %v", err)
+	}
+	return receiverWallet, nil
```
nksazonov, line 93: The returned `receiverWallet` is discarded by both callers (`submit_state.go:45` and `request_creation.go:89`) with `_`. `issueTransferReceiverState` then independently re-normalizes the same `senderState.Transition.AccountID`, making the return value redundant. Either wire the returned wallet into `issueTransferReceiverState` to eliminate the duplicate normalization, or change the signature to return `error` only.

nksazonov (commented):
Excellent implementation. The fix correctly addresses the MF3-L18 audit finding: both balance rows are now locked at the top of the transaction in deterministic ascending-wallet order, validation and self-transfer rejection happen before any lock is taken, and the receiver lock has been cleanly removed from `issueTransferReceiverState`. The tests cover both transfer directions and the two rejection cases.

---

**Out-of-diff notes:**

**`nitronode/api/channel_v1/handler.go:114-116` (unchanged line in `issueTransferReceiverState`):** The `strings.EqualFold` self-transfer check here is now dead code — `lockTransferBalances` (which all `TransferSend` callers must invoke first per the documented contract) already rejects self-transfers and returns an error before reaching this function. Not a bug, but worth removing to avoid confusion about when the invariant is actually enforced.

**`nitronode/api/app_session_v1/handler.go:125` (out of scope for this PR):** `issueReleaseReceiverState` still acquires `LockUserState` inside itself, the same structural pattern the audit flagged for `issueTransferReceiverState`. If a multi-participant session close runs concurrently with another session close or transfer that shares participants, the non-deterministic iteration order of `currentAllocations` could produce a deadlock of the same class. Outside the scope of this audit finding — flagging for tracking.

ihsraham (approved):
Approved for L-18 closure. I verified the lock ordering end to end: both transfer entry points call `lockTransferBalances` before receiver issuance, and the helper's lowercase ordering matches the `LockUserState` Postgres `FOR UPDATE` key.

No blockers from me. The cleanup and follow-up notes I would have raised are already covered by @nksazonov, so I'm not adding duplicate inline comments.

nitronode/api/channel_v1/handler.go
philanton: Yeah, good catch. Wired the normalized wallet through instead — `lockTransferBalances` now returns `receiverWallet` and the callers feed it into `issueTransferReceiverState`, which drops its own `NormalizeHexAddress` + self-transfer check. Single source of truth now. Fixed in [dd92882](https://github.com/layer-3/nitrolite/pull/830/changes/dd928223cc4807bb3cd862f696b6a075e1fec0e9)

philanton:
@nksazonov thanks, addressed both out-of-diff notes:

1) `handler.go:114-116` self-transfer check — removed it together with the duplicate normalization in the same [dd92882](https://github.com/layer-3/nitrolite/pull/830/changes/dd928223cc4807bb3cd862f696b6a075e1fec0e9) commit, so `lockTransferBalances` is the only place that rejects self-transfers now;
2) `app_session_v1/handler.go:125` `issueReleaseReceiverState` — yeah, same class of pattern, but I'd keep it out of this PR. It's a separate finding from L-18 and the fix lives on the session-close path, not transfers.

nksazonov (approved):
All review comments addressed in dd92882. `lockTransferBalances` is now the single source of truth for receiver address normalization, self-transfer rejection, and deterministic lock ordering. LGTM.

PR #831, opened 2026-06-08
MF3-L03: docs: document app session close atomicity blocking on in-flight escrow

nksazonov:
Documents the audit finding (MF3-L03) that app session cooperative close is atomic across all participants and aborts if any participant has an in-flight escrow operation (escrow lock, mutual lock, escrow deposit/withdrawal awaiting finalization).

- `contracts/SECURITY.md` — new Behavior rule 8 with full mechanism, safety rationale, recovery paths, and invariant.
- `protocol-description.md` — short note in the "Off-chain application sessions" section, linking to SECURITY.md rule 8.
- `docs/protocol/security-and-limitations.md` — bullet in "Known Limitations" matching the existing voice.

No code changes.


philanton (approved):

contracts/SECURITY.md
```diff
@@ -51,6 +51,30 @@ Invariant:
...
 > Credits accrued while a channel was in `CHALLENGED` status are preserved — either reintegrated into the channel on challenge clearance, or carried into the next epoch on closure — and cannot be shadowed by a concurrently-opened replacement channel.
 
+---
+
+8. App session closure is **atomic across all participants**. The Node issues a new release receive-state on every participant's home channel as part of a single close transaction; a failure on any single participant aborts the entire close, leaving the session open and no release credits issued to anyone.
+
+The Node refuses to issue a release state when the recipient has an **in-flight escrow operation on the affected asset**, including:
+
+* an escrow lock or mutual lock that has not yet settled,
+* an escrow deposit whose finalization has not yet been confirmed on-chain,
```
ihsraham: This is a bit broader than the current gate. `escrow_lock` and `mutual_lock` always block, and escrow withdrawals block until the escrow channel version catches up, but escrow deposits have allowed cases: caught-up, plus the one-behind `Open`/`Closed` states used around finalize/purge.

Could we spell this as “unresolved escrow deposit states that the node still treats as unsafe,” or otherwise mirror `EnsureNoOngoingEscrowOperation()`? Since this docs text is the remediation for L-03, over-describing the block makes the documented behavior drift from the code.

contracts/SECURITY.md
```diff
@@ -51,6 +51,30 @@ Invariant:
...
+* if the session's state machine permits intermediate updates, individually unwind their share via off-chain transfers out of the session and re-attempt closure with the remaining (non-blocked) members.
+
+This is an accepted trade-off for now, which may be lifted in the future: protocol safety (no release state co-signed against a potentially-reverting escrow chain) takes precedence over close-time liveness. Sessions whose participants require independent exit guarantees should be designed with state machines that support partial settlement rather than relying solely on cooperative close.
+
+Unlike the `CHALLENGED` channel path (rule 6), which queues incoming receive-states as unsigned entries so the channel's state history can carry the credit forward, the close path does not currently fall back to issuing an unsigned release. Extending that pattern here is a possible future improvement but is non-trivial: the safety of stacking an unsigned release on an unfinalized escrow depends on later reconciling the eventual escrow outcome with the queued credit, and the current implementation prefers refusal over partial trust.
```
ihsraham: This needs one qualification. The release issuer can still store an unsigned release for a non-open home channel after the escrow predicate passes. What it does not do is create an unsigned release fallback when `EnsureNoOngoingEscrowOperation()` rejects the participant.

I would narrow this paragraph to that case, so it does not contradict the challenged-channel rescue path.

docs/protocol/security-and-limitations.md
```diff
@@ -81,6 +81,7 @@ The following capabilities are not yet implemented or have acknowledged design t
...
+- App session cooperative closure is atomic across all participants. The Node refuses to issue a release receive-state to any participant with an in-flight escrow operation on the affected asset (escrow lock, mutual lock, or escrow deposit/withdrawal awaiting on-chain finalization), because stacking a co-signed release on an unfinalized escrow risks state-chain invariant violations if the escrow ultimately reverts or settles to an unexpected version. As a consequence, a single participant with a pending escrow blocks cooperative close for all others in the session until their escrow resolves. Affected participants may wait for the obstruction to clear, or — where the session state machine permits intermediate updates — unwind their share individually via off-chain transfers out of the session and re-close without the blocked participant. See [`/contracts/SECURITY.md`](../../contracts/SECURITY.md) Behavior rule 8 for the full rationale.
```
ihsraham: This link works in the repo, but it is likely broken in the packaged MCP content. `sdk/mcp/scripts/prepare-package-content.mjs` copies `docs/protocol` into `content`, but not `contracts/SECURITY.md`, while this page is served as the MCP security overview.

Could we avoid the out-of-package relative link, use a stable repo URL, or include `contracts/SECURITY.md` in the packaged content?

ihsraham (changes requested):
This is the right shape for L-03: documenting the cooperative-close liveness tradeoff is an acceptable remediation path, and the main atomic-close behavior matches the current transaction flow.

I would revise a few details before closure, though. Since this PR is the remediation, the docs need to match the release/escrow code path closely, and the public security doc should not point at a file that is missing from the packaged MCP content.

contracts/SECURITY.md
nksazonov: Replaced the bullet list to mirror `EnsureNoOngoingEscrowOperation`'s actual semantics, including the one-version-behind allowance for `escrow_deposit` in `Open`/`Closed` status. Propagated to `protocol-description.md` and `docs/protocol/security-and-limitations.md`. Fixed in 3f7b89b6.

contracts/SECURITY.md
nksazonov: Narrowed the paragraph to the escrow-rejected case and now acknowledges that the release issuer does store unsigned releases for non-`Open` home channels (the path that underwrites rule 6). The "no fallback" claim now applies specifically to the `EnsureNoOngoingEscrowOperation`-rejected branch. Fixed in 3f7b89b6.

docs/protocol/security-and-limitations.md
nksazonov: Took a fourth route: `contracts/SECURITY.md` is internal dev-oriented documentation, so we've decided `docs/` should not reference it at all. The trailing pointer is dropped and replaced with a self-contained closing sentence carrying the rationale. Fixed in 43b28ff4.

docs/protocol/security-and-limitations.md
```diff
@@ -81,6 +81,7 @@ The following capabilities are not yet implemented or have acknowledged design t
...
+- App session cooperative closure is atomic across all participants. The Node refuses to issue a release receive-state to any participant whose latest signed state encodes an escrow operation that the off-chain gate does not yet treat as safely settled — covering any pending `escrow_lock`/`mutual_lock`, plus `escrow_deposit`/`escrow_withdraw` whose on-chain escrow-channel version has not caught up. Stacking a co-signed release on an unfinalized escrow risks state-chain invariant violations if the escrow ultimately reverts or settles to an unexpected version. As a consequence, a single participant with a pending escrow blocks cooperative close for all others in the session until their escrow resolves. Affected participants may wait for the obstruction to clear, or — where the session state machine permits intermediate updates — unwind their share individually via off-chain transfers out of the session and re-close without the blocked participant. This is an accepted trade-off favouring protocol safety over close-time liveness: every release the Node co-signs must remain enforceable on-chain, and an unfinalized escrow cannot offer that guarantee.
```
ihsraham: Non-blocking, but worth tightening: the full rule in `contracts/SECURITY.md` includes the one-version-behind `escrow_deposit` allowance for `Open` or `Closed` escrow channels, while this shorter summary still reads like any not-caught-up `escrow_deposit` blocks.

Could we either mirror that allowance here too, or phrase this as “escrow deposit states the gate still treats as unsafe”? That keeps the packaged/public docs aligned with `EnsureNoOngoingEscrowOperation`.

ihsraham (approved):
Approving for L-03 closure. The docs now make the cooperative-close liveness tradeoff explicit: app-session close is all-participant atomic, a participant rejected by the escrow gate can block everyone, and the recovery paths are described. The previous `contracts/SECURITY.md` wording and packaged MCP link concerns look addressed.

I left one non-blocking precision note on the public summary. It does not reopen L-03 for me, but it would keep the packaged docs closer to `EnsureNoOngoingEscrowOperation`.

docs/protocol/security-and-limitations.md
nksazonov: Tightened both `docs/protocol/security-and-limitations.md` and `protocol-description.md` to mirror the gate semantics — now mentions the narrow one-version-behind allowance for `escrow_deposit` during normal finalize/purge transitions. Fixed in df445097.

PR #832, opened 2026-06-08
MF3-L15: fix(nitronode): fix reorg double-spend via confirmation gate

nksazonov:
## Summary

Implements the reorg double-spend fix specified in **[`nitronode/reorg-fix-spec.md`](../blob/fix/reorgs/nitronode/reorg-fix-spec.md)** (the canonical implementation spec for this PR — start there for design rationale, the residual-finality trade-off, and the per-component contracts).

The Nitronode previously credited a user's off-chain balance the instant it observed a deposit event on-chain. A subsequent reorg would remove the on-chain deposit while the off-chain credit persisted, allowing an attacker to drain that credit to a receiver before the node could detect the reorg. This PR introduces a per-chain confirmation window, a startup reconciliation walk over stored block hashes, and a reactor-level idempotency pre-check that together close the window and make replay safe.

## Spec at a glance

| Section | Component |
| --- | --- |
| §2 Solution overview, §2.1 Residual risk | Why a delay, what it does and doesn't guarantee |
| §3 Configuration | New `confirmation_delay_secs` per chain in `blockchains.yaml` |
| §4.1–4.3 Window behavior | Queue, cancel on `Removed:true`, replace-on-re-add |
| §4.4 Startup reconciliation | `block_hash` column + common-ancestor walk via `eth_getBlockByHash` |
| §6.1–6.3 Component placement, queue keying, two-goroutine design | `ConfirmationGate` design |
| §6.4 API exposure | `confirmation_delay_secs` on `node.v1.GetConfig` |
| §6.5 Post-gate reorg detection | `recentlyForwarded` map, WARN signal |
| §6.6 Reactor pre-check | `IsContractEventProcessed` skips already-committed events |
| §6.7 Block timestamp cache | `arrivedAt` from block time so historical replay matures immediately |

### On-chain event pipeline (`pkg/blockchain/evm/`)
- **`ConfirmationGate`** (new): in-memory buffer between the Listener and reactor. Holds events for `confirmation_delay_secs` keyed by block timestamp; cancels on `Removed:true` arrivals; logs WARN on post-gate reorgs.
- **`Listener`**: now accepts two handlers — `eventHandler` (live, through gate) and `historicalEventHandler` (Phase 1 replay, direct to reactor). Historical events bypass the gate per spec §4.4 step 5, avoiding spurious `HeaderByHash` calls during startup. `Removed:true` logs are forwarded to the live handler (the gate) instead of being silently dropped, so the gate can cancel pending entries.
- **`ChannelHubReactor`**: pre-check via new `IsContractEventProcessed(txHash, logIndex, blockchainID)` (§6.6) — converts what was a constraint-violation rollback on the reconciliation replay path into a clean INFO exit.
- **`findCommonAncestor`** (new in `reconciler.go`): startup walk over stored `block_hash` values, calling `eth_getBlockByHash` until it finds a still-canonical block; treats pre-migration empty hashes as canonical fallback.

### Storage (`nitronode/store/database/`, `pkg/core/`)
- New migration `20260608000000_add_block_hash_to_contract_events.sql` — adds `CHAR(66) NOT NULL DEFAULT ''` column.
- `core.BlockchainEvent` gains `BlockHash`; reactors populate it on every store.
- New store methods: `IsContractEventProcessed`, `GetLatestContractEventBlockHashAndNumber`, `GetPreviousDistinctBlockHash`.

### Config & API surface
- `core.Blockchain.ConfirmationDelaySecs` (`uint32`) flows from `blockchains.yaml` → memory store → `node.v1.GetConfig` → `BlockchainInfoV1`.
- Go SDK (`sdk/go`) and TS SDK (`sdk/ts`) updated to expose the field; SDK tests updated.

### Wiring (`nitronode/main.go`)
- ChannelHub path: `Listener → ConfirmationGate → reactor` for live events; `Listener → reactor` directly for Phase 1.
- LockingContract path: unchanged in behavior (no gate); same handler passed for both phases for compile parity.

### Tests
- `confirmation_gate_test.go` — 13 tests (T1–T13) covering delay=0 transparent path, normal queue/forward, pre-gate cancel, out-of-order re-delivery, post-gate reorg WARN, unknown removal, block-timestamp bypass, partial delay, fetcher error fallback, cache dedup, shutdown, eviction, ordering.
- `reconciler_test.go` — 7 tests covering canonical, single reorg, walk-to-genesis, pre-migration latest row, pre-migration mid-walk, RPC error, no-stored-events.
- `channel_hub_reactor_test.go` — `TestChannelHubReactor_HandleEvent_AlreadyProcessed` and `TestChannelHubReactor_HandleEvent_PreCheckError` cover the §6.6 pre-check.
- `listener_test.go` — `TestListener_RemovedLog_ForwardedToHandler` verifies Removed-log routing; `TestListener_PhaseHandlerRouting` verifies Phase 1 / Phase 2 handler split.

## Notes for reviewers

- The spec text and code were reconciled in commit `44efbe45` — anywhere they currently appear to diverge, the spec is the intended source of truth and the code should match it (or both should be updated together).
- §6.6 documents a known limitation: the idempotency pre-check uses block-level `log_index` which is not stable across reorgs that change the tx's position within a block. Business-logic idempotency in the existing handlers covers this case; tx-relative `log_index` is filed as a follow-up task in the spec.

## Release Notes

* **New Features**
  * Added configurable event confirmation delay per blockchain to control event processing timing
  * Enhanced blockchain reorganization protection through block hash tracking in stored events

* **Documentation**
  * Added detailed reorg-fix specification document

* **Tests**
  * Added comprehensive test coverage for event confirmation and blockchain reorganization scenarios


nitronode/store/memory/blockchain_config.go
```diff
@@ -43,6 +43,10 @@ type BlockchainConfig struct {
 	ChannelHubAddress string `yaml:"channel_hub_address"`
 	// ChannelHubSigValidators maps validator IDs to the addresses of signature validators for the ChannelHub contract on this blockchain
 	ChannelHubSigValidators map[uint8]string `yaml:"channel_hub_sig_validators"`
+	// ConfirmationDelaySecs is the number of seconds to wait before processing an event.
```
ihsraham, line 46: This default leaves the mitigation off unless every deployed config is updated. `ConfirmationDelaySecs` stays at `0` when the YAML key is omitted, and `0` makes the gate transparent. The checked-in prod/sandbox/stress `blockchains.yaml` files do not set `confirmation_delay_secs`, so the deployed node would still process live deposits immediately.

For L-15 closure, please make the confirmation delay non-zero in shipped configs or enforce a safe default/validation for enabled EVM chains.

nitronode/main.go
```diff
@@ -119,9 +121,26 @@ func main() {
...
+				return time.Unix(int64(header.Time), 0), nil
+			}
+
+			confirmationDelay := time.Duration(b.ConfirmationDelaySecs) * time.Second
+			gate := evm.NewConfirmationGate(confirmationDelay, b.ID, reactor.HandleEvent, blockTimestampFetcher, logger)
+			gate.Start(blockchainCtx)
+
+			// Live events flow through the confirmation gate; historical events from eth_getLogs
+			// are already canonical and go directly to the reactor (see reorg-fix-spec.md §4.4 step 5).
+			l := evm.NewListener(common.HexToAddress(b.ChannelHubAddress), client, b.ID, b.BlockStep, logger, gate.HandleEvent, reactor.HandleEvent, bb.DbStore)
```
ihsraham: Routing historical replay directly to the reactor can bypass the new confirmation window after a restart or subscription flap. The live listener advances `lastBlock` when it queues an event, before the gate has committed it downstream. If the process reconnects inside the delay window, replay can treat that block as historical and send the same not-yet-confirmed event straight to the reactor.

Please either persist pending confirmation state, replay unconfirmed ranges through the gate, or otherwise prove that replay cannot include events still inside the configured confirmation window.

pkg/blockchain/evm/reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+		return blockNum, nil
+	}
+
+	for {
+		if ctx.Err() != nil {
+			return 0, ctx.Err()
+		}
+
+		hash := common.HexToHash(blockHash)
+		header, err := client.HeaderByHash(ctx, hash)
```
ihsraham: `HeaderByHash` does not prove this stored hash is canonical. Depending on the backend, a reorged hash may come back as `ethereum.NotFound`, which this currently treats as a fatal startup error, or it may return a known side-chain header by hash. In the second case this accepts an orphaned block as the common ancestor.

For the restart/reorg closure path, please compare the stored hash with the canonical header/hash at that block number and walk back on `NotFound` or mismatch.

pkg/blockchain/evm/listener.go
```diff
@@ -106,9 +114,9 @@ func (l *Listener) logBackOff(count uint64, originator string) (time.Duration, b
 // On subscription failure it retries with exponential backoff. Returns non-nil only
 // when the handler or the event-presence check fails.
 func (l *Listener) listenEvents(ctx context.Context) error {
-	lastBlock, err := l.eventGetter.GetLatestContractEventBlockNumber(l.contractAddress.String(), l.blockchainID)
+	lastBlock, err := findCommonAncestor(ctx, l.client, l.eventGetter, l.contractAddress.String(), l.blockchainID, l.logger)
```
ihsraham: This new return value is later treated the same as an empty store, but `findCommonAncestor` also uses `0` when every stored block was reorged out. In that full-depth reorg case the listener skips historical replay entirely, leaving stale event-derived DB state in place instead of rebuilding from a canonical range.

Please distinguish the empty-store case from the all-stored-blocks-reorged case, or replay from the configured start/genesis boundary when reconciliation returns this because no stored block is still canonical.

pkg/blockchain/evm/confirmation_gate.go
```diff
@@ -0,0 +1,278 @@
...
+					txHash:    entry.log.TxHash,
+					blockHash: entry.log.BlockHash,
+					logIndex:  uint(entry.log.Index),
+				}
+				g.recentlyForwarded[fk] = now
+
+				g.mu.Unlock()
+
+				evCtx := log.SetContextLogger(context.Background(), g.logger)
+				if err := g.handler(evCtx, entry.log); err != nil {
```
ihsraham: With a non-zero delay, handler failures become best-effort here. The poller removes the event from the queue before calling the reactor, logs the error, and keeps going; the listener never sees the failure, unsubscribes, or retries. That weakens the old listener invariant that transient handler failures retry instead of silently dropping an on-chain event.

Please keep confirmed events retryable, or surface the failure back into the listener lifecycle so deposits/withdrawals/challenges are not lost until an external restart happens.

ihsraham (changes requested):
Thanks for putting the confirmation-gate path together. The uninterrupted live path is moving in the right direction: Removed logs now reach the gate, pending logs can be cancelled before processing, and the tests cover the basic pre-gate cancellation cases.

I would not close L-15 on this version yet. A few paths still either bypass the confirmation window or weaken the old listener retry behavior, so a reorg/restart or transient reactor failure can still leave the node in the wrong state. I left the closure blockers inline.

nksazonov:
@coderabbitai review

pkg/blockchain/evm/reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return 0, fmt.Errorf("check canonicality of block %d (%s): %w", blockNum, blockHash, err)
+		}
+
+		if header != nil {
+			// This block is still in the canonical chain.
+			if blockNum != header.Number.Uint64() {
+				// Sanity check: the block at this hash should have the number we stored.
+				return 0, fmt.Errorf("block hash %s has unexpected number: stored %d, chain %d", blockHash, blockNum, header.Number.Uint64())
```
philanton: What happens next?

pkg/blockchain/evm/reconciler.go
philanton, line 1: it's better to document it more, not enough information to understand the purpose fully

pkg/blockchain/evm/listener.go
```diff
@@ -263,7 +274,7 @@ func (l *Listener) processEvents(
 			}
 			l.logger.Debug("received historical event", "blockchainID", l.blockchainID, "contractAddress", l.contractAddress.String(), "blockNumber", eventLog.BlockNumber, "logIndex", eventLog.Index)
 			evCtx := log.SetContextLogger(context.Background(), l.logger)
-			if err := l.handleEvent(evCtx, eventLog); err != nil {
+			if err := l.handleHistoricalEvent(evCtx, eventLog); err != nil {
```
philanton: I don't understand the purpose of separation, events received from reconcile should be directed to confirmation gate as well

nitronode/store/database/contract_event.go
philanton, line 1: I don't like the idea of having both `IsContractEventProcessed` and `IsContractEventPresent` that are different only by the blockNumber filter. And not only having both of these functions, I don't understand why we need to run pretty much the same query twice for each event.

nitronode/main.go
```diff
@@ -119,9 +121,26 @@ func main() {
 				return wrapInTx(func(s database.DatabaseStore) error { return h(s) })
 			}
 
-			reactor := evm.NewChannelHubReactor(b.ID, bb.StateSigner.PublicKey().Address().String(), eventHandlerService, bb.MemoryStore, useCHRStoreInTx)
+			reactor := evm.NewChannelHubReactor(b.ID, bb.StateSigner.PublicKey().Address().String(), eventHandlerService, bb.MemoryStore, useCHRStoreInTx, bb.DbStore)
 			reactor.SetOnEventProcessed(bb.RuntimeMetrics.IncBlockchainEvent)
-			l := evm.NewListener(common.HexToAddress(b.ChannelHubAddress), client, b.ID, b.BlockStep, logger, reactor.HandleEvent, bb.DbStore)
+
+			blockTimestampFetcher := func(blockHash common.Hash) (time.Time, error) {
```
philanton: This is too specific to blockchain, it shouldn't be declared here

nitronode/docs/reorg-fix.md
philanton, line 1: Move into better document before merge

pkg/blockchain/evm/channel_hub_reactor.go
```diff
@@ -178,7 +186,25 @@ func (r *ChannelHubReactor) HandleEvent(ctx context.Context, l types.Log) error
 	}
 	logger.Debug("received event", "name", eventName, "blockNumber", l.BlockNumber, "txHash", l.TxHash.String(), "logIndex", l.Index)
 
-	err := r.useStoreInTx(func(store ChannelHubReactorStore) error {
+	// Pre-check: skip already-committed events without opening a transaction.
+	// This converts the constraint-violation rollback path into a clean early exit and
+	// is required for the reconciliation walk (§4.4) to replay events without errors.
+	// Reorged events with a changed block-level logIndex pass through this check;
+	// they are handled by the reactor's business-logic idempotency (see reorg-fix-spec.md §6.6).
```
philanton: don't reference temporary doc `reorg-fix-spec.md`

pkg/blockchain/evm/channel_hub_reactor.go
```diff
@@ -178,7 +186,25 @@ func (r *ChannelHubReactor) HandleEvent(ctx context.Context, l types.Log) error
...
+	processed, err := r.store.IsContractEventProcessed(l.TxHash.String(), uint32(l.Index), r.blockchainID)
+	if err != nil {
+		logger.Warn("failed to check if contract event was already processed, proceeding",
+			"error", err, "txHash", l.TxHash.String(), "logIndex", l.Index, "chainID", r.blockchainID)
```
philanton: shouldn't the error be returned?

pkg/blockchain/evm/channel_hub_reactor.go
```diff
@@ -178,7 +186,25 @@ func (r *ChannelHubReactor) HandleEvent(ctx context.Context, l types.Log) error
...
+	} else if processed {
+		logger.Info("skipping re-delivered event, already committed",
+			"event", eventName, "txHash", l.TxHash.String(), "logIndex", l.Index, "chainID", r.blockchainID)
+		if r.onEventProcessed != nil {
+			r.onEventProcessed(r.blockchainID, true)
```
philanton: why twice?

pkg/blockchain/evm/confirmation_gate.go
philanton, line 1: Had a closer look at `ConfirmationGate` and I think we can simplify it quite a bit. A few things I'd change:

1) **Drop the 50ms poll for a timer.** Instead of `pollInterval` ticking every 50ms, I'd keep a single `time.Timer` aimed at the head entry's deadline and wake the consumer early via a small `kick chan struct{}` (1-buffered, non-blocking send) whenever `HandleEvent` touches the queue. No poll jitter, no idle wakeups, and `HandleEvent` still returns immediately so it never back-pressures the listener;

2) **Get rid of `blockTimestampFetcher` + `blockTimestampCache`.** `types.Log` already carries `BlockTimestamp`, so we don't need the extra `HeaderByHash` RPC or the cache around it. One caveat: it's populated only if the node returns `blockTimestamp` in the log JSON — so I'd add a `Warn` when it comes back `0`, that way a misbehaving node defeats the delay *loudly* instead of silently. If we're not confident our nodes populate it, we keep the fetcher — but I'd bet we're fine here;

3) **Replace the O(n) queue scans with a tombstone map.** `removeFromQueueByKey`/`removeFromQueueByFullKey` both scan + shift the slice on every removal and re-delivery. I'd add a `pending map[eventKey]common.Hash` (key → live blockHash) as the source of truth: enqueue overwrites it, a removed-log cancels by `delete` when the blockHash still matches, and the queue becomes append-tail / pop-head only. Stale entries are just tombstones skipped at the head. Makes cancel and reorg re-delivery O(1);

4) **Same slice trick for `recentlyForwarded`.** Right now eviction scans the whole map every pass. Since we forward in order, I'd pair the map (membership, for the post-gate-reorg lookup) with a small expiry FIFO and pop it from the front — amortized O(1). One thing: key the TTL on `forwardedAt`, not `BlockTimestamp` — `forwardedAt` is monotonic so front-eviction stays correct, whereas a zero-`BlockTimestamp` event would break the ordering;

5) **Move the `delay == 0` check into the constructor.** Instead of branching on it in every `HandleEvent` call, I'd have the constructor return an error for `delay <= 0` and wire `reactor.HandleEvent` straight into the listener when the delay is off. Kills the per-call branch and the whole transparent-path.

I've got a working draft of all of this locally if it's easier to just look at the diff — lmk and I'll push it.

pkg/blockchain/evm/channel_hub_reactor.go
nksazonov: I plan to keep `reorg-fix-spec.md` as a documentation on how Nitronode is guarded against reorgs.

nitronode/docs/reorg-fix.md
nksazonov: I considered several locations for this file. 
- `/` is not relevant, as this file only relates to Nitronode
- `/docs` is not relevant either, as neither `/docs/protocol`, `/docs/legacy` nor `/docs/communication_flows` are suitable
- there is no `/nitronode/docs` directory created

Thus, I have decided just to put it in `/nitronode`. However, from what I see now, we can put it to a separate `/docs/nitronode`. WDYT?

nitronode/reorg-fix-spec.md
```diff
@@ -0,0 +1,374 @@
...
+
+No event handling, no Listener awareness. Drain-and-forward only.
+
+---
+
+**Properties**
+
+| Property | Detail |
+| --- | --- |
+| Zero RPC calls in the gate | Delay is a pure `time.Duration`; no chain queries |
```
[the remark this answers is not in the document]
nksazonov: b1fb759182412af5c2242d62b4b2d16a573d9bb1

nitronode/store/memory/blockchain_config.go
```diff
@@ -43,6 +43,10 @@ type BlockchainConfig struct {
...
+	// Set to 0 to process events immediately (disables the confirmation gate).
+	// Maximum meaningful value is ~780s (Ethereum Casper FFG hard finality).
+	ConfirmationDelaySecs uint32 `yaml:"confirmation_delay_secs"`
```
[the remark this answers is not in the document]
nksazonov: b1fb759182412af5c2242d62b4b2d16a573d9bb1

nitronode/store/memory/blockchain_config.go
nksazonov: Yes, I will set those for sure before the merge

nitronode/main.go
nksazonov: Fixed in ce0198753b057a701ee46efe10b059f30e387b95. I have made the listener to route only those historical events, that are older than `confirmation_window`, directly to the Reactor.

nitronode/docs/reorg-fix.md
philanton: We also planned to have `architecture` inside `docs`, so it'd make sense to locate it there

philanton: No, I think `/nitronode/docs` is better, because most of the SC-related docs are inside `contracts` dir, so nitronode-specific should be placed accordingly

pkg/blockchain/evm/reconciler.go
nksazonov: Indeed, nice catch!
Fixed in 5e234bb07d4916ba4eaadc7ba9c2a17d59c70428 by comparing stored hash with the canonical hash at that block number, and also handling `ethereum.NotFound` correctly.

pkg/blockchain/evm/confirmation_gate.go
nksazonov: @philanton — this is exactly what is now on the branch:

(1) The 50 ms `pollInterval` is gone — the gate now runs a single `time.Timer` aimed at the head deadline plus a 1-buffered `kick chan struct{}` that `HandleEvent` pings on every queue touch (4ff2a73b).
(2) Dropped `blockTimestampFetcher` and `blockTimestampCache` from the gate. The gate now reads `eventLog.BlockTimestamp` directly; the listener owns the fallback in `ensureBlockTimestamp` with a single-entry cache and falls through to a `HeaderByHash` only when the field is zero — kept because Avalanche C-Chain (ava-labs/libevm) and older BSC dataseed nodes don't populate it (76fc4cb7).
(3) `pending map[eventKey]common.Hash` is now the source of truth; the queue is append-tail / pop-head, and stale entries are tombstoned by a `pending[ek] != entry.log.BlockHash` check at the head (4ff2a73b).
(4) `forwardedSet` is paired with a `forwardedQueue` FIFO keyed on `forwardedAt`; eviction pops the front and the value-check (`storedAt.Equal(popped.forwardedAt)`) handles the re-forward case (4ff2a73b).
(5) `NewConfirmationGate` now returns an error on `delay <= 0`; `nitronode/main.go` branches on `confirmationDelay > 0` and wires `reactor.HandleEvent` straight into the listener otherwise — no per-call branch in the gate (4ff2a73b).

Tests and `reorg-fix-spec.md` were reworked to match in 7e0996b1.

pkg/blockchain/evm/reconciler.go
nksazonov: Fixed in 5e234bb0 — the reconciler now uses `HeaderByNumber(blockNum)` (in `isStoredBlockCanonical`) and compares its hash to the stored hash, so the stored-number-vs-header-number sanity branch is gone.

pkg/blockchain/evm/reconciler.go
nksazonov: Fixed in 5e234bb0 — `findCommonAncestor` and the new `isStoredBlockCanonical` helper both have expanded docstrings covering the canonicality check, the pre-migration / genesis fallback, and the `HeaderByNumber` vs `HeaderByHash` rationale.

pkg/blockchain/evm/listener.go
nksazonov: Fixed in ce019875 (and tidied in 76fc4cb7) — historical events younger than `confirmationDelay` are now routed through the gate via `routeHistoricalEvent`; only events whose block is past the reorg window go straight to the reactor, so the separation is age-based rather than phase-based.

nitronode/main.go
nksazonov: Fixed in 4ff2a73b — the gate-side block-timestamp fetcher was removed from `main.go` altogether, so `blockTimestampFetchTimeout` is gone. The listener's `ensureBlockTimestamp` uses the existing `rpcRequestTimeout` from `pkg/blockchain/evm`.

pkg/blockchain/evm/listener.go
nksazonov: Acknowledged in 2ead6119: added a code comment at the `prevHash == ""` branch and a §4.4 note explaining the empty-store / full-depth-reorg conflation is intentional; full-depth reorg to genesis is treated as a chain-level incident outside the gate's scope

pkg/blockchain/evm/confirmation_gate.go
nksazonov: Fixed in c398e28f: added a fatal-error channel (`gate.FatalErrors()`) plumbed into the listener's Phase 1+2 selects so handler errors unsubscribe → `logger.Fatal` → supervisor restart → DB-cursor replay, restoring the pre-PR crash-restart-replay semantics (verified by `TestConfirmationGate_HandlerErrorPropagatesFatal`)

nitronode/store/database/contract_event.go
nksazonov: Fixed in 4e7b954c: deleted `IsContractEventPresent`; listener now calls `IsContractEventProcessed` (which matches the actual unique index on `(transaction_hash, log_index, blockchain_id)`), keeping the per-phase boundary catch-up shortcut

nitronode/docs/reorg-fix.md
nksazonov: Fixed in 22159a9b: moved to `nitronode/docs/reorg-fix.md` (dropped the `-spec` suffix since it documents shipped behavior) and updated the in-code references

pkg/blockchain/evm/channel_hub_reactor.go
nksazonov: Fixed in 3512884b: pre-check now returns `errors.Wrap(err, "pre-check IsContractEventProcessed failed")` instead of warning and proceeding; the wrapped error flows through Comment 2's fatal channel to trigger restart + retry

pkg/blockchain/evm/channel_hub_reactor.go
nksazonov: Fixed in 3512884b: `HandleEvent` now uses a named return with a single deferred `onEventProcessed` call site, which also covers the previously-uncovered unknown-topic and pre-check-error paths

nitronode/store/memory/blockchain_config.go
ihsraham: Rechecked current HEAD (`3512884`). This still looks open in the diff: the checked-in `prod-v1`, `sandbox-v1`, and `stress-v1` `blockchains.yaml` files still omit `confirmation_delay_secs`, and omission still decodes as `0`, so `main.go` takes the no-gate path.

A pre-merge config change is fine operationally, but for the audit closure I would want this represented in the PR itself or enforced through validation/defaults. Also, `nitronode/docs/reorg-fix.md` currently documents `confirmation_delay_sec` singular, while the YAML tag/API field are `confirmation_delay_secs` plural. Following the doc as written would also leave the field unset.

ihsraham (commented):
Re-reviewed current head. The age-based historical routing, canonical `HeaderByNumber` check, and fatal error propagation look addressed. I am not approving yet because the checked-in config still leaves the gate disabled, and I left one more no-gate `Removed:true` path note inline.

pkg/blockchain/evm/listener.go
```diff
@@ -287,31 +349,47 @@ func (l *Listener) processEvents(
...
+						"blockHash", eventLog.BlockHash.Hex(),
+					)
+				} else {
+					eventLog = ensured
 				}
-				currentCheckDone = true
 			}
-			l.logger.Debug("received current event", "blockchainID", l.blockchainID, "contractAddress", l.contractAddress.String(), "blockNumber", eventLog.BlockNumber, "logIndex", eventLog.Index)
-			evCtx := log.SetContextLogger(context.Background(), l.logger)
 			if err := l.handleEvent(evCtx, eventLog); err != nil {
```
ihsraham, line 450: Could we keep `Removed:true` logs from reaching the reactor on the no-gate path? The base listener dropped removed logs before the handler. This PR forwards them so the gate can cancel pending entries, but when `confirmationDelay == 0`, `main.go` wires `handleEvent` directly to `reactor.HandleEvent`. In that mode a reorg removal is parsed like a normal log by the reactor.

One option is to only forward removed logs when `handleEvent` is the gate, or make the listener/gate wiring pass a removed-log handler that is never the raw reactor.

pkg/blockchain/evm/listener.go
ihsraham: Re-reading this after the latest commits, I do not think the code comment quite matches the operational case. This branch is reached whenever every stored *event block* is no longer canonical. That is not necessarily a chain-level reorg to genesis: a fresh node, or a chain with only one processed ChannelHub event, can hit it on a shallow reorg of that one event block.

In that case `findCommonAncestor` returns `0`, and `listenEvents` treats `lastBlock == 0` as “skip historical logs fetching”, so canonical replacement logs that already exist between the start boundary and the subscription tip are missed. For L-15 closure I would still prefer distinguishing empty store from “no stored event block is canonical” and replaying from a configured start/genesis boundary for the latter.

pkg/blockchain/evm/listener.go
philanton: Make sense to me, maybe we could have alternative simple implementation of ConfirmationGate that would simply drop + log removed events

pkg/blockchain/evm/confirmation_gate.go
```diff
@@ -0,0 +1,298 @@
...
+		done:           make(chan struct{}),
+	}, nil
+}
+
+// FatalErrors returns a read-only channel that receives the first handler error
+// encountered after the confirmation delay. The channel is buffered (size 1);
+// only the first error is delivered. When the channel fires, the gate's drain
+// goroutine has already stopped forwarding. The listener should unsubscribe and
+// return the error to trigger process restart and DB-cursor replay.
+func (g *ConfirmationGate) FatalErrors() <-chan error {
```
philanton: One more thing I don't like here — the `fatalCh` plumbing. The gate reports its failure through a sibling: `FatalErrors()` → constructor param → two select cases in the listener. So the listener has to know that its downstream handler may fail asynchronously, and the param is specific to this gate implementation.

I'd make the gate conform to the lifecycle contract `Listener` and the worker already have in `main.go`:

```go
func (g *ConfirmationGate) Start(ctx context.Context, handleClosure func(err error)) {
	go func() { handleClosure(g.run(ctx)) }()
}
```

and wire it in `main.go` identically to the other two:

```go
gate.Start(blockchainCtx, func(err error) {
	if err != nil {
		logger.Fatal("confirmation gate stopped", "error", err, "blockchainID", b.ID)
	}
})
```

`run` returns the handler error instead of pushing into `fatalCh`. Then `fatalCh`, `done`, `FatalErrors()`, the listener's `handleEventFatalCh` and both select cases all go away, and the listener stays gate-agnostic — error propagation lives in the supervisor, where it already is for the other two components. The unsubscribe-before-fatal we have now doesn't buy us anything since the process dies anyway.

pkg/blockchain/evm/listener.go
nksazonov: Fixed in 47098799: `findCommonAncestor` now splits the all-walk-exhausted branch — empty store (`latestNum == 0`) still skips replay; non-empty store with all blocks reorged (`prevNum == 0 && prevHash == "" && latestNum > 0`) returns the original `latestNum` as the resume anchor, so the listener re-fetches canonical replacements via `eth_getLogs` from the orphaned height (the orphan hash is discarded). Empty-store and all-reorged are no longer conflated.

pkg/blockchain/evm/listener.go
nksazonov: Fixed in 33c4b3a6: listener now drops `Removed:true` live logs at the Phase 2 boundary when `confirmationDelay == 0`, with a `Warn` log carrying `blockchainID`/`contractAddress`/`blockNumber`/`blockHash`/`txHash`/`logIndex`. The reactor no longer receives `Removed:true` logs in either mode. `TestListener_RemovedLog_ForwardedToHandler` refactored into `WithGate` (delay > 0, forwarded) and `NoGate` (delay == 0, dropped + follow-up event still flows) subtests.

pkg/blockchain/evm/confirmation_gate.go
nksazonov: Fixed in fcc4f8e8: replaced the fatal-channel plumbing with the closure-pattern lifecycle. `gate.Start(ctx, handleClosure func(err error))` mirrors `Listener.Listen` and `BlockchainWorker.Start`; `run(ctx) error` and `drainAndReschedule() error` propagate handler errors up to `main.go`'s `logger.Fatal`. Removed `fatalCh`, `done`, `FatalErrors()`, `handleEventFatalCh`, and the two listener select cases — listener is gate-agnostic again.

nitronode/store/memory/blockchain_config.go
ihsraham: Still open on current head (`1bc0d565`). The field-name/docs mismatch is fixed now, but the shipped chart configs still do not set `confirmation_delay_secs`.

Since the loader leaves the omitted field as `0` and `main.go` skips `NewConfirmationGate` when the delay is zero, the prod/sandbox/stress deploy path still processes live deposits immediately. The deploy workflow also passes those exact files via `--set-file=config.blockchains=.../blockchains.yaml`, so I would keep this open until the PR either sets non-zero values in those configs or enforces a safe default/validation for ChannelHub-enabled chains.

pkg/blockchain/evm/listener.go
ihsraham: The split helps, but I think there is still one replay boundary left here. In the all-stored-blocks-reorged branch we now return `latestNum`; if the canonical replacement block is already the current tip at that same height, `listenEvents` calls `reconcileBlockRange(currentBlock, lastBlock)` with equal values and `reconcileBlockRange` never enters `for currentBlock > startBlock`.

That means the replacement log at the resume height is not fetched. Could we make the replay range include the resume height for this case, or return a start height that guarantees the orphaned height is queried?

pkg/blockchain/evm/listener.go
```diff
@@ -287,27 +340,47 @@ func (l *Listener) processEvents(
...
-				present, err := l.eventGetter.IsContractEventPresent(l.blockchainID, eventLog.BlockNumber, eventLog.TxHash.Hex(), uint32(eventLog.Index))
-				if err != nil {
-					eventSubscription.Unsubscribe()
-					return fmt.Errorf("failed to check current event presence: %w", err)
-				}
-				if present {
-					l.logger.Debug("skipping already present current event", "blockchainID", l.blockchainID, "blockNumber", eventLog.BlockNumber, "logIndex", eventLog.Index)
-					continue
+			if !eventLog.Removed {
+				*lastBlock = eventLog.BlockNumber
```
ihsraham, line 421: This still leaves a reconnect gap for pending gate entries. The listener advances `lastBlock` before the gate has committed the event, and `HandleEvent` only queues it. If the subscription drops after an orphaned live log is queued but before the matching `Removed:true` arrives, the next loop replays only canonical logs and never sends a cancellation for that pending entry. The old pending entry can then mature in the gate and reach the reactor even though its block is no longer canonical.

Could we keep the resume cursor tied to committed events, clear/reconcile pending gate entries on reconnect, or have replay explicitly cancel queued entries that no longer appear canonically before they can mature?

ihsraham (changes requested):
Requesting changes for L-15 closure. The main implementation is much closer now: no-gate `Removed:true` logs no longer reach the reactor, recent historical events route through the gate, canonicality uses `HeaderByNumber`, and handler failures after dequeue now restart instead of being swallowed. I resolved/reacted on the threads that look addressed.

I would still keep L-15 open on three points: shipped configs still decode to `confirmation_delay_secs: 0`, pending gate entries can survive a reconnect if the removal signal is missed, and the all-stored-blocks-reorged replay path still misses the same-height replacement boundary. I replied on the existing config/full-depth threads and left one new inline comment for the reconnect case.

philanton (approved):

pkg/blockchain/evm/listener.go
nksazonov: Fixed in 50cfb038: relaxed `reconcileBlockRange`'s loop guard from `>` to `>=` so the boundary block at the resume height is fetched (`FilterLogs(N, N)` when `currentTip == lastBlock`); new `TestListener_ReconcileBlockRange_EqualBounds` catches the regression, and downstream dedup at `listener.go:357,423` absorbs any boundary re-fetch on healthy reconnects.

pkg/blockchain/evm/listener.go
nksazonov: Fixed in a952b79c: on subscription drop the listener now flushes the gate's pending entries via the new `FlushPending()` and rewinds the in-memory cursor by re-running `findCommonAncestor` per `runOneListenPass` iteration (cursor is now per-pass local, not a closure-captured variable), so Phase 1's `eth_getLogs(committedCursor, tip)` re-fetches the canonical events that Geth won't replay on a fresh subscription; see `nitronode/docs/reorg-fix.md` §6.9 for the full semantics.

nitronode/store/memory/blockchain_config.go
nksazonov: Addressed in c094109e — added `confirmation_delay_secs` to all three chart configs (`prod-v1`, `sandbox-v1`, `stress-v1`) with per-chain values targeting ~99.9% reorg-safety and inline justifications.

nksazonov:
With the confirmation gate now inserting a deliberate per-chain delay between on-chain event observation and off-chain state commit, I audited downstream consumers (SDKs, cerebro, playground, MCP) for impact.

## Confirmation-Window Impact Across Consumers

### TL;DR

The wire field `confirmation_delay_secs` is **already plumbed end-to-end** through every SDK and surface — `pkg/rpc/types.go:269`, `pkg/core/types.go:1097`, `sdk/go/utils.go:33`, `sdk/ts/src/utils.ts:45`. The protocol surface is fine. The real risk is **stale UX** in interactive surfaces ("✓ Confirmed!" toast followed by 13 minutes of unchanged balance on Ethereum mainnet).

### Critical findings

**Playground.** `useChannelOps.tsx:139,171,267` does `waitForTransactionReceipt` and treats that as "done." But the receipt completes at the moment the gate's countdown *starts*. Result: green success toast + unchanged balance for `confirmation_delay_secs`. The UX label `'confirming'` (`ActionPanel.tsx:231,239`) actively mislabels what's happening. Worse: `useNitrolite.ts:163` already fetches `cfg.blockchains` but **discards every field except `id`** (`:169`) — `confirmationDelaySecs` is on the wire and ignored.

**Cerebro.** `checkpoint` (`cerebro/commands.go:442`) returns immediately on tx-hash; `nodeInfo` (`:470`) hides `ConfirmationDelaySecs` despite it being on the struct. No timeout (REPL won't hang), but user runs `balances` and sees stale numbers with **zero explanation**.

**SDK MCP.** `sdk/mcp/src/index.ts:1518` literally claims **"Instant finality — no waiting for block confirmations"**. Misleading post-gate — should clarify that off-chain transfers are instant; on-chain deposits/withdrawals lag by `confirmation_delay_secs`.

### Recommended changes (priority order)

1. **Playground** (significant code + UX work):
   - Persist full `Blockchain[]` including `confirmationDelaySecs` (`useNitrolite.ts:163,183`)
   - Add `'awaiting_node'` phase after `waitForTransactionReceipt` (`useChannelOps.tsx:135-141, 167-173, 264-269`)
   - Relabel `'confirming'` → "Tx mined, awaiting node (~Xs)…" (`ActionPanel.tsx:225-245`)
   - Defer success toast until polled credit lands, or split into "Tx mined" + "Credit confirmed"

2. **TS + Go SDKs** (small ergonomic additions):
   - `Client.getConfirmationDelay(chainId)` — convenience wrapper
   - `Client.waitForCheckpoint(asset, txHash, opts)` — polls `getBalances`/`getLatestState` with lower-bound wait of `confirmation_delay_secs`

3. **Cerebro** (text + optional helper):
   - `nodeInfo` and `listChains` should print delay per chain (`commands.go:482, 522`)
   - `checkpoint` should print "Off-chain credit in ~Ns" after tx (`commands.go:442`)
   - Optional `wait-credit <asset>` command (sleeps + polls)

4. **Docs sweep**:
   - `sdk/ts/README.md` near `:142, 377, 905`
   - `sdk/go/README.md:131, 336`
   - `sdk/ts-compat/docs/migration-onchain.md:94` (migration note for v0.5.3 push-based consumers)
   - `sdk/mcp/src/index.ts:1518` (the "Instant finality" line)
   - `cerebro/README.md`, `playground/README.md:62-69`
   - `nitronode/docs/reorg-fix.md` — add "Client-side implications" section

5. **sdk/ts-compat** — primarily docs. `EventPoller` (5s polling) already handles arbitrary delay; v0.5.3 push-style consumers just need a migration note.

ihsraham (approved):
Approved. The latest updates address the L-15 closure points I was looking for: the shipped configs now carry explicit `confirmation_delay_secs`, replay/reconciliation goes back through the confirmation path, same-height/full-depth reorg replay is covered, and `Removed:true` logs are routed through the gate instead of being dropped.

Issue #836, opened 2026-06-14
Scheduler dedup: blockchain_actions table allows duplicate Pending rows on event replay


nksazonov:
## Summary

`scheduleStateEnforcement` in `nitronode/store/database/blockchain_action.go` unconditionally inserts a new `blockchain_actions` row via `db.Create`, with no dedup on `(state_id, action_type)` or `(state_id, blockchain_id, action_type)`. Same-event replay enqueues a duplicate Pending action.

## Hazard

When an event handler that schedules an action (most notably `HandleEscrowDepositInitiated` → `ScheduleInitiateEscrowDeposit`) receives an **equal-version replay** — which can happen on indexer replay or chain reorg replay — a second `blockchain_actions` row is inserted for the same `(state_id, action_type)`.

The action worker (`GetActions`, same file) filters by `status=Pending` and picks up both rows. The second broadcast reverts on chain (the contract state-machine rejects re-initialization of the same state), but the Node records the duplicate as a failed `blockchain_actions` row that an operator must inspect.

## Why this is exposed

The event-handler version guards in `nitronode/event_handlers/service.go` (helper `guardEventVersionMonotonic`) use strict `<` (not `<=`) and **deliberately admit equal-version replays as a no-op**, because legitimate indexer-replay and reorg-replay re-deliver identical `(channelId, stateVersion)` pairs. Most of the handler body is idempotent on that replay path (`UpdateChannel`, `RefreshUserEnforcedBalance`, `UpdateStateSigsIfMissing`). Scheduling is the one side effect that is **not** idempotent — hence this gap.

## Proposed fix

1. Add a unique partial index on `(state_id, action_type)` for `status=Pending` rows.
2. Change `scheduleStateEnforcement` to use `INSERT ... ON CONFLICT DO NOTHING` (or the GORM equivalent: `clause.OnConflict{DoNothing: true}`).

This makes the schedule path fully idempotent across the replay chain — the complement to the existing `<` (not `<=`) version guard in the event handlers.

## Files involved

- `nitronode/store/database/blockchain_action.go` (~L102–113) — the unconditional `db.Create`.
- `nitronode/store/database/` — wherever the model migrations live, for the unique-index migration.

## Operational impact today

Low — duplicate broadcasts are bounded (one per replay event), revert on chain (gas cost only, no state corruption), and surface as visible failed `blockchain_actions` rows. No user funds at risk. This is a hygiene issue, not a safety issue.

## Labels

[...]

PR #837, opened 2026-06-15
MF3-L19: fix: prevent SC reentrancy + event-handler monotonicity

nksazonov:
## Summary

Addresses audit finding MF3-L19 (reentrancy in `ChannelHub` lifecycle entrypoints) with paired contract-side and Nitronode-side fixes, plus supporting documentation.

- **`fix(contracts)`** — added `nonReentrant` to every external/public lifecycle entrypoint in `ChannelHub.sol` (root-cause remediation). Migrated guards upward from internal fund helpers since OZ `ReentrancyGuard` uses a single shared status slot. New regression tests for the four reentrancy scenarios.
- **`fix(nitronode)`** — added version-monotonicity guards to six previously guard-less event handlers (`HandleHomeChannelCheckpointed`, `HandleHomeChannelClosed`, four escrow Initiated/Finalized handlers) via a shared helper. Drops stale events with a structured warn log instead of regressing local state or silently clearing challenges.
- **`feat(nitronode)`** — `ChainStateRefresher` interface + EVM implementation. On home-channel guard drop, fetches authoritative on-chain state via `getChannelData` and overwrites the local row. Defense-in-depth against any class of out-of-order delivery (indexer mis-order, reorg replay, future contract changes).
- **`docs`** — new sections in `contracts/SECURITY.md`, `docs/protocol/security-and-limitations.md`, `nitronode/README.md`; new deployment-constraint matrix at `contracts/deployments/HOOK-TOKEN-COMPATIBILITY.md` recording per-chain hook-token support; WARNING blocks in the three `nitronode/chart/config/*/assets.yaml` files.

Follow-up tracked at #836 (scheduler dedup for `blockchain_actions`).

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./nitronode/... ./pkg/core/... ./pkg/blockchain/evm/...` — clean
- [x] `go test ./nitronode/event_handlers/... ./pkg/blockchain/evm/... ./pkg/core/... -count=1` — green
- [x] `forge test` — 292 passed, 0 failed (288 pre-existing + 4 new reentrancy regression tests)
- [x] `forge fmt --check` — clean

## Release Notes

* **Security Enhancements**
  * Added reentrancy protection to all channel lifecycle operations, preventing cross-function reentrancy attacks via token hooks
  * Node now detects out-of-order blockchain events and automatically refreshes state from the chain to ensure convergence

* **Known Limitations**
  * Hook-enabled tokens (ERC777, ERC1363, and certain non-standard ERC20s) are not supported on currently deployed ChannelHub instances; node operators must not onboard such tokens


nksazonov:
@coderabbitai review

nitronode/event_handlers/service.go
```diff
@@ -18,16 +18,115 @@ var _ core.ChannelHubEventHandler = &EventHandlerService{}
...
+	tx core.ChannelHubEventHandlerStore,
+	channel *core.Channel,
+	droppedIntent string,
+) error {
+	logger := log.FromContext(ctx)
+	refreshed, err := s.refresher.RefreshChannelFromChain(ctx, channel.ChannelID)
+	if err != nil {
+		// Surface; if the on-chain read fails we cannot safely converge and must
+		// retry the event by returning a non-nil error so the listener replays.
+		return fmt.Errorf("refresh after dropped %s: %w", droppedIntent, err)
```
ihsraham: Before approval, I would not let this error reach the listener fatal path. Returning a refresh error does roll back the event tx, but `processEvents` returns it and `main.go` calls `logger.Fatal`, so a transient `getChannelData` failure on a stale event can stop nitronode. Can we keep the replay semantics with listener-level retry/backoff or a retryable error path instead of terminating the process?

pkg/blockchain/evm/chain_state_refresher.go
```diff
@@ -0,0 +1,117 @@
...
+	if !ok || caller == nil {
+		return nil, fmt.Errorf("no ChannelHub caller registered for chain %d (channel %s)", chainID, channelID)
+	}
+
+	channelIDBytes, err := hexToBytes32(channelID)
+	if err != nil {
+		return nil, fmt.Errorf("invalid channel ID %q: %w", channelID, err)
+	}
+
+	data, err := caller.GetChannelData(&bind.CallOpts{Context: ctx}, channelIDBytes)
```
ihsraham: Could we put a deadline on this `GetChannelData` call? The listener builds event contexts from `context.Background()`, and the home-channel handlers call this after taking the user balance lock. If the RPC hangs, the listener and that user's locked row can stay blocked. Reusing the listener RPC timeout or passing a bounded refresh context would close that gap.

contracts/deployments/HOOK-TOKEN-COMPATIBILITY.md
```diff
@@ -0,0 +1,34 @@
...
+| 137 | Polygon | `0x1a2f750170474d4c54f8d318d9d4343588b4c4d1` | `e07ad9c2` | prod v1.3.0 |
+| 480 | World Chain | `0x1a2f750170474d4c54f8d318d9d4343588b4c4d1` | `e07ad9c2` | prod v1.3.0 |
+| 8453 | Base | `0x1a2f750170474d4c54f8d318d9d4343588b4c4d1` | `e07ad9c2` | prod v1.3.0 |
+| 59144 | Linea | `0x1a2f750170474d4c54f8d318d9d4343588b4c4d1` | `e07ad9c2` | prod v1.3.0 |
+| 80002 | Polygon Amoy | `0x5dba8515af063db0c243c15ece7b99f91459c7c3` | `b88d511c` | sandbox v1.3.0 |
+| 84532 | Base Sepolia | `0x5dba8515af063db0c243c15ece7b99f91459c7c3` | `b88d511c` | sandbox v1.3.0 |
+| 59141 | Linea Sepolia | `0x5dba8515af063db0c243c15ece7b99f91459c7c3` | `b88d511c` | sandbox v1.3.0 |
+| 1440000 | XRPL EVM | `0x1a2f750170474d4c54f8d318d9d4343588b4c4d1` | `e07ad9c2` | prod v1.3.0 |
+| 1449000 | XRPL EVM Testnet | `0x5dba8515af063db0c243c15ece7b99f91459c7c3` | `b88d511c` | sandbox v1.3.0 |
+| 11155111 | Sepolia | `0x5dba8515af063db0c243c15ece7b99f91459c7c3` | `b88d511c` | sandbox v1.3.0 |
```
ihsraham, line 35: This matrix also needs the active `stress-v1` ChannelHub deployments. `stress-v1/assets.yaml` points operators here, but `stress-v1/blockchains.yaml` uses Sepolia `0x7d61ec...` and Base Sepolia `0x61b9e...`, while this matrix only lists the sandbox rows for those chain IDs. Since current deployments rely on the no-hook-token policy, missing active environment rows weakens the closure proof.

contracts/SECURITY.md
```diff
@@ -540,6 +553,8 @@ Inbound transfer failures occur during:
 
 **Mitigation**: The Nitronode only processes operations after observing successful on-chain events. If a user signs a deposit state but the transfer fails on-chain, the state is never enforced, and the Node does not provide services based on unconfirmed deposits.
 
+**Reentrancy via inbound hooks**: Tokens whose `transferFrom` invokes a recipient hook (ERC777-style `tokensReceived`, ERC1363 callbacks, or non-standard ERC20 implementations) could in principle re-enter `ChannelHub` lifecycle entrypoints during an inbound pull. This is blocked by the `nonReentrant` guard on every lifecycle entrypoint — see invariant 26 above and `contracts/src/ChannelHub.sol`. The guard makes already-deployed contracts that historically operated under an off-chain "no hook-bearing tokens" policy safe against this class of attack at the contract layer; the off-chain policy remains in effect for already-deployed contracts at specific addresses as a defense-in-depth measure pending operator-controlled token onboarding decisions.
```
ihsraham: This overstates protection for existing deployments. My read from the remediation note and the matrix is that current deployed addresses are not redeployed, so they are not made safe by this new `nonReentrant` guard; their control is the no-hook-token onboarding policy. Please split the wording: future or remediated deployments are protected at the contract layer, while current listed deployments are protected by policy and config only.

ihsraham (changes requested):
The contract-side reentrancy fix looks solid: moving `nonReentrant` to the lifecycle entrypoints is the right shape, and the new Forge coverage exercises the pull-hook cases. I would still block this for now because the new node refresh path can fail closed by terminating or hanging the listener, and the current-deployment hook-token policy is not documented precisely enough for closure.

I also agree with the existing thread on the `channelHubCallers` map race and reacted there rather than repeating it.

nitronode/main.go
```diff
@@ -78,7 +78,23 @@ func main() {
...
+		if err != nil {
+			return 0, fmt.Errorf("lookup channel %s: %w", channelID, err)
+		}
+		if ch == nil {
+			return 0, fmt.Errorf("channel %s not found in DB", channelID)
+		}
+		return ch.BlockchainID, nil
+	}
+	chainStateRefresher := evm.NewEVMChainStateRefresher(channelHubCallers, resolveChannelChain)
+	eventHandlerService := event_handlers.NewEventHandlerService(nodeChannelSigner, eventHandlerStatePacker, chainStateRefresher)
```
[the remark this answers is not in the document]
nksazonov: `channelHubCallers` was removed in [1bed006](https://github.com/layer-3/nitrolite/pull/837/commits/1bed006fc56e2f7333557396c0b3c2b5b203befc)

contracts/SECURITY.md
```diff
@@ -51,6 +51,13 @@ Invariant:
...
 
+---
+
+8. The Node processes on-chain channel-lifecycle events with per-channel version monotonicity. An event whose `StateVersion` is strictly less than the row's current `StateVersion` is dropped with a structured warn log (see `nitronode/event_handlers/service.go`). For home-channel events (`ChannelChallenged`, `ChannelCheckpointed`, `ChannelClosed`), a dropped event additionally triggers an on-chain `getChannelData` read via the `ChainStateRefresher` (`pkg/blockchain/evm/chain_state_refresher.go`, interface in `pkg/core/interface.go`) that overwrites the local row's `Status`, `StateVersion`, and `ChallengeExpiresAt`. This is defense-in-depth against out-of-order event delivery from indexer mis-order, reorg replay, or any future contract change that re-introduces a same-transaction event-order quirk. Escrow event handlers enforce the guard without the refresh hook; cross-chain RPC plumbing for escrow refresh is a follow-up item.
+
```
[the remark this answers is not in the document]
nksazonov: Fixed in 163a2fdf. Aligned the wording across all three docs — `contracts/SECURITY.md` and `nitronode/README.md` now match `docs/protocol/security-and-limitations.md`'s "deferred follow-up item" phrasing and include the interim-divergence sentence about escrow rows remaining divergent until the next on-chain event arrives.

contracts/deployments/HOOK-TOKEN-COMPATIBILITY.md
nksazonov: Fixed in 163a2fdf. Added two stress-v1 rows to the matrix:
- Sepolia (chain 11155111): `0x7d61ec428cfae560f43647af567ea7c6e2cc5527`, commit `104c13df`, tag "stress v1.3.0"
- Base Sepolia (chain 84532): `0x61b9e0767f2eca7e33802e82f9c64b1ebe72ba31`, commit `9110ba06`, tag "stress v1.3.0"

Both predate `2a6a9f0d` and rely on the no-hook-token policy.

contracts/SECURITY.md
nksazonov: Fixed in 163a2fdf. Split the wording to distinguish vintages: future deployments built from commit `2a6a9f0d` or later carry the contract-layer guard; currently-deployed addresses listed in `contracts/deployments/HOOK-TOKEN-COMPATIBILITY.md` are protected only by the off-chain no-hook-token policy.

nitronode/event_handlers/service.go
nksazonov: Addressed in 92299e39. Switched `refreshAfterDroppedEvent` from "return wrapped error" to "log Error and return nil" — the dedup row commits, the listener moves on, and the process no longer exits on a transient RPC blip. Trade-off documented in the helper's doc-comment: in the rare case of a sustained RPC outage coinciding with a guard-drop, the local row may stay divergent from chain until the next on-chain event for that channel arrives. For terminal states (Closed) this may be indefinite — accepted as strictly better than terminating the node. Tests renamed to `LoggedAndIgnored` to reflect the new contract.

pkg/blockchain/evm/chain_state_refresher.go
nksazonov: Addressed in 92299e39. Added `refreshRPCTimeout = 5 * time.Second` package constant in `pkg/blockchain/evm/chain_state_refresher.go` and wrapped the incoming `ctx` with `context.WithTimeout` before the `GetChannelData` call. 5s matches other per-operation RPC timeouts in `nitronode/main.go` and `nitronode/runtime.go`; comfortably above realistic call latency on slow public RPCs while bounding tail latency under the per-user balance lock. On timeout, the failure flows through the log-and-continue path from the other comment.

pkg/core/types.go
```diff
@@ -140,6 +140,21 @@ func NewChannel(channelID, userWallet, asset string, ChType ChannelType, blockch
...
 
+// RefreshedChannel carries the authoritative on-chain channel snapshot used by
+// ReadOnlyChannelHub to converge a Node row that has diverged from chain.
+//
+// The snapshot reflects on-chain state at RPC-read time, not event-emit time:
+// the contract may have advanced the channel through additional transitions
+// between when the dropped event was emitted and when the refresh RPC ran. The
+// Node row may therefore briefly skip an intermediate status it never observed,
+// but it will always converge to a status the chain currently asserts.
+type RefreshedChannel struct {
```
philanton: please find a better name

pkg/blockchain/evm/channel_hub_reader.go
philanton, line 1: rename the file

nitronode/event_handlers/service.go
```diff
@@ -22,14 +22,112 @@ type EventHandlerService struct {
...
+	droppedIntent string,
+) error {
+	logger := log.FromContext(ctx)
+	refreshed, err := hub.FetchChannel(ctx, channel.ChannelID)
+	if err != nil {
+		logger.Error("refresh after dropped event failed, leaving row possibly divergent from chain",
+			"channelId", channel.ChannelID,
+			"droppedIntent", droppedIntent,
+			"error", err)
+		return nil
```
ihsraham, line 183: This avoids the process exit, but I would still block here because the failed refresh is now acknowledged as processed. After this returns nil, the reactor records the contract event, so the listener will not replay it; the local row can stay divergent from chain on the guard-drop path this refresh is meant to close.

Can we keep the non-fatal behavior while making refresh failure retryable, for example with listener-level retry/backoff or a bounded local retry before the event is deduped?

ihsraham (changes requested):
The contract-side guard placement looks good now, and the deployment policy/docs issues I raised are addressed. The per-chain reader also removes the map race, and the refresh RPC is bounded.

I’m keeping this as changes requested for one remaining closure issue: refresh failures are now logged and swallowed, so a guard-drop event can be marked processed without the row converging to chain.

nitronode/event_handlers/service.go
ihsraham: Keeping this unresolved because the current change fixes the node-exit part, but drops the replay/convergence property. Once `refreshAfterDroppedEvent` logs and returns nil, the reactor records the event, so a transient RPC failure can leave the local row stale with no replay.

I moved the current-head blocker here: https://github.com/layer-3/nitrolite/pull/837#discussion_r3413953011

pkg/core/types.go
nksazonov: Addressed in e93821d8. Renamed `RefreshedChannel` → `OnChainChannelSnapshot` (it is the on-chain snapshot returned by `ReadOnlyChannelHub.FetchChannel`, not a thing whose identity is "having been refreshed").

pkg/blockchain/evm/channel_hub_reader.go
nksazonov: Addressed in e93821d8. Renamed `pkg/blockchain/evm/chain_state_refresher.go` → `channel_hub_reader.go` to match its primary type `EVMChannelHubReader`.

nitronode/event_handlers/service.go
nksazonov: Addressed in ff789533. Added bounded in-handler retry (3 attempts, 100ms+200ms exponential backoff, ~700ms worst case before fall-back to log+nil) so a transient RPC blip is recovered without process exit; tests cover both the second-attempt-success and ctx-cancel-during-backoff paths.

ihsraham (approved):

pkg/blockchain/evm/channel_hub_reader.go
```diff
@@ -0,0 +1,100 @@
...
+	onchainChannelStatusClosed      uint8 = 3
+	onchainChannelStatusMigratingIn uint8 = 4
+	onchainChannelStatusMigratedOut uint8 = 5
+)
+
+// refreshRPCTimeout bounds the time `FetchChannel` will wait for a
+// `getChannelData` RPC response. Set to 5s to match other per-operation RPC
+// timeouts in the codebase (`nitronode/main.go`, `nitronode/runtime.go`) and to
+// stay well above realistic call latency on slow public RPCs.
+const refreshRPCTimeout = 5 * time.Second
```
philanton: rename pls

philanton (approved):
Good work!

pkg/blockchain/evm/channel_hub_reader.go
nksazonov: Renamed, extracted and replace some similar values in `nitronode/runtime.go` in e3c34063.

PR #880, opened 2026-08-02
fix(core): validate decimal precision by value, not by scale (unblocks app-session Withdraw)

panzagianluca:
## Problem

`submit_app_state` with `intent=Withdraw` fails for any 6-decimal asset on mainnet:

```
invalid withdraw amount for allocation with asset usdt and participant 0x77b8…8609:
amount exceeds maximum decimal precision: max 6 decimals allowed, got 18
```

The rejected value is not one the client sends. `handleWithdrawIntent` derives it:

```go
withdrawAmount := currentAmount.Sub(incomingAmount)
...
core.ValidateDecimalPrecision(withdrawAmount, decimals)
```

The client's own allocation amount is already validated earlier in the shared path and passes — we send canonical 2-dp strings such as `0.03`.

## Cause

`ValidateDecimalPrecision` tests the decimal's **exponent** (its scale) rather than its actual precision:

```go
if amount.Exponent() < -int32(maxDecimals) {
```

Scale is widened by ordinary arithmetic and by storage round-trips. `GetParticipantAllocations` reads balances via

```sql
COALESCE(SUM(l.credit), 0) - COALESCE(SUM(l.debit), 0) AS balance
```

which returns an 18-place scale; `decimal.Sub` rescales to the smaller exponent of its operands, so `withdrawAmount` inherits scale −18 and is rejected even though the value (e.g. `0.030000000000000000`) is exactly representable in 6 decimals.

`handleOperateIntent` performs the same subtraction and does not precision-check the result, which is why Operate works and only Withdraw fails.

## Fix

Compare against the truncated value instead of inspecting the exponent. Amounts that genuinely require more precision than the asset allows are still rejected; amounts that merely carry a wide scale are accepted.

Alternative if you prefer a narrower change: normalize `withdrawAmount` to the asset's decimals inside `handleWithdrawIntent` before validating, and leave `ValidateDecimalPrecision` untouched. Happy to reshape the PR that way.

## Tests

All existing `TestValidateDecimalPrecision` cases are unaffected — none of them uses a value whose scale exceeds its significant digits. Added three cases:

- `trailing_zeros_within_precision` — `0.030000000000000000` accepted at 6 dp
- `trailing_zeros_from_subtraction` — reproduces the withdraw path (`0.060000000000000000 - 0.03`)
- `wide_scale_with_real_excess_precision_still_rejected` — `0.030000010000000000` still rejected at 6 dp

Verified locally with `go1.26.5`:

```
go test ./pkg/core/ -count=1                     ok
go test ./pkg/... -count=1                       ok  (app, blockchain/evm, core, log, rpc, sign, sign/kms, sign/kms/gcp)
go test ./nitronode/api/app_session_v1/ -count=1 ok  (the deposit/operate/withdraw callers)
```

## Reproduction

Mainnet, `wss://nitronode.yellow.org/v1/ws`, Base 8453, USDT, 2-party app session. Deposit and Operate both succeed; only Withdraw fails. Two runs with different amounts produced the identical error:

| UTC | Session | Slot → target |
|---|---|---|
| 2026-07-29 12:03:45 | `0x249dfd0e086eb44dd84c896f435a129abbe1e3b37c6e5c32048c0e09fe104797` | 0.06 → 0.03 |
| 2026-07-29 12:05:27 | `0x7589be438fa4a4249cf913f43121a20d893d3832bf42d07ba9e7d8819c7e89ab` | 0.06 → 0.05 |

## Impact

Without this, funds can only leave an app session via `Close`, so a participant cannot move a realized balance back to their home channel while the session is still open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decimal precision validation to accept values with trailing zeros when they remain representable at the allowed precision.
  * Continued rejecting values containing genuinely unsupported fractional digits.
  * Improved handling of precision for values produced through subtraction.


panzagianluca:
Ran the suites locally after opening this (go1.26.5): `./pkg/core`, the full `./pkg/...` tree, and `./nitronode/api/app_session_v1/` — all green, including the existing `TestValidateDecimalPrecision` cases. Updated the description accordingly.

Also worth flagging: `ValidateDecimalPrecision` is called from `submit_deposit_state.go`, `submit_app_state.go` and `pkg/core/state_advancer.go`. In each, the change is strictly more permissive and only for values that are already exactly representable in the asset's decimals — amounts genuinely needing more precision are still rejected.

panzagianluca:
Thanks — addressed. Note there was already a doc comment on the function; rather than adding a second one I rewrote it, since the previous wording ("doesn't exceed the maximum allowed decimal places") is ambiguous now that the rule is value-based:

```go
// ValidateDecimalPrecision returns an error when amount is not exactly
// representable with maxDecimals fractional digits.
//
// The rule is value-based, not scale-based: an amount carrying trailing
// zeros beyond maxDecimals (for example 0.030000000000000000 against a
// 6-decimal asset) is valid, because no significant digit is lost by
// representing it with maxDecimals.
```

`./pkg/core` still green.

